PR #8, opened 2020-04-23
Format code with goimports

mrKisaoLamb:
- reformat code with goimports
- add info about goimports into README

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -37,8 +37,8 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost       = "0.0.0.0"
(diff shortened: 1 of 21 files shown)


cmd/manager/main.go
```diff
@@ -2,49 +2,41 @@ package main
 
 import (
 	"context"
-	"errors"
```
stuggi, line 5: This looks to be the old main.go from before https://github.com/openstack-k8s-operators/nova-operator/commit/9d783766293f93116524e046d541e8d30cedd6ee#diff-9e63f7b0791edcb4a53c944552b88970 . have you rebased after that?

mrKisaoLamb: Done

PR #13, opened 2020-05-04
Initial travis setup that supports lint and vet

mpryc:
Simple style type check jobs for the project with place for
further stages such as unit tests.

### DIFF
--- a/pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
+++ b/pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
@@ -31,7 +31,7 @@ var ospHostAliases = []corev1.HostAlias{}
 
 // TODO move to spec like image urls?
 const (
-	COMMON_CONFIGMAP string = "common-config"
+	CommonConfigMAP string = "common-config"
 )
 
 // Add creates a new NovaMigrationTarget Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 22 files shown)


pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
```diff
@@ -237,11 +235,11 @@ func (r *ReconcileNovaMigrationTarget) setDaemonsetHash(instance *novav1.NovaMig
 }
 
 func newDaemonset(cr *novav1.NovaMigrationTarget, cmName string, configHash string) *appsv1.DaemonSet {
-	var hostToContainer corev1.MountPropagationMode = corev1.MountPropagationHostToContainer
```
mpryc, line 240: I am unsure if this can be removed. Was there intention to convert the type of that variable?

pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
```diff
@@ -237,11 +235,11 @@ func (r *ReconcileNovaMigrationTarget) setDaemonsetHash(instance *novav1.NovaMig
...
-	var trueVar bool = true
-	var userId int64 = 0
+	var hostToContainer = corev1.MountPropagationHostToContainer
+	var trueVar = true
+	var userID int64
 	var configVolumeDefaultMode int32 = 0600
-	var dirOrCreate corev1.HostPathType = corev1.HostPathDirectoryOrCreate
```
mpryc, line 244: Same here I am unsure if this can be removed. Was there intention to convert the type of that variable?

pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
mrKisaoLamb: IMO there is no reasone for doing that and 'var hostToContainer = corev1.MountPropagationHostToContainer' is enought. @stuggi ?

mrKisaoLamb:
Generall note: this changes are in quite big conflict with https://github.com/openstack-k8s-operators/nova-operator/pull/11 maybe it will be better to merge it first? What do you think?

pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
stuggi: this is intentional to set the MountPropagationMode to the predefined constant MountPropagationHostToContainer [1]

[1] https://godoc.org/k8s.io/api/core/v1#MountPropagationMode

pkg/controller/novamigrationtarget/novamigrationtarget_controller.go
stuggi: same here

stuggi:
yes, we should first merge https://github.com/openstack-k8s-operators/nova-operator/pull/11 as it is a big change to the operator.

stuggi:
> yes, we should first merge #11 as it is a big change to the operator.

I have merged #11 , we should rebase this.

PR #15, opened 2020-05-05
Initial travis setup that supports lint and vet

mpryc:
Simple style type check jobs for the project with place for
further stages such as unit tests.

### DIFF
--- a/pkg/novamigrationtarget/secret.go
+++ b/pkg/novamigrationtarget/secret.go
@@ -7,10 +7,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// BITSIZE
 const (
 	BITSIZE int = 4096
 )
 
+// Secret func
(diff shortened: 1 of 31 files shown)


pkg/novamigrationtarget/secret.go
```diff
@@ -7,10 +7,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// BITSIZE
 const (
 	BITSIZE int = 4096
```
stuggi, line 12: wondering why it does not complain for BITSIZE here, but for the other const in CAP

stuggi:
other then the above lgtm

pkg/novamigrationtarget/secret.go
mpryc: @stuggi checked and it complains for any name with _ so it would with BIT_SIZE but BITSIZE is ok for that lint. Not ideal but was fixing just the ones that were visible. Later we may want to switch to something more sophisticated like golint-ci if required.

PR #26, opened 2020-05-28
Automate SecurityContextConstraints creation

dprince:
This patch automatically adds the SCC creation for
nova-operator during startup. This is the one remaining
configuration item not automated by OLM which we require
for nova-operator.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -27,6 +27,8 @@ import (
 	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


cmd/manager/main.go
```diff
@@ -217,3 +235,56 @@ func serveCRMetrics(cfg *rest.Config, operatorNs string) error {
...
+			Labels: map[string]string{
+				"nova.openstack.org": "",
+			},
+		},
+		Priority: &[]int32{10}[0],
+		FSGroup: secv1.FSGroupStrategyOptions{
+			Type: secv1.FSGroupStrategyRunAsAny,
+		},
+		AllowPrivilegedContainer: true,
+		AllowPrivilegeEscalation: &[]bool{true}[0],
```
stuggi, line 258: Probably bit nicer to read using:

```
trueVar := true
...
               AllowPrivilegeEscalation: &trueVar,
```

dprince: See the new patch I just pushed which includes this suggestion. I also handled the 'priority' in the same way.

PR #27, opened 2020-06-22
Align scc perms in deploy/role.yaml

dprince:
This aligns the SCC permissions in deploy/role.yaml with
what we do for OLM in the csv-generator.

Need to discuss a better way to keep these aligned, possibly with
either shared generators or testing.


dprince:
This is incorrect. we need to use clusterpermissions

PR #61, opened 2022-09-08
Nova CRDs

gibizer:


api/v1beta1/nova_types.go
```diff
@@ -27,12 +30,160 @@ import (
...
+	// APIMessageBusSelector to select the Message Bus Service instance used
+	// by the Nova top level services to communicate. If not provided then
+	// the deployment's default Message Bus instance will be used.
+	APIMessageBusSelector map[string]string `json:"apiMessageBusSelector,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=[]
+	// Cells is a list of cell names pointing to the NovaCell CR instances
+	// representing the cell that belongs to this Nova deployment.
+	Cells []string `json:"cells"`
```
kk7ds, line 50: Am I reading this right that we're going to define cells both by NovaCell CRs and by listing them here? This is probably an ignorant question but it seems like we should be able to look for any NovaCell CRs and use those as the definitive list, no? Why do we need to also list them?

gibizer: This is a good question. I assumed that when a new Nova cell is added to an existing deployment then we might want to have a separate step to deploy it (i.e. create the NovaCell CR) and then a later step to expose it by adding the name of it to this list.

gibizer:
There is a lively discussion with Sean over the design in an etherpad https://etherpad.opendev.org/p/nova-crd-stucture

api/v1beta1/nova_types.go
kk7ds: Hmm, okay. I guess I would have expected that to be a field on the CR (i.e. state=production or something). It seems like we've got these nice complex data structures with labels and such, so we should be able to define "this belongs to that" through those means, without needing a list of cell names that have to be identical to those on the actual NovaCell CR. Like if you imagine CERN with their 90 cells, will have to be listed here with no "foreign key" sort of relationship.

Anyway, I'm not sure what the convention is for things like this, so maybe this makes the most sense to be able to trigger the reconcile (?) when the list changes, I dunno. It just surprised me :)

gibizer: Also good point on triggering a reconcile. We do have to trigger one to add the cell mapping to the API DB for the new cell as the cell itself might not have access to the API DB. If nothing changes on the CR that owns the API DB then that CR will not be reconciled. So this list makes sure that the top level CR knows if there is a new cell to be mapped.

abays: In regards to triggering reconcile, if a top-level CR watches a sub-CR, then changes to the sub-CR will cause the top-level CR to reconcile.  In that specific consideration, the list could be avoided.  So the `Nova`-type or `NovaAPI`-type controller could watch for `NovaCell` CRs and reconcile according when something changes with the latter.

SeanMooney: ok so we shoudl be able to not have a list of cells or compute

have the top level nova watch for cell and compute crd creation and and look for status changes.

e.g. when the compute CR goes to provisined we can have the top level controller run nova-mageage cell_v2 discover_hosts

so in that model deploying a new cell is just done by defining a new cell CR and adding compute to that cell is done by defining the compute cr with reference to the cell and the top level cr will update the API db when the provision of that compute/cell is done by tho cell and compute contolers

SeanMooney:
> There is a lively discussion with Sean over the design in an etherpad https://etherpad.opendev.org/p/nova-crd-stucture

hehe well i would hate for it to be boring :)
I'm glad you feel engaged by the discourse :)

api/v1beta1/nova_types.go
kk7ds: And presumably we could have a `mode=deployment` and `mode=production` flag on the cell CR that lets us decide whether or not we should actually trigger the cell add and discover at the top level.

In order to discover_hosts when a new host gets added, does the top level need to watch for all hosts as well? Or could the cell watch for a new host and increment a counter on itself that then triggers the top level that is just watching cells?

Either way, I think the watching arrangement would be preferable to having to manually provide a list here, so that's cool.

SeanMooney: well the cells could proably watch for host that reference that cell and we could have a status conditional that went to dirty or something when it noticed that an action was required.

and the top-level could look at that
but I'm not sure what the overhead fo the top level looking at the cells and host is vs this two step approach when top level watches for cell and cells watch for hosts.

i like the ideal fo teier level of watches so that one controller does not have to handle events for all compete crs but in practice i don't know if that makes sense in a kubernettes world.

abays: FWIW, we've used the tiered, asynchronous approach with a good deal of success in previous operators.  Each operator/controller has its own CRD magisteria and examines the status or calculated hashes of sub-CRDs whenever it wakes on an event for those sub-CRDs.  Then you just repeat the pattern for the sub-CRD and its sub-sub-CRDs or other child resources (where, in essence, the sub-CRD becomes the top-level CRD in relation to the sub-sub-CRD/resource, etc).  This worked for us, for instance, with OSP-D operator with a chain of `OpenStackControlPlane -> OpenStackVMSet -> KubevirtVM`.

kk7ds: > well the cells could proably watch for host that reference that cell and we could have a status conditional that went to dirty or something when it noticed that an action was required.

Yeah, this is what I meant.. Setting a dirty flag, or just incrementing an index if that's easier.

> i like the ideal fo teier level of watches so that one controller does not have to handle events for all compete crs but in practice i don't know if that makes sense in a kubernettes world.

Presumably there could also be some benefit to not notifying the top level until some cell-level work has been done, which would make the tiered watches better.

stuggi: I think checking the conditions on the cell cr should be used for this. the conditions show the steps, but also overall status of the cell CR. When the cell is full deployed the overall status get set and follow up tasks are good to run to e.g. create the cell via nova-manage.

SeanMooney: im not conviced we should try to report the overall status

i i was discussing this with gibi on irc this morning and i dint thinkthis is the direction we should go but we can discuss that in the meeting later today. 

https://github.com/gibizer/nova-operator/tree/per-service-crds is the other approach we w are considering.  with that we will not have a nova or nova cell CRD

this is described here https://etherpad.opendev.org/p/nova-crd-stucture#L184

api/v1beta1/nova_types.go
```diff
@@ -27,13 +28,81 @@ import (
 type NovaSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	NovaAPI NovaAPI `json:"novaapi,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// APIDBSelector to select the DB Service instance used for the Nova API
+	// DB. If not provided the the default DB Service of the deployment will
+	// be used for the Nova API DB.
+	APIDBSelector map[string]string `json:"apiDBSelector,omitempty"`
```
bogdando, line 36: My proposition is to use gneric DBSelector and MessageBusSelector. Then either it holds an API or a Cell value, might be decided on a defined or not CellName (also to be added into this type). Although I'm not sure how we should then encode both API and Cell values, if there is such a case.

gibizer: There are CRDs where both the APIDB and the CellDB need to be selected independently. So we cannot model this with a single field.

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,127 @@
...
+	// CellDBSelector to select the DB Service instance used for the Nova Cell
+	// DB for this cell. If not provided the the default DB Service of the
+	// deployment will be used for this DB.
+	CellDBSelector map[string]string `json:"cellDBSelector,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// CellMessageBusSelector to select the Message Bus Service instance used
+	// in this Cell to communicate. If not provided then
+	// the deployment's default Message Bus instance will be used.
+	CellMessageBusSelector map[string]string `json:"cellMessageBusSelector,omitempty"`
```
gibizer, line 49: We discussed with Sean over Slack that NovaCell should not create the user and the database for the Cell but let the Nova CR do that. So this needs to be replaced with CellDBSecretSelector and CellMessageBusSecretSelector to only get user, password and the hostname of the already initialized services.

doc/design.md
```diff
@@ -0,0 +1,28 @@
...
+# Decisions
+
+1. The Nova operator provides a single top-level API for the OpenStack umbrella
+operator to instantiate a Nova control plane by creating a single Custom
+Resource (CR). We support this to hide any nova-specific deployment logic from
+the OpenStack umbrella operator.
+
+2. The Nova operator allows deploying every Nova service independently by
+instantiating the matching Nova*ServiceType* CRD without the need to create
+higher-level umbrella CRDs. We support this to limit the dependency of each
```
kk7ds, line 16: In this case does "higher-level umbrella CRDs" refer to the top-level nova one, or the umbrella operator stuff? I assume the former, in which case perhaps you should use a word other than "umbrella" here?

gibizer: Good point. I need to be more explicit. (I blame the too much new terminology. We might need a terminology doc in the doc repo where we agree which word is used for what.)

For the NovaAPI and NovaScheduler CRD the higher level CRD in Nova CRD. We don't want NovaAPI and NovaScheduler to depend on the Nova CRD we only want them to depend on the minimal input (message bus and DB connection info) generated either by the Nova CRD or generated manually.

For the NovaConductor CRD the higher level CRD is the NovaCell CRD right now. Sean suggested to move the Cell database creation to the Nova CRD too. If we do that then real dependency we want to limit would be between Nova CR and NovaCell CR.

kk7ds: Yeah, I was short-circuiting the example for something like NovaAPI where the next level up was Nova just to ask the question. Point being, if we're going to use "umbrella" for the top-top-top-level meta "all of openstack" operator, we should probably not use it to mean "my parent" for clarity :)

But yes, lots of new terminology, which is why I (had to) ask :)

kk7ds: I see you use the term "umbrella" in your Nova CRD commit message (which also has a typo of CRD->CDR btw). Definitely seems like it would be better to avoid that terminology as well.

api/go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/openstack-k8s-operators/nova-operator/api
```
SeanMooney, line 1: so i have no idea how to ask this on the commit message because github sucks at code review
but can you explain why we need this.

i could see us needing it eventually but do we need it now?

api/v1beta1/novacell_types.go
SeanMooney: we also discussed taht the term "umbrella CRD" would be avoided
but yes thanks for capturing it.

i would like to ensure that all other CRDs can be used with the Nova CRD and have the nova CRD do the depency in ject ie. creating the DB instance and passing in the selector.

but i don't think we need to remove CellMessageBusSelector

we just need the password to be passed vai a secrete.
and the respocnablity for creating the rabiit user/vhost should be in the NovaCRD ideally.

so i don't think need to be removed you just need to prove a CellMessageBusSecretSelector which the nova CRD will create and pass in to get the user name and password.

doc/design.md
SeanMooney: yep we should avoid that terminogy and use the concrete name "Nova CRD"

the OpenStack top level operator can just be referred to as the "OpenStack" operator

SeanMooney:
just some addtional top level comments

- we need some selectors for the keystone service
- we need to pass a secret containing the username and password for the nova user and the nova project name/id to the lower level crds. (that user should be created by the Nova CRD)
- we need to decide how we will decide if we enable feature based on if other service are deployed. i.e. do we need to have a reference to a config mapp with a bunch of <service>_enabled keys so that we can enabled per service options like adding cidner if cinder is installed. that is more a cluster wide quequstion but ill need to answer it for the config generation. so ill likely add that in that PR.
- as noted in line your missing the novnc proxy CRD. i think that is top level only today although i know cern wanted to run it per cell which i dont think you can do anymore since we remove nova-consoleauth.

those are the main points i wanted to mention now

we could address those in a separate pr to keep this one short

config/crd/bases/nova.openstack.org_novaschedulers.yaml
```diff
@@ -35,13 +35,163 @@ spec:
...
+                additionalProperties:
+                  type: string
+                description: APIMessageBusSecretSelector - labels to find the Secret
+                  that holds username, password, and hostname of the message bus service
+                  to access the nova (top level) bus
+                type: object
+              containerImage:
+                description: The service specific Container Image URL
                 type: string
+              customServiceConfig:
```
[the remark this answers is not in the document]
abays: This can happen if the CRD struct was changed but `make manifests` was not execute before commiting/pushing the PR.  `make manifests` takes the CRD structs and creates their YAML representations.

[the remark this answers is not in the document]
SeanMooney: ah ok well we need this info so the struct are missing content

that makes sense I'm probably going to try an guard against that by using pre-commit and or the ci in the future so they are always in sysnc but for now ill keep an eye out for this when reviewing.

doc/design.md
gibizer: OK. I will go through the commits and replace umbrella.

api/v1beta1/nova_types.go
```diff
@@ -28,14 +29,80 @@ type NovaSpec struct {
...
+	// ServiceUser - optional username used for this service to register in keystone
+	ServiceUser string `json:"serviceUser"`
+
+	// +kubebuilder:validation:Required
+	// Secret is the name of the Secret instance containing password
+	// information for nova like the keystone service password and DB passwords
+	Secret string `json:"secret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// PasswordSelectors - Selectors to identify the DB and ServiceUser
```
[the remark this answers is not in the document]
gibizer: Probably true. The Secret name and the PasswordSelector passed in to the Nova CRD by the OpenStack operator will need to have admin passwords to each infrastructure services (e.g. DB and Message bus). Then the Nova CRD controller will use that to get the nova specific user/pass pairs registered and only pass those down to the cell and service CRDs.

api/v1beta1/nova_types.go
```diff
@@ -28,14 +29,80 @@ type NovaSpec struct {
...
+	APIService Service `json:"apiService"`
+
+	// +kubebuilder:validation:Required
+	// SchedulerService - define the nova-scheduler service
+	SchedulerService Service `json:"schedulerService"`
+
+	// +kubebuilder:validation:Optional
+	// MetadataService - defines the metadata service that is global for the
+	// deployment serving all the cells.
+	MetadataService NovaMetadata `json:"metadataService"`
```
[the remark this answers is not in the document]
gibizer: Good point. I have to add commits for that too

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// +kubebuilder:default="# add your customization here"
+	// CustomServiceConfig - customize the service config using this parameter to change service defaults,
+	// or overwrite rendered information using raw OpenStack config format. The content gets added to
+	// to /etc/<service>/<service>.conf.d directory as custom.conf file.
+	CustomServiceConfig string `json:"customServiceConfig,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ConfigOverwrite - interface to overwrite default config files like e.g. logging.conf or policy.json.
+	// But can also be used to add additional files. Those get added to the service config dir in /etc/<service> .
+	DefaultConfigOverwrite map[string]string `json:"defaultConfigOverwrite,omitempty"`
```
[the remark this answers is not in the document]
gibizer: So what is your suggestions? Remove this for now and add it back when we see the need or keep it here?

config/crd/kustomization.yaml
```diff
@@ -3,17 +3,32 @@
 # It should be run by config/default
 resources:
 - bases/nova.openstack.org_novaapis.yaml
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
gibizer: /o\ that is a result of a bad conflict resolution. I will fix it. I need to get used to it that some things are generated but also stored in the repo.

api/v1beta1/novascheduler_types.go
```diff
@@ -28,14 +29,41 @@ type NovaSchedulerSpec struct {
...
+	// username, password, and hostname of the message bus service to access
+	// the nova (top level) bus
+	APIMessageBusSecretSelector map[string]string `json:"apiMessageBusSecretSelector"`
+
+	// +kubebuilder:validation:Optional
+	// Debug - enable debug for different deploy stages. If an init container
+	// is used, it runs and the actual action pod gets started with sleep
+	// infinity
+	Debug Debug `json:"debug,omitempty"`
+
```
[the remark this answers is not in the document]
gibizer: Good point.

config/crd/bases/nova.openstack.org_novaschedulers.yaml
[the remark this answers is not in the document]
gibizer: Good point. I have to fix this. I agree to have a pre-commit check that enforces that the generated stuff are in sync with the code.

api/go.mod
gibizer: I thought I put reasoning to the commit message but I can try to be more specific there. The core reason to allow other module (other operator) to depend on only the API structs but not the implementation (the controller) of the operator. This will limit the amount of transitive dependency an operator will gain by such direct dependency on Nova. 
A good existing example is between Placement and Keystone. The placement controller implementation depends on the KeystoneService CRD and uses the KeystoneServiceSpec struct from the keystone-operator repo. But Placement does not need to (and therefore does not want to) depend on the implementation of the Keystone controller. The latter has a lot longer list of dependencies than the former. So keystone-operator was split, the keystone-operator/api module only contains the CRD defining structs, while the keystone-operator module contains the CRD controller implementations.

api/v1beta1/nova_types.go
```diff
@@ -28,14 +29,80 @@ type NovaSpec struct {
...
+
+	// APIServiceReadyCount defines the number or replicas ready from nova-api
+	APIServiceReadyCount int32 `json:"apiServiceReadyCount,omitempty"`
+
+	// SchedulerServiceReadyCount defines the number or replicas ready from nova-scheduler
+	SchedulerServiceReadyCount int32 `json:"schedulerServiceReadyCount,omitempty"`
+
+	// MetadataReadyCount defines the number of replicas ready from
+	// nova-metadata service
+	MetadataServiceReadyCount int32 `json:"metadataServiceReadyCount,omitempty"`
```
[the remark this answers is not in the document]
stuggi: isn't novnc proxy more tight to the cell as we need a vnc proxy per cell?

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,127 @@
...
+	// passwords from the Secret
+	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// ConductorService - defines the cell conductor deployment for the cell
+	ConductorService NovaConductor `json:"conductorService"`
+
+	// +kubebuilder:validation:Optional
+	// MetadataSerivce - defines the metadata serive dedicated for the cell.
+	MetadataSerivce NovaMetadata `json:"metadataService"`
```
stuggi, line 82: should have novnc proxy?

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,127 @@
...
+	// Conditions
+	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
+
+	// ConductorReadyCount defines the number of replicas ready from
+	// nova-conductor service in the cell
+	ConductorServiceReadyCount int32 `json:"conductorServiceReadyCount,omitempty"`
+
+	// MetadataReadyCount defines the number of replicas ready from
+	// nova-metadata service in the cell
+	MetadataServiceReadyCount int32 `json:"metadataServiceReadyCount,omitempty"`
```
stuggi, line 101: for other operators we record the db hostname in the status for easy consumption, e.g. to create the config. should we record the db/mq hostname here as well?

gibizer:
> just some addtional top level comments
> 
>     * we need some selectors for the keystone service

Could you elaborate this a bit. 

To register the nova service and endpoint in keystone the NovaAPI CRD controller will create a KeystoneService CR instance with the nova service and endpoint parameters the keystone-operator will reconcile that to add the service and endpoint to the keystone DB. 

Also the NovaAPI CRD controller will look up the KeystoneAPI CR instance to get the keystone auth URL for the service config. I assume you want to make it configurable which KeystoneAPI CR will be looked up at this point to allow two KeystoneAPI to coexist in the deployment. 

However then we have to make sure then when the NovaAPI CRD controller creates the KeystoneService CR to register the nova service and endpoints, then KeystoneService CR is connected to the specific KeystoneAPI CR somehow.

> 
>     * we need to pass a secret containing the username and password for the nova user and the nova project name/id to the lower level crds. (that user should be created by the Nova CRD)
> 
>     * we need to decide how we will decide if we enable feature based on if other service are deployed. i.e. do we need to have a reference to a config mapp with a bunch of _enabled keys so that we can enabled per service options like adding cidner if cinder is installed. that is more a cluster wide quequstion but ill need to answer it for the config generation. so ill likely add that in that PR.

I agree that this needs to be raised for the whole team. I looked at the existing services but I don't see such connection modeled so far. 
I think we can take two approaches:
- a) auto discovery: the Nova<ServiceType> controllers could watch for the CinderAPI CRD and if that is created the given nova service could enable cinder related features in this config.
- b) receiving this as an input when the Nova<ServiceType>CR is created / updated.
  - b.1) we can have those input via fields on the CR. e.g. CinderEnable bool on the Nova<ServiceType> CR or
  - b.2) we can have those input as a config map as you suggested.

> 
>     * as noted in line your missing the novnc proxy CRD. i think that is top level only today although i know cern wanted to run it per cell which i dont think you can do anymore since we remove nova-consoleauth.

I will add a NovaNoVNCProxy CRD  and create it from the Nova CR level.

> 
> 
> those are the main points i wanted to mention now
> 
> we could address those in a separate pr to keep this one short

api/v1beta1/novascheduler_types.go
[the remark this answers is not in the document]
stuggi: it is part of the Service, do we need another one?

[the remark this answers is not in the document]
gibizer: ahh right. :)

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,88 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
```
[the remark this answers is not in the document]
gibizer: Yeah, it feels pretty common for all operators we have. I will propose a lib-common change. If we land that then I can either update this series to use it from there, or if this series is landed by then, then propose a patch to make the switch.

api/v1beta1/novacell_types.go
gibizer: I intentionally left it our for now to trigger a discussion. As think we can always get the databaseHostname via lib-common querying the DB Service instance via our *DBSelector. So I'm wondering, is it is safe to cache the value here, or it will get out sync. I.e during adoption, switching from the external DB Service instance to the internal one will change the databaseHostname value.

api/v1beta1/novascheduler_types.go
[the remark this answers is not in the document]
SeanMooney: yes we need this per CRD we don't want just one customisation point on the top level CRD there are things that should only be in some config files and not others
we don't want this just on the top level once

api/v1beta1/common_types.go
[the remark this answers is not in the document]
SeanMooney: yes and no. 

you cannot customize policy.json normally
its explicitly not supported.

so we can provide this interface but we should not use that as an example for two reasons.

first policy.json is deprecated we should only use/support policy.yaml
second while we can support this type of customisation for provider.ymal as a supported usecase we should not document the unsupported usecase.

if we need to allow custom policy.yaml under a support exception we can explain how this can be used but i don't want anyone that is using this to be tempted to tweak things

api/v1beta1/novascheduler_types.go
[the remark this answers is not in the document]
stuggi: yes, and its already there via the Service bellow, or am I missing something?

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
SeanMooney: ya that is what i was thinking

api/v1beta1/common_types.go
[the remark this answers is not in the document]
stuggi: agreed that we should not list unsupported files here. it was just the first one I came up.

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
SeanMooney: no you cannot run it at the cell level currently.
when we had the hova console auth service you could but now that we store the console tokens in the API db the vnc proxy service is a top level service only as far as i recall. unless we fixed that again already. cern were complaining about this in one of the more recent ptgs but i don't think we have actually addressed there feedback.

api/v1beta1/novacell_types.go
SeanMooney: i kind of feel like that is an abuse of the status filed and we should not do that and should not be doing that in other operators.

the status stuct should not be used as a proxy datasruyctre for configuration info.

if we want to have the host name it should be in a config map that we pass down.
otherwise it should be a selector.

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
stuggi: from https://review.opendev.org/c/openstack/devstack/+/566446/ we need a vncproxy per cell

bogdando:
> * we need to decide how we will decide if we enable feature based on if other service are deployed. i.e. do we need to have a reference to a config mapp with a bunch of _enabled keys so that we can enabled per service options like adding cidner if cinder is installed. that is more a cluster wide quequstion but ill need to answer it for the config generation. so ill likely add that in that PR.

we can detect deployed services via selectors, no need to list _enabled keys, nor generate it by the openstack meta operator

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
SeanMooney: hum i tought we did this the other way around.

how do we know which console auth proxy instance to loadbalnace too.
i guess that would mean we need a separate k8s service per cell per vnc proxy and 
we ould need to use a different cell specific base url in the config of each vnc proxy CR

perhaps cern were unhappy that they cant run it centrally anymore and i got that revered.

ok kolla also seams to deploying it at the cell level so yes this should be added to the cell

api/v1beta1/novascheduler_types.go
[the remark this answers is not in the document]
gibizer: @SeanMooney it is in the Service struct the scheduler gets via its Service Service field.

config/crd/kustomization.yaml
```diff
@@ -2,21 +2,33 @@
...
-- bases/nova.openstack.org_nova.yaml
 - bases/nova.openstack.org_novaapis.yaml
+- bases/nova.openstack.org_novaschedulers.yaml
+- bases/nova.openstack.org_novaconductors.yaml
+- bases/nova.openstack.org_novametadata.yaml
+- bases/nova.openstack.org_novacells.yaml
+- bases/nova.openstack.org_nova.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
 patchesStrategicMerge:
```
bogdando, line 13: unrelated: I have a warning here:
Incorrect type. Expected "array".yaml-schema: kustomization.yaml
(see https://json.schemastore.org/kustomization.json)

I think that should default to [] to make it happy

api/v1beta1/common_types.go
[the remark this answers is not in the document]
gibizer: I've proposed the lib-common change https://github.com/openstack-k8s-operators/lib-common/pull/58 but it might not be accepted as is. So I will keep the struct on the nova side now.

api/v1beta1/novacell_types.go
gibizer: looking at the novncproxy code and the way we test it in zuul I think you are right we need a novncproxy per cell https://zuul.opendev.org/t/openstack/build/3b72907be6ae4cb69c4ef0d3ca40bbbc/logs

config/crd/kustomization.yaml
gibizer: this is generated code but I have some merge conflict issues commented above. I hope if that is resolved then this will go away too.

config/crd/bases/nova.openstack.org_novaapis.yaml
```diff
@@ -34,9 +34,194 @@ spec:
...
+                type: object
+              containerImage:
+                description: The service specific Container Image URL
+                type: string
+              customServiceConfig:
+                default: '# add your customization here'
+                description: CustomServiceConfig - customize the service config using
+                  this parameter to change service defaults, or overwrite rendered
+                  information using raw OpenStack config format. The content gets
+                  added to to /etc/<service>/<service>.conf.d directory as custom.conf
```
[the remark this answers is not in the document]
gibizer: sure. go for it.

SeanMooney:
> > * we need to decide how we will decide if we enable feature based on if other service are deployed. i.e. do we need to have a reference to a config mapp with a bunch of _enabled keys so that we can enabled per service options like adding cidner if cinder is installed. that is more a cluster wide quequstion but ill need to answer it for the config generation. so ill likely add that in that PR.
> 
> we can detect deployed services via selectors, no need to list _enabled keys, nor generate it by the openstack meta operator

that would require use to regenerate config every time a new operator is deployed

for example if you deployed without cidner then added it we would have to update every config in nova to have the cinder section including on the compute nodes.

i was hoping to void that but we could use the selector if and only if the name of the sector is fixed or passed into us.

however for thing like cinder we do not need the selector in nova we never need to know the URL for cinder as that is discover by nova (not the operator) dynamically form the keystone API endpoint records so depending on the service selectors for discover is not really a good solution in my view

SeanMooney:
> > just some addtional top level comments
> > ```
> > * we need some selectors for the keystone service
> > ```
> 
> Could you elaborate this a bit.
sure i need the keystone service selector to generate the keystone auth URL in the nova.conf
so the nova python binary can auto discover the URLs to all the other service via the keystone service catalog.

so i don't want to assume the at the keysont service is called keystone and hardcode that
as such it would be good to have that as an optional input to the NovaCRD
we can default it to whatever the default name will be so its not required but i would like that to a parameter that all.


> 
> To register the nova service and endpoint in keystone the NovaAPI CRD controller will create a KeystoneService CR instance with the nova service and endpoint parameters the keystone-operator will reconcile that to add the service and endpoint to the keystone DB.
> 
> Also the NovaAPI CRD controller will look up the KeystoneAPI CR instance to get the keystone auth URL for the service config. I assume you want to make it configurable which KeystoneAPI CR will be looked up at this point to allow two KeystoneAPI to coexist in the deployment.
> 
> However then we have to make sure then when the NovaAPI CRD controller creates the KeystoneService CR to register the nova service and endpoints, then KeystoneService CR is connected to the specific KeystoneAPI CR somehow.
> 
> > ```
> > * we need to pass a secret containing the username and password for the nova user and the nova project name/id to the lower level crds. (that user should be created by the Nova CRD)
> > 
> > * we need to decide how we will decide if we enable feature based on if other service are deployed. i.e. do we need to have a reference to a config mapp with a bunch of _enabled keys so that we can enabled per service options like adding cidner if cinder is installed. that is more a cluster wide quequstion but ill need to answer it for the config generation. so ill likely add that in that PR.
> > ```
> 
> I agree that this needs to be raised for the whole team. I looked at the existing services but I don't see such connection modeled so far. I think we can take two approaches:
> 
>     * a) auto discovery: the Nova controllers could watch for the CinderAPI CRD and if that is created the given nova service could enable cinder related features in this config.
> 
>     * b) receiving this as an input when the NovaCR is created / updated.
>       
>       * b.1) we can have those input via fields on the CR. e.g. CinderEnable bool on the Nova CR or
>       * b.2) we can have those input as a config map as you suggested.
> 

ya i considered autodiscovery my concern there is needing to do thing several times
as different operator reconsile

i would prefer either of option b to avoid that but as you said this is something we need to decide at the top level.

perhaps we just need a reference to the meta operator cr?
it can have those top level "enable_*" fields and all the service operator can check that?


> 
> > ```
> > * as noted in line your missing the novnc proxy CRD. i think that is top level only today although i know cern wanted to run it per cell which i dont think you can do anymore since we remove nova-consoleauth.
> > ```
> 
> I will add a NovaNoVNCProxy CRD and create it from the Nova CR level.
> 

ack

api/v1beta1/nova_types.go
```diff
@@ -27,13 +28,81 @@ import (
...
+
+	// +kubebuilder:validation:Optional
+	// Debug - enable debug for different deploy stages. If an init container
+	// is used, it runs and the actual action pod gets started with sleep
+	// infinity
+	Debug Debug `json:"debug,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// APIService - define the nova-api service
+	APIService Service `json:"apiService"`
```
gibizer, line 74: This should be APIService NovaAPISpec and all the rest below too should point to the Spec Stuct

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,127 @@
...
+	Secret string `json:"secret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// PasswordSelectors - Selectors to identify the DB and ServiceUser
+	// passwords from the Secret
+	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// ConductorService - defines the cell conductor deployment for the cell
+	ConductorService NovaConductor `json:"conductorService"`
```
gibizer, line 78: This should be ```ConductorService NovaConductorSpec``` we don't need the Status parts of the CR here. Same true below for the Metadata

api/v1beta1/novacell_types.go
gibizer: done

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
gibizer: done

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,83 @@
...
+// Service contains the fields that are needed for each nova service CRD
+type Service struct {
+	// +kubebuilder:validation:Required
+	// The service specific Container Image URL
+	ContainerImage string `json:"containerImage"`
```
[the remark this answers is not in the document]
gibizer: ack

config/crd/kustomization.yaml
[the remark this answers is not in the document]
gibizer: Done

api/go.mod
gibizer: done

api/v1beta1/nova_types.go
gibizer: done

api/v1beta1/novacell_types.go
gibizer: Done

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
gibizer: Done

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
gibizer: Added on the cell level.

api/v1beta1/novacell_types.go
gibizer: Done.

gibizer:
I think I handled all the major concerns.

Notes to the reviewers:
* There are now multiple examples for NovaAPI, NovaCell and Nova CRs in yaml format. These are not tested yet so I'm sure there are typos there. But I added them to show how the CRs will be used from different perspectives (i.e. from top, or as individual service CR)
* I removed the multiple Secret selectors. There is one Secret passed by name to every CR, and there are field selectors (PasswordSelector) to instruct a given CR which field from that Secret is relevant for it. This allows using one single big Secret for easy testing but also allows per service level Secret usage for security.
The user names and hostnames are not in the secret today so I did not add field selectors for them. Instead I passed them as individual fields to lower level CRs. See the different Slack discussions from Friday for more details.
* There are various fields that is now Optional but should be Required like KeystoneAuthURL. I noted inline hpw we can resolve this but I intentionally not resolved the issue yet as first I need more experience to be able decide what is the good resolution. Right now I tend towards splitting XXXSpec and XXXTemplate as two separate structs with a possible common base.
* I still believe that having a separate SuperConductor and not representing cell0 as individual NovaCell CR would result in a simpler API. But for the sake of progress I followed the suggestions to have cell0 as a NovaCell and represent the SuperConductor as NovaConductor in NovaCell.CellName=cell0.

enjoy :) and thank you for the effort reading through all these commits.

config/samples/nova_v1beta1_nova.yaml
```diff
@@ -1,6 +1,70 @@
 apiVersion: nova.openstack.org/v1beta1
 kind: Nova
 metadata:
-  name: nova-sample
+  name: nova-minimal-example
```
gibizer, line 4: This example is now passing basic API validation. You can try it with https://github.com/openstack-k8s-operators/install_yamls/pull/27

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -0,0 +1,410 @@
+apiVersion: nova.openstack.org/v1beta1
+kind: Nova
+metadata:
+  name: nova-multi-cell-sample
```
gibizer, line 4: This is now passing basic API validation. You can try it with https://github.com/openstack-k8s-operators/install_yamls/pull/27 and 
```
export NOVA_REPO=https://github.com/gibizer/nova-operator.git
export NOVA_BRANCH=nova-top-level-crd
export NOVA_IMG=quay.io/gibi/nova-operator-index:v0.0.2
export NOVA=config/samples/nova_v1beta1_nova-multi-cell.yaml
make nova
make nova_deploy
```

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,183 @@
...
+	//    both NovaCellSpec and NovaCellTemplate
+	// c) add a validating webhook here that runs only when NovaCell CR is
+	//    created and does not run when Nova CR is created and make this field
+	//    required via that webhook.
+	KeystoneAuthURL string `json:"keystoneAuthURL"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=nova
+	// APIDatabaseUser - username to use when accessing the API DB
+	APIDatabaseUser string `json:"apiDatabaseUser"`
```
kk7ds, line 75: For some reason I thought that we talked about putting the username in the secret as well. There has been a lot of discussion, so perhaps I'm mis-remembering. Seems to me like it makes sense to put "credentials" in the secret and not just the password. Is there any reason not to?

The username is almost as sensitive as the password and it makes the secret be "how we authenticate" instead of just "all our passwords". Anyway, this is not important, I just thought I'd mention it here.

api/v1beta1/novacell_types.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// +kubebuilder:validation:Optional
+	// MetadataServiceTemplate - defines the metadata serive dedicated for the cell.
+	MetadataServiceTemplate NovaMetadataSpec `json:"metadataServiceTemplate"`
+
+	// +kubebuilder:validation:Optional
+	// NoVNCProxyServiceTemplate - defines the novvncproxy serive dedicated for
+	// the cell.
+	NoVNCProxyServiceTemplate NovaNoVNCProxySpec `json:"noVNCProxyServiceTemplate"`
+}
```
kk7ds, line 135: Okay, so no NovaAPI or NovaScheduler Spec(s) in here, because we're going to consider API to not be in any cell for the time being, right?

api/v1beta1/novaconductor_types.go
```diff
@@ -0,0 +1,162 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NovaConductorSpec defines the desired state of NovaConductor
+type NovaConductorSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +kubebuilder:validation:Required
+	// CellName is the name of the Nova Cell this conductor belongs to.
+	CellName string `json:"cellName,omitempty"`
```
kk7ds, line 34: Do we still need the cell name attached here? If I understand this correctly, these will (generally) be created by the NovaCell, so it will just be duplicated information that one piece of code is setting on another, right? Is there some benefit here for the case where you create a NovaConductor manually?

api/v1beta1/novaconductor_types.go
```diff
@@ -0,0 +1,162 @@
...
+	// ServiceUser - optional username used for this service to register in
+	// keystone
+	ServiceUser string `json:"serviceUser"`
+
+	// +kubebuilder:validation:Optional
+	// KeystoneAuthURL - the URL that the nova-conductor service can use to
+	// talk to keystone
+	// NOTE(gibi): This is made optional here to allow reusing the
+	// NovaConductorSpec struct in the Nova CR via the NovaCell CR where this
+	// information is not yet known. We could make this required via multiple
```
kk7ds, line 57: When or why would it not be known? Is there some ordering chicken-and-egg reason for this?

[Later] .. Oh is this because we want to use this data structure in the top-level thing where you're just saying "I want some conductors but I don't know all the details", which are later resolved by Nova and filled in here for the actual creation? I guess I'm not sure why you'd need to specify these things in the top-level CR. If you want "give me N conductors in this cell" then this should all be implied, and if you want to deploy a specific conductor with specific details, you can do it manually right?

That leads me back up to the novacell_types.go, where I'd say: Do we need a full NovaConductor spec in there, vs just "give me N conductors in this cell"? I *think* that conductors and metadatas within one cell would pretty much always be configured identically, right?

api/v1beta1/novacell_types.go
bogdando: Changing API version would allow us to add it into (a copy?) of the struct that corresponds that API change

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -0,0 +1,410 @@
...
+          stopService: False
+          preserveJob: False
+      metadataServiceTemplate:
+        cellName: cell2
+        secret: osp-secret
+        passwordSelectors:
+          service: NovaPassword
+          # We don't provider the API DB password selector here as this
+          # nova-metadata service is expected to run within cell2.
+          apiDatabase: null
```
bogdando, line 343: is this the same if having it omitted?

api/v1beta1/novacell_types.go
gibizer: No, you are remembering correctly we talked about it. I will start a discussion around it on Slack as currently every existing operators expecting just the password from the secret so this probably needs a wider change if we want to use the same tooling for dev testing (i.e. install_yamls). What about the service (DB, rabbit) hostname? If that is also considered sensitive and needs to be in a secret then we need to pass around multiple secrets as user/password input comes outside of the cluster but hostnames are generated inside and I if I understand correctly the controller code should not change existing secrets. Anyhow it worth discussing it.

api/v1beta1/novacell_types.go
gibizer: @kk7ds: Yes, they is here as they are not part of any cell (cell0) in the current proposal.
@bogdando : as far as I know in v1beta1 we are allowed to change the API even in backward incompatible way.

api/v1beta1/novaconductor_types.go
gibizer: Wondering how we can differentiate between a cell0 super conductor and a cell conductor that has upcall support and hence it has API DB and MQ access.

api/v1beta1/novaconductor_types.go
gibizer: in Nova CR (and NovaCell) we need all input that the Nova CR controller cannot invent itself (e.g.: it can query the db hostname via the DB Service object but it cannot invent NovaConductor ContainerImage or ConfigOverwrite). So we don't need a full NovaConductorSpec in NovaSpec, but we need like 70% of it. As I noted in the comment we can create a separate struct, lets name it NovaConductorTemplate, that only contains the fields Nova CR need from outside to create a NovaConductor CR. And keep that separate from NovaConductorSpec. This will create a bit more repetition but we can make the KeystoneAuthURL and hostname fields mandatory in NovaConductorSpec and absent in NovaConductorTemplate. More I think about it more I feel that such split will avoid some future confusion. 

Note that this would not lead to an as simple NovaConductorTemplate as you hinted with "give me N conductors in this cell" though :)

config/samples/nova_v1beta1_nova-multi-cell.yaml
gibizer: yes and no :) (sorry) So apiDatabase field is defaulted in the PasswordSelector struct to NovaAPIDatabasePassword. So if it is omitted then the default will be used, which might or might not a valid field of the Secret passed to this CR. In install_yamls we use one big Secret holding all the password passed to every CRD. In that case the default would happen to be valid and this metadata service would have access to the password to the API DB.

I assume that null means this will have a nil value in the controller code during reconciliation so it can suggest that this metadata service has no access to the API DB password even if the Secret has field with that password. Only "suggest" as the controller can still read out all the fields from the Secret if it wants. 

This is not super nice. But with a given approach that we pass to a single Secret to a CR with not hardcoded field names, I think we cannot avoid this shenanigans. (And the current approach has benefits as it support both a single big Secret and multiple small per CRD Secrets to work seamlessly)

api/v1beta1/novacell_types.go
SeanMooney: the rabbit and db hostname would be discovered form the service via the selector at the top level CRD i would probably pass the same selector to the other CRDs personally and only get the hostname when we need it when generating the config but if we want to do that in the top level nova CRD and pass it via the secret that would be fine by me.

we could pass the hostname to in the nova-api CR provided the top level nova controller will update them if the service changes

api/v1beta1/novacell_types.go
SeanMooney: i dont want to consider any of our apis as stable at this point

we can after we tag a 1.0 release of the operator but i would not expect us to do that for some time now.

once we have 1.0 we can apply the normal api stability critia to when we do a api version bump.

until then we should not. worry about that as much as we will in the future.

i would consider all our current apis as alpha by the way
but did not think we want to get into that level of specific now considering we don't have testing in place to enforce this.

api/v1beta1/novaconductor_types.go
SeanMooney: this should proably be a selector for the cell CR that it belongs too.
if we are keeping this filed.

but to gibis question, we can likely make that decision based on if the APIdb field is populated.

if we popluplate we can enable the upcall if not we can disable it.

api/v1beta1/novaconductor_types.go
SeanMooney: i feel like we will need to continue to evolve this interface.

the service CRDs will need to have a way (selector, configmap,secret or direct field) to pass the info required to generate the service config.

i would prefer to pass the user and passwords for via a secret and have a credential selector as dan suggested before.

so i would expect the top-level nova controller to create secreates for each top level service CR and likely the Cell CRs
the cell CR could  should create secrets for each of the cell services.

if we go with putting these fields directly in the CRD then instead of the cell controller creating the secret and passing the selector it would just populate the fields.

kk7ds: > So we don't need a full NovaConductorSpec in NovaSpec, but we need like 70% of it

I guess this is a dumb question, but why do we need 70% of it in the top-level Nova CR? Of the things that are here, I think NovaCell knows about all of them:

CellName, Secret, ServiceUser, KeystonAuthURL, APIDatabaseUser, APIDatabaseHostname, APIMessageBusUser, APIMessageBusHostname, CellDatabaseUser, CellDatabaseHostname, CellMessageBusUser,CellMessageBusHostname.

The only thing I'm not sure of is PasswordSelector as I don't fully understand what that means or is used for, but I assume it falls in the same camp.

So, why would I not be able to say, in the top-level Nova CR, "Here's a NovaCell spec with details and it should have N conductors" ?

> the service CRDs will need to have a way (selector, configmap,secret or direct field) to pass the info required to generate the service config.

I realize that NovaConductor should have these fields, but do we need to be allowed to *specify* those things in the Nova CR?

api/v1beta1/novacell_types.go
kk7ds: Yeah, having the hostname outside the secret is fine, but it seems like the username and password are roughly equivalent in terms of security-sensitivity and instead of passing a username and a ref to a secret, it seems like keeping the username there as well makes plenty of sense.

config/samples/nova_v1beta1_nova.yaml
SeanMooney: i think this minimal exampel is still far to verbose

the following is what i would consider a minimal example of a default env
with default image and replicas 

kind: Nova
kind: Nova
metadata:
metadata:
  name: nova-sample
  name: nova-minimal-example


that should result in a central control plane with its own API db on a separate maradb server. with a cell 0 and cell 1 with a super conductor and metadata service deploy on the cells.

if you wanted to use a different "stable/zed" container tag for all service container we should add

OpenStackTag:stable/zed

if we wanted to do that just for the apiSertice then we would added 
  apiServiceTemplate:
    containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:stable/zed

SeanMooney (commented):
parts of this i like but much of it i think needs work.

I'm kind of torn between merging this as is so we can iterate on indiviutal parts in smaller chunks and waiting till we have better agreement on the CRD design.

as it stands this is not the Nova CRD interface i think we should be exposing its far to pourous.

if you want that level of granualtrity you should not use the nova CRD
The reason the Nova CRD exists is to expose a narrow contract to the cluster operator which this does not currently provide.

kk7ds:
> I'm kind of torn between merging this as is so we can iterate on indiviutal parts in smaller chunks and waiting till we have better agreement on the CRD design.

Yeah, I'm okay with merging this as it is (or close to it) and iterating from there, if that's what we want to do. Because this is forming what will be a versioned API in the long-run, my gut reaction is to be ultra-concerned about what it looks like during review like this. But, this is a long way from being released and thus locked in, so it probably makes sense to iterate on smaller chunks of this in the short term.

api/v1beta1/novaconductor_types.go
gibizer: Both cell0 and a normal cell with upcall support will have API DB. So my wondering is about how we differentiate between cell0 and a normal cell with upcall support.

Also we said we want that lower level CRs are not depending on higher level CRs, so having a selector here in NovaConductor for a NovaCell would revers that dependency.

api/v1beta1/novaconductor_types.go
gibizer: > i feel like we will need to continue to evolve this interface.

I agree.

> 
> the service CRDs will need to have a way (selector, configmap,secret or direct field) to pass the info required to generate the service config.

I agree. I use direct fields now, except for the passwords.

> 
> i would prefer to pass the user and passwords for via a secret and have a credential selector as dan suggested before.

As I said above we need a wider agreement about putting user names in secrets as the current operator pattern and many existing operators does not do that and these Secrets are expected to come from outside (from the deployer) so we need some level of consistency in the user interface. 

Also we need to figure out if we want to put the DB/MQ/Keystone URLs to Secrets. Those are not coming from outside so they cannot be in the same Secret as the passwords. I think we can use plain fields for these hostname as they are already available in the kubernetes API.

I started a slack thread for this: https://coreos.slack.com/archives/CQXJFGMK6/p1663660928278619

> 
> so i would expect the top-level nova controller to create secreates for each top level service CR and likely the Cell CRs the cell CR could should create secrets for each of the cell services.

Right now the proposal already prescribes that the NovaCell CR and the Nova<SeviceType> CRs get a Secret. Also Nova CR gets a Secret. So the API allows what you are after. 

However we have to agree on how to use Secrets.

A) Secret + field selector. This is the pattern used today. A CRD controller need to assume nothing about the structure of the Secret as it gets not just a Secret but also field names (PasswordSelector struct today) that defines where (at which field) a needed input (e.g. the API DB password) stored in that Secret.

B) Just a Secret with predefined (hardcoded) fields. E.g. we document that the NovaCell CR expects a Secret that always have a CellDatabasePassword field.

The A) is more flexible than B) as it allows storing the Cell1DatabasePassword and Cell2DatabasePassword in the same Secret if somebody wish to manages less Secret objects. So personally I'm happy with A) and this is how the current CRDs in this PR is designed.

> 
> if we go with putting these fields directly in the CRD then instead of the cell controller creating the secret and passing the selector it would just populate the fields.

Yes, this is the current proposal for usernames, keystoneAuthURL, and hostnames. Plain fields are less complex to handle than a Secret selector that we need to use to query a Secret and then a watch on that Secret to detect changes. So in the first solution I would like to use plain fields where ti is possible (I agree to store passwords and maybe usernames in Secret, depending on the general agreement on that).

gibizer: > > So we don't need a full NovaConductorSpec in NovaSpec, but we need like 70% of it
> 
> I guess this is a dumb question, but why do we need 70% of it in the top-level Nova CR? Of the things that are here, I think NovaCell knows about all of them:
> 
> CellName, Secret, ServiceUser, KeystonAuthURL, APIDatabaseUser, APIDatabaseHostname, APIMessageBusUser, APIMessageBusHostname, CellDatabaseUser, CellDatabaseHostname, CellMessageBusUser,CellMessageBusHostname.
> 
> The only thing I'm not sure of is PasswordSelector as I don't fully understand what that means or is used for, but I assume it falls in the same camp.
> 
> So, why would I not be able to say, in the top-level Nova CR, "Here's a NovaCell spec with details and it should have N conductors" ?
> 
> > the service CRDs will need to have a way (selector, configmap,secret or direct field) to pass the info required to generate the service config.
> 
> I realize that NovaConductor should have these fields, but do we need to be allowed to _specify_ those things in the Nova CR?

If something is allowed to be configured on a NovaConductor CR externally then that should be allowed to be configured via creating a Nova CR. Otherwise the whole top down management approach of the deployment would not work as after the OpenStack operator created a NovaConductor (indirectly) based on the input in the Nova CR the deployer engineer need to go and fine tune that NovaConductor CR instance with the configuration that was not possible via the Nova CR. My understanding of  the overall strategy is that we want to allow managing the deployment via the OpenStack operator.

I see the following configuration that needs to be allowed on a NovaConductor CR from outside (and hence needs to be allowed via the Nova CR too):
* replica count
* container image
* resource limits
* node selector (placement constraints of the replicas)
* config override of the default config of the nova-conductor binary (CustomServiceConfig, DefaultConfigOverwrite)
* username and password for the cell and optionally to API DB and MQ (this is now the username fields, the Secret field and the password selector)
* username and password for the keystone service
* debug options for the operator controller

Things that cannot be configured on a NovaConductor CR from outside(and hence should not be exposed on the Nova CR):
* hostnames of the DB, MQ and Keystone services

This is where my 70% estimate comes from. And looking at the list now I think it is more like 85% :)

gibizer: > no the nova cr will need to create the db instance and rabbit instance i think and then pass the db/rabbit user/host/password info to the CRs it create
> 

I agree. This is how this PR proposes I believe.

> those are not field we would expect to be set or exposed via the top level Nova CRD.

DB / MQ / keystone hostnames are not exposed on the Nova CRD in this PR. The Nova CR will calculate those based on the Service selectors. 

> 
> the caveat to that is if we want to allow the nova CRD to use precreated server db/rabbit instance for adoption we need to allow disabling the creation fo these and resuing exsitng instance which will be placed behind a k8s service
> 
> where the current conflict arises in my mind is the desire to have operators use the CRD interface to define the db schmea instance and create the operator user.
> 
> that wont work for external rabbit or db so the nova operator need to cater for that but non of the lower level crds like the conductor should know about any of that.

I don't see this conflict.
The DB and MQ are selected via k8s Service objects. So for an external DB and MQ an externally mapped k8s Service object can be configured and the Nova CR can also be configured to select those externally mapped k8s Service objects. 

Creating the DB user / password and schema and creating the MQ user / password should be idempotent. I.e. when a DB is adopted the Nova CR can run user and password creation on it and can run nova-manage db sync on it, both should be noop at that point. (Same for the MQ user / pass)


> 
> so in my view the nova crd should not be a direct passthrough of 70% of the conductor CR.

See my above answer to @kk7ds. It is all about our strategy that in my understanding states that the whole nova control plane should be configurable via a Nova CR instance. 

> 
> if we are deciding that the nova top-level controller is respocnable for creating the db and messagebus instances or resuing exsitng ones then none of the field related to that need to be exposed in the novaCRD representation fo the conductor.
> 
> that is all internal.
> 
> so i would hope that 99% of the conductor CRD will be absent in the top level Nova crd.

Could you please create an example where the Nova CR has no details about the NovaCell and NovaConductor serivces but still allows fine tuning details on the conductors. I believe what is exposed today on Nova CR is what needed to be configured on a NovaConductor by a deployer.

api/v1beta1/novaconductor_types.go
```diff
@@ -0,0 +1,162 @@
...
+	// options:
+	// a) create a NovaConductorTemplate that duplicates NovaConductorSpec
+	//    without this field. Use NovaCondcutorTemplate in NovaCellSpec.
+	// b) do a) but pull out a the fields to a base struct that are used in
+	//    both NovaConductorSpec and NovaCondcutorTemplate
+	// c) add a validating webhook here that runs only when NovaConductor CR is
```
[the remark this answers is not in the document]
gibizer: Here we can avoid it. 
I think we cannot avoid validation webhooks in other cases where we need special casing NovaCell for cell0, NovaConductor for super conductor, or metadata service to support both top level and cell level.

gibizer:
> parts of this i like but much of it i think needs work.
> 
> I'm kind of torn between merging this as is so we can iterate on indiviutal parts in smaller chunks and waiting till we have better agreement on the CRD design.
> 
> as it stands this is not the Nova CRD interface i think we should be exposing its far to pourous.

I reached the point where I would like to ask for a counter proposal of a Nova CRD that does satisfy both requirements
* allow configuring the Nova control plane purely via the Nova CR
* not porous 

I think these are conflicting requirements.

> 
> if you want that level of granualtrity you should not use the nova CRD The reason the Nova CRD exists is to expose a narrow contract to the cluster operator which this does not currently provide.

I think this is about strategy. As far as I understand we want to allow configuring everything via the Nova CR in a way that the deployer never needs to touch lower level CRDs.

api/v1beta1/novaapi_types.go
```diff
@@ -28,14 +30,90 @@ type NovaAPISpec struct {
...
+	//    both NovaAPISpec and NovaAPITemplate
+	// c) add a validating webhook here that runs only when NovaAPI CR is
+	//    created and does not run when Nova CR is created and make this field
+	//    required via that webhook.
+	KeystoneAuthURL string `json:"keystoneAuthURL"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=nova
+	// APIDatabaseUser - username to use when accessing the API DB
+	APIDatabaseUser string `json:"apiDatabaseUser"`
```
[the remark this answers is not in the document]
stuggi: > as noted in other place i agree with dan that the user should also be in the secrete alternatively we should just take the db URL in the API via a secrete and have the top level nova controller render it.
> 
> it really a question of do we want to render the URL in the config template or in go in the nova Contorler.
> 
> the same applies to the message bus

we could do one or the other, but I think we should do such rendering in a single place, not something in the controller and another setting being rendered in the template.

api/v1beta1/novacell_types.go
stuggi: I am not against it, just some points we need to think about. would the operator fail/wait if the user is not provided or default in the controller? with not using the param, we loose to default the service user in the way we do it now, so its a required input. defaulting in the controller would make it harder to validate via a validation webhook that the user won't change after initial create, or we have to keep track of it in a different place and validate against that.

api/v1beta1/novacell_types.go
stuggi: +1 on Sean's comment on api stability

api/v1beta1/novacell_types.go
stuggi: I don't think it gets out of sync as during each reconcile it gets checked, but I am ok not to have it in the status. it can make live easier when as we then have the information in the object of the helper, but could be done different as well.

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,107 @@
...
+// Debug allows enabling different debug option for the operator
+type Debug struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// StopDBSync allows stopping the init container before running db sync
+	// to apply the DB schema
+	// QUESTION(gibi): Not all CR will run dbsync, should we have per CR
+	// Debug struct or keep this generic one and ignore fields in the
+	// controller that are not applicable
+	StopDBSync bool `json:"stopDBSync,omitempty"`
```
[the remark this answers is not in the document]
gibizer: This is coming from the devs who already implemented several operators so I trust their judgement that this is useful for development. What if we keep it here while we implement the majority of the nova operator, gather the experience to see if this is useful or not, and remove it before we release the API if we feel this is not something we need?

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// StopService allows stopping the service container before staring
+	// the openstack service binary
+	// QUESTION(gibi): Not all CR will run a service, should we have per CR
+	// Debug struct or keep this generic one and ignore fields in the
+	// controller that are not applicable
+	StopService bool `json:"stopService,omitempty"`
```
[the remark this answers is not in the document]
gibizer: see above

api/v1beta1/novaconductor_types.go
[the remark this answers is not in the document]
stuggi: I don't think we can get around webhooks. might not need all the types, but when it comes to api changes with a new api version there is the conversion webhook to migrate existing objects to the new version. the validation webhook is also one we will see. maybe not in nova but in other operators

api/v1beta1/common_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// PreserveJobs - do not delete jobs after they finished e.g. to check logs
+	PreserveJobs bool `json:"preserveJobs,omitempty"`
+}
+
+// PasswordSelector to identify the DB and AdminUser password from the Secret
+type PasswordSelector struct {
```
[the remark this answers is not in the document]
gibizer: Both the Secret and the field of the Secret is configurable. This approach has the benefit to not prescribe how many Secrets the deployer wants to manage. It allows creating one Secret for the whole deployment without the problem of clashing field names from different operators. Again this is en external UX question. So I'm pulling this up to a Slack thread: https://coreos.slack.com/archives/CQXJFGMK6/p1663667119947399

api/v1beta1/novaapi_types.go
```diff
@@ -28,14 +30,90 @@ type NovaAPISpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Foo is an example field of NovaAPI. Edit novaapi_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// +kubebuilder:validation:Required
+	// Secret is the name of the Secret instance containing password
+	// information for the nova-api sevice.
+	Secret string `json:"secret,omitempty"`
```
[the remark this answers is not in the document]
gibizer: There are small differences but I agree that both could work for us.

- passing object by name: 
    - simpler but more rigid interface
    - uniquely identify 0 or 1 object in the namespace
    - if the value of the name is changes then that clearly trigger a reconcile on the CR
 - passing a map of label selectors
     - more complex but more flexible interface
     - need to handle cases when more than one object matches the labels
     - the controller needs to autonomously re-execute the label query to see if the result is changed and a different object should be used. (Can we have a label query based watch?)

api/v1beta1/novaapi_types.go
```diff
@@ -28,14 +30,90 @@ type NovaAPISpec struct {
...
+
+	// +kubebuilder:validation:Optional
+	// PasswordSelectors - Field names to identify the passwords from the
+	// Secret
+	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
```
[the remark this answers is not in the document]
gibizer: this is UX question so pulled this out to Slack

config/samples/nova_v1beta1_novaapi.yaml
```diff
@@ -3,4 +3,39 @@ kind: NovaAPI
 metadata:
   name: novaapi-sample
 spec:
-  # TODO(user): Add fields here
+  secret: osp-secret
```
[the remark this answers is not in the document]
gibizer: The current proposal is flexible enough to allow using one single Secret holding the passwords for the whole control plane. but also allows passing in a smaller scale Secret if we want. The example is using the secret name as an example here. It is intentionally matching the install_yamls default secret name to show that it is possible to use this example with the install_yamls tooling.

config/samples/nova_v1beta1_novaapi.yaml
```diff
@@ -3,4 +3,39 @@ kind: NovaAPI
...
+  passwordSelectors:
```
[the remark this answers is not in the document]
gibizer: move to slack

config/samples/nova_v1beta1_novaapi.yaml
```diff
@@ -3,4 +3,39 @@ kind: NovaAPI
...
+    service: NovaPassword
+    apiDatabase: NovaAPIDatabasePassword
+    apiMessageBus: NovaAPIMessageBusPassword
+    cellDatabase: null
+    cellMessageBus: null
+  serviceUser: nova
+  keystoneAuthURL: http://keystone-internal-openstack.apps-crc.testing
+  apiDatabaseUser: nova
+  # This is the hostname of the default DB today
+  apiDatabaseHostname: openstack
```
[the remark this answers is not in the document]
gibizer: This is an example. I intentionally proposed the values here to matching with the install_yamls tooling so I will be able to validate the example there.  Feel free to propose another example file that shows the multi DB service approach of deploying nova. That would be a useful example showing the flexibility of our CRD model

config/samples/nova_v1beta1_novaapi.yaml
```diff
@@ -3,4 +3,39 @@ kind: NovaAPI
...
+  apiMessageBusUser: nova
+  # Q: do we need db an rabbit port? it seems today we use the default port
+  # This is the hostname of the default Rabbit today
+  apiMessageBusHostname: default-security-context
```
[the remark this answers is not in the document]
gibizer: I think this is the hostname the current rabbitmq operator generates for the service.

> [fedora@gibi-f35 ~]$ oc describe service/default-security-context
> Name:              default-security-context
> Namespace:         openstack
> Labels:            app.kubernetes.io/component=rabbitmq
>                    app.kubernetes.io/name=default-security-context
>                    app.kubernetes.io/part-of=rabbitmq
> Annotations:       <none>
> Selector:          app.kubernetes.io/name=default-security-context
> Type:              ClusterIP
> IP Family Policy:  SingleStack
> IP Families:       IPv4
> IP:                10.217.5.33
> IPs:               10.217.5.33
> Port:              amqp  5672/TCP
> TargetPort:        5672/TCP
> Endpoints:         10.217.0.50:5672
> Port:              management  15672/TCP
> TargetPort:        15672/TCP
> Endpoints:         10.217.0.50:15672
> Port:              prometheus  15692/TCP
> TargetPort:        15692/TCP
> Endpoints:         10.217.0.50:15692
> Session Affinity:  None
> Events:            <none>
> 

I think we agreed that Nova CR controller will get the Service selector labels, it selects the rabbit Service, creates the user in it, and passes down the rabbit hostname to the lower level CRs. So here in the NovaAPI this is the hostname.

config/samples/nova_v1beta1_novacell0.yaml
```diff
@@ -0,0 +1,74 @@
...
+    # This is the hostname of the default DB today
+    apiDatabaseHostname: openstack
+    apiMessageBusUser: nova
+    # Q: do we need db an rabbit port? it seems today we use the default port
+    # This is the hostname of the default Rabbit today
+    apiMessageBusHostname: default-security-context
+    cellDatabaseUser: nova
+    cellDatabaseHostname: openstack
+    cellMessageBusUser: nova
+    cellMessageBusHostname: default-security-context
```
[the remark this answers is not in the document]
gibizer: First, I think we agreed that cell0 is represented as a NovaCell CR. If so, then that cell should be listed in the NovaSpec.cellTemplates field as that field describes the list of cells that expected to exists. I think the whole concept of Spec and Status is that in the Spec you describe what is the expected state, and the controller works to make that happen, and report the actual state in the Status field to you. So if we don't have cell0 in NovaSpec.cellTemplates mapping then that means there is no expectation for cell0 NovaCell to exists. 

Second, there was a wish from us Compute DFG to make things deployable independently from the Nova CR. So if we deploy a nova control plane without Nova CR, then we do need to manually create cell0 based on this example.

config/samples/nova_v1beta1_novacell1-upcall.yaml
```diff
@@ -0,0 +1,113 @@
...
+        cpu: 10m
+        memory: 64Mi
+    debug:
+      stopDBSync: False
+      stopService: False
+      preserveJob: False
+  # lets assume in this example that metadata is at the top
+  metadataServiceTemplate: null
+  noVNCProxyServiceTemplate:
+    CellName: cell1
```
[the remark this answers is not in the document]
gibizer: As I noted above I don't know how to distinguish the super conductor in cell0 from a cell conductor in cellx with upcall support. Both has the same set of parameters other than cell name.

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -0,0 +1,410 @@
+apiVersion: nova.openstack.org/v1beta1
+kind: Nova
+metadata:
+  name: nova-multi-cell-sample
+spec:
+  apiDBSelector:
+    # This is how we point to the single global DB Service object today
+    cr: mariadb-openstack
```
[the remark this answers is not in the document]
gibizer: This is an example. I intentionally made an example that I can eventually validate with install yamls. Feel free to propose another example with split DB setup, that will be a good addition to our examples showing the flexibility of the CRD design.

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -0,0 +1,410 @@
...
+    containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:current-tripleo
+    replicate: 3
+    nodeSelector: {}
+    resources:
+      limits:
+        cpu: 500m
+        memory: 128Mi
+      requests:
+        cpu: 10m
+        memory: 64Mi
```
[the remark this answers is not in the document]
gibizer: This is an example. I intentionally made it vebose to show all the possible configuration. Please look at the minimal example https://github.com/openstack-k8s-operators/nova-operator/pull/61/commits/ce2645c0dd976ba348afd0620b93468dbe217606 to see what the the minimum to be configured.

config/samples/nova_v1beta1_nova.yaml
gibizer: I don't see in you proposal how the controller for the Nova CR would know:
- what passwords to be used in the various infra services
- which k8s Service object to be used to access various DB and message bus services. You especially want to use different DB for different cell so a single default value for all cells would not work

I agree we can remove the config overrides from the the minimal example and probably default the ContainerImages (but I afraid the default value will be up for debate).

Probably CellName can be removed from the NovaConductorTemplate (after we split out the Template struct from the Spec struct). But it needs to be in the NovaCellSpec and Template both.

SeanMooney (approved):
we have been discussing this on irc slack and github extensively.
as it stand having the basic crd is blocking some other work we would like to start.

while we have several outstanding issues to address in the PR we (dan gibi and i) feel that the best way forward is to merge this monolic PR and then opens a smaller more targeted pr using this as the base.

that will unblock me allowing me to work on the config generation and start on some basic envtest setup. and allow gibi to focus on a single example crd where we can have a simpler more targeted conversation of how to evolve the cards.

with that context I'm going to approve this and merge the PR

config/samples/nova_v1beta1_novaapi.yaml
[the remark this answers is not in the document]
bogdando: +1 that makes sense to pass down the hostname so that lower level CR controllers would not to repeat all the service label mathching logic of the top level one (and have it defined only there)

PR #62, opened 2022-09-13
[WIP] basic nova.conf config generation.

SeanMooney:
This change demonstrates how to use the lib-common Template
functionality to to generate a ConfigMap with a templated
nova.conf and static test data.


templates/nova/config/nova.conf
```diff
@@ -0,0 +1,190 @@
...
+never_download_image_if_on_rbd=true
+reserve_disk_resource_for_image_cache=true
+wait_for_vif_plugged_event_during_hard_reboot=["normal"]
+enable_qemu_monitor_announce_self=true
+# we might want to set this for upgrades.
+# disable_compute_service_check_for_ffu=true
+# skip_cpu_compare_on_dest=true
+# skip_hypervisor_version_check_on_lm=true
+
+[libvirt]
```
kk7ds, line 177: Lots of this file doesn't apply to every service. Were you thinking you'd just blow everything into every service, or be more selective? If you're already doing conf.d, you could have a 01-nova-base, 02-nova-servce, etc and put the service-specific things (like libvirt, scheduler config, etc) into the latter.

I dunno how I feel about it really, but it always seems cleaner to me to just have the things you can actually tweak for a given service in that file.

SeanMooney: more selective locally i have hte libvirt section only rendering if service_name=='nova-compute'

so i was going to have one template and only render the subsection that are relevnet to each service.

for the db for example it would only render if the function that create the template had the db info passed in the template parameters

the schdluer part will only render for the schudler

so each nova service crd they would have a config map with a nova.conf that is talored to that service.

right now I'm thinking of 3 files

01-nova.conf
02-nova-secret.conf
03-nova-overrides.conf

02 will have the password info for keystone db ectra.
but i have not refactored that yet.

kk7ds: > more selective locally i have hte libvirt section only rendering if service_name=='nova-compute'
> 
> so i was going to have one template and only render the subsection that are relevnet to each service.

Cool, sounds good.

> for the db for example it would only render if the function that create the template had the db info passed in the template parameters

Right for the two DBs, the presence or absence should be the key. But for things like [libvirt] I'd expect that to be the service name.

> 01-nova.conf 02-nova-secret.conf 03-nova-overrides.conf
> 
> 02 will have the password info for keystone db ectra. but i have not refactored that yet.

Sure, sounds good.

stuggi: if we think it gets to complex doing this all in a single template file and have the service_name conditions there, could also split out to multiple templates. The func will render all the files in the directory and add them to the configmap or secret depending where it should be stored.

dprince: If we do multiple templates/config files then we also need to be clear how the config parameters in the CRD impact each of these resulting config files.

SeanMooney: the config struct for the most part should be abstracted form the CRD perspective but the intent is that the CustomServiceConfig section of the card will gon directly into 03-nova-overrides.conf which will have the highest precedence and be able to override all the rest.

this precedence will be garuenteed by using config dir.

by the way i confirmed last night that crudini --merge does __not__ support MultiOptStrs

so the current way of merging config in use in the other operator is incorrect and should not be used.

in oslo_config

key=val1
key=val2 

results in an array {"key":["val1","val2"]}

conf format is not the same as ini format.

openstack uses conf format 

in ini format when parsed into a dict you will get

{"key": "val2"}

so key will have both a different type and a different value

i have not found a go config lib that actually support multiopt types so i have not found any way to merge multiple config in go correctly without writing a custom parser which i don't want to do.

that is why I'm looking at using oslos --config-dir instead as that is the only way to have the same snipt in multiple files without undefined behavior.

I'm splitting 01-nova.conf and 02-nova-secret.conf
because martin previously said we should not store passwords in a config map but we need usernames and password in the congis so the idea is to store 02-nova-secret.conf instead of a config map.

that feels a little overkill to me but we can do that

if i don't need too we would just have 2 files

01-nova.conf
02-nova-overrides.conf


each stateful set will have there own custom copy of the 2/3 config files that are ended with only the parameters that nova service requires.

to falicate that we will need CustomServiceConfig parmater per nova service type or some other way to address that in the top level CRD alternatively 

the CustomServiceConfig content itself can and should be treated as a go template so that single filed can use service_name and ifs to specify more detailed info.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
+[DEFAULT]
+{{/*controller services will have stable names provide by a stateful set*/}}
+{{ if eq .service_name "nova-compute"}}
+host={{ .fqdn }}
+{{/*the current default in ooo is wrong we should limit to 1*/}}
+# concurrent live migrations are more likely to fail and are slower
+# overall then serializing live migrations so set this to 1 explictly
+max_concurrent_live_migrations=1
```
gibizer, line 8: 1 is the default: https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.max_concurrent_live_migrations

btw, what is our strategy:
- a) add only lines that are changing something over the default
- b) generate a full config including defaults

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+{{ if .enable_ceilometer }}
+instance_usage_audit = true
+instance_usage_audit_period = hour
```
gibizer, line 11: The if ceilometer is enabled then enable emitting instance_usage_audit is business logic. I'm wondering if we want to mingle config generation with business logic.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+{{end}}
+{{end}}
+debug={{if .debug }}true{{else}}false{{end}}
+state_path = /var/lib/nova
+allow_resize_to_same_host = true
```
gibizer, line 16: why we enable this by default?

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+{{/*we can do log rotation differntly later but for now lets limit by size*/}}
+# enable log rotation in oslo config by default
+max_logfile_count=5
+max_logfile_size_mb=50
+log_rotation_type=size
+# ensure safe defaults for new hosts
```
gibizer, line 22: fine by me. This will be default (except the disk) as soon as we merge https://review.opendev.org/c/openstack/nova/+/830829

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+initial_cpu_allocation_ratio=4.0
+initial_ram_allocation_ratio=1.0
+initial_disk_allocation_ratio=0.9
+{{/*we really should change this upstream too*/}}
+# use predictable instnace names for simpler debuging.
+instance_name_template="instance-%(uuid)s"
```
gibizer, line 28: We need to be very careful not to have this for the adoption case. Quote from the config doc: "If you already have instances in your deployment when you change this, your deployment will break."

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+#rabbit_quorum_queue=true
+amqp_auto_delete=true
+{{/*we might just want to make this always false*/}}
+{{ if eq .service_name "nova-api"}}
+heartbeat_in_pthread=true
+{{else}}
+heartbeat_in_pthread=false
+{{end}}
+
+{{ if eq .service_name "nova-api"}}
```
gibizer, line 47: at this point I started feeling that we would be better off having a different template per service binary. That would make these templates a lot more readable.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+[oslo_policy]
+enforce_new_defaults=true
+{{end}}
+
+{{ if eq .service_name "nova-scheduler"}}
+[filter_scheduler]
+available_filters = nova.scheduler.filters.all_filters
+enabled_filters = ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,ServerGroupAntiAffinityFilter,ServerGroupAffinityFilter,SameHostFilter,DifferentHostFilter,PciPassthroughFilter,NUMATopologyFilter
+track_instance_changes = false
+shuffle_best_same_weighed_hosts = true
```
gibizer, line 57: I would disable it by default to keep the scheduler as deterministic as possible to ease debugging.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+{{ if eq .service_name "nova-api"}}
+[oslo_policy]
+enforce_new_defaults=true
+{{end}}
+
+{{ if eq .service_name "nova-scheduler"}}
+[filter_scheduler]
+available_filters = nova.scheduler.filters.all_filters
+enabled_filters = ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,ServerGroupAntiAffinityFilter,ServerGroupAffinityFilter,SameHostFilter,DifferentHostFilter,PciPassthroughFilter,NUMATopologyFilter
+track_instance_changes = false
```
gibizer, line 56: - 1) Do we want to make this dependent on the cells having upcall support? I.e. if each cell in the deployment has upcall support then this can be enabled to improve performance
- 2) I think we need to make sure that this is set both here and in the compute configs in the same way. So this should probably come from outside of the template as an input.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+shuffle_best_same_weighed_hosts = true
+
+[scheduler]
+max_attempts = 10
```
gibizer, line 60: what is the reason diverging from the default 3?

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+{{/*We should turn this off but for now 60 will work*/}}
+discover_hosts_in_cells_interval = 60
```
gibizer, line 62: I agree. For now we can keep it for testing but at some point implement a watch for a CRD when we see how the external compute will be represented.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+backend = oslo_cache.memcache_pool
+memcache_servers = {{range .memcache_servers -}}{{.address}}:{{.port}},{{- end}}
+{{else}}
+# on compute nodes or where memcache is not deployed we should use an in memory
+# dict cache
+backend = oslo_cache.dict
+{{end}}
+
+[os_vif_ovs]
+ovsdb_interface=vsctl
```
gibizer, line 126: Do we need this outside of a nova-compute?

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+
+[workarounds]
+{{/*we can deside if we want to support this later*/}}
+disable_group_policy_check_upcall=true
```
gibizer, line 130: the CR model already has enough input to decide on this from the cell conductor perspective. If the Conductor CR  has an API DB and message bus hostname provided then we can enable upcalls.

templates/nova.conf
```diff
@@ -0,0 +1,156 @@
...
+disable_fallback_pcpu_query=true
```
gibizer, line 131: yeah this make sense for new deployments

templates/nova-secret.conf
```diff
@@ -0,0 +1,80 @@
+[DEFAULT]
+transport_url={{.transport_url}}
+
+[keystone_authtoken]
+service_type = compute
+www_authenticate_uri = {{ .keystone_internal_url }}
+auth_url = {{ .keystone_internal_url }}
```
gibizer, line 7: this is deprecated in favor of www_authenticate_uri

SeanMooney: oh i tought it was likely the otherway
i can prefer www_authenticate_uri

templates/nova.conf
SeanMooney: so os-vif is initallised i think in the conductor to but i should check.

in generally we likely should not set this by default so i might drop it.

common/config.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func (r *ReconcilerBase) GenerateServiceConfigMaps(
+	ctx context.Context, h *helper.Helper,
```
SeanMooney, line 17: ah this is from go linit
ok was jus trunnign govet and go fmt

templates/nova.conf
SeanMooney: we dont enable it by default currently.
so in the template I'm setting it to false by default

in previous patches i had this hardcoded to true in the templateParameters
to confirm that multiple types work properly i.e. strings and booleans

right now i have not hooked this up to the CR in put and the current templateParameters
do not set it so it was produced the string <no value>

unset values in go templates are not just the empty string or omitted by default unfortunately 

https://levelup.gitconnected.com/when-using-go-templates-dont-neglect-the-missingkey-option-cc28a7e967b3

https://pkg.go.dev/text/template#Template.Option

"missingkey=default" or "missingkey=invalid"
	The default behavior: Do nothing and continue execution.
	If printed, the result of the index operation is the string
	"<no value>".
"missingkey=zero"
	The operation returns the zero value for the map type's element.
"missingkey=error"
	Execution stops immediately with an error.


i could also do this as

{{if .debug }}debug=true{{end}}

SeanMooney:
this is what the config map will look like https://paste.opendev.org/show/bH1s1G7c6OHbPJEYfm6l/ 
using the sample novaapi  config/samples/nova_v1beta1_novaapi.yaml to provide the input data

this is not complete but its a good start.

templates/nova.conf
SeanMooney: in general to only set what is needed that differs form the default

however ooo overrides our upstream default os i wanted to explict set it back to our upstream default as we as a dfg have asked for this to be reverted to one several tiems and it has not been done in ooo.

templates/nova.conf
SeanMooney: well we will be seeting the template parmater i was thinking that all condtional config logic would be handeled in the template and we would just set high level options like this in the paramters

templates/nova.conf
SeanMooney: ya i was wondering after i wrote this if i should change the disk default in that patch too. again these are just safe defaults.

templates/nova.conf
SeanMooney: however one exctption to that is default that change between releases

basically while we will very likely use zed as the base of this poc some of the options have diffent default on wallaby so there speicificaly im intending to use the zed value.

templates/nova.conf
SeanMooney: that is a very good point.

i would really like to make this change upstream too but that is why we have not change it there.

i would still like to make this the new default but yes for adoption this will need to be overriden. 

one of the big adgantaces to sueing the uuid in the name is when you live/cold migrate between host, hard reboot or rebuild we will always use the same libvirt domain name meaing all the libvirt/qemu logs for that instance at least on a given host remain in one file so you can see the full history without having to look at multiple files or check what it will be called on the remote host.

i would also like ot knwo how it can break things.
but thats is somethign ill look into later.

templates/nova.conf
SeanMooney: i kind of wanted to avoid that to not repeate any content between files but i may be able to decompose this into a more granulare approch.

im certenly open to that but im used to reading teamplates like this form kolla over the years so its seam pretty intuitive to me to have one templaste per config file with section generated dynamicaly based on multiple invocations.

templates/nova.conf
SeanMooney: your right it does need to be set int he compute config i remeber seeing that...

and yes it shuold be conditional since we are unfortunelly going to support that case.

i had hopped orginaly to not support upcalls at all.
but since dansmith thinks we should support it for anti affintiy i guess we will have to.

i was going to orginaly base this contionality on if the api db info was set but that wont work so ill add an upcalls parmaters to the template parmaters and use that to set this.

templates/nova.conf
SeanMooney: we had agreed to set this to true by default upstream.
i have a patch somewhere
this will only shuffle host with the same weight.

the host subset size is still 1

i would prefer not to disable this by defualt and i will see if i can review my old patch to change this upstream.

https://etherpad.opendev.org/p/nova-wallaby-ptg#L942

the reason that has not merged yet is i had my other schduler default changes in the same patch and the other were not aggree but this was.

templates/nova.conf
SeanMooney: i based the template on kollas and they use 10
no other reason i can drop this

SeanMooney (commented):
thanks for the feedback gibi ill rebase this in the morning and ill remove some of the overrides.

gibizer:
Can we close this in favor of #107 ?

SeanMooney:
yep i just had it open to copy paste form but we don't need this anymore

PR #65, opened 2022-09-21
Log build hash

gibizer:
Log the git hash of the operator at startup. This will help correlating logs of the operator to a given version of the operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,8 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+BUILD_HASH = "main.gitBuildHash=$(shell git describe --always --dirty 2>/dev/null || echo "<failed to generate>")"
+
 .PHONY: all
 all: build
 
@@ -109,11 +111,11 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	go build -ldflags "-X $(BUILD_HASH)" -o bin/manager main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+	go run -ldflags "-X $(BUILD_HASH)" ./main.go
 
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -37,8 +38,9 @@ import (
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme       = runtime.NewScheme()
+	setupLog     = ctrl.Log.WithName("setup")
+	gitBuildHash = "<not provided during build>"
 )
 
 func init() {
@@ -64,6 +66,8 @@ func main() {
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	setupLog.Info(fmt.Sprintf("Git build hash: %s", gitBuildHash))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 RUN if [ ! -f $CACHITO_ENV_FILE ]; then go mod download ; fi
 
 # Build manager
-RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; CGO_ENABLED=0  GO111MODULE=on go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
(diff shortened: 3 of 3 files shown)


main.go
```diff
@@ -37,8 +38,9 @@ import (
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme       = runtime.NewScheme()
+	setupLog     = ctrl.Log.WithName("setup")
+	gitBuildHash = "<unknown>"
```
SeanMooney, line 43: this shoudl be a constant not a varible

main.go
```diff
@@ -65,6 +67,8 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	setupLog.Info(fmt.Sprintf("Git build hash %s", gitBuildHash))
```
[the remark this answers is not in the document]
gibizer: I tested it with CRC and it does print the hash for me. So at least the docker based build works as expected. So I assume the `make build` part is breaking somehow.

main.go
gibizer: setupLog was generated code.

main.go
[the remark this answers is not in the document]
gibizer: For me it works for the `make build` path too. 

> gibi@riverbed:~/upstream/git/openstack-k8s-operators/nova-operator [log-build-hash]                                                                                             [000] [13:15:37]
> $ make build
> test -s /home/gibi/upstream/git/openstack-k8s-operators/nova-operator/bin/controller-gen || GOBIN=/home/gibi/upstream/git/openstack-k8s-operators/nova-operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2
> /home/gibi/upstream/git/openstack-k8s-operators/nova-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
> go fmt ./...
> go vet ./...
> go build -ldflags "-X main.gitBuildHash=cnosp-34-g6a3f641" -o bin/manager main.go
> gibi@riverbed:~/upstream/git/openstack-k8s-operators/nova-operator [log-build-hash]                                                                                             [000] [13:15:44]
> $ ./bin/manager 
> 1.6637589542539115e+09	INFO	setup	Git build hash cnosp-34-g6a3f641
> 1.663758954254002e+09	ERROR	controller-runtime.client.config	unable to get kubeconfig	{"error": "invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable", "errorCauses": [{"error": "no configuration has been provided, try setting KUBERNETES_MASTER environment variable"}]}
> sigs.k8s.io/controller-runtime/pkg/client/config.GetConfigOrDie
> 	/home/gibi/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.13.0/pkg/client/config/config.go:153
> main.main
> 	/home/gibi/upstream/git/openstack-k8s-operators/nova-operator/main.go:72
> runtime.main
> 	/usr/lib/go-1.18/src/runtime/proc.go:250
> 

So I assume it does not work for you as the build is done by vscode and not by make

main.go
gibizer: I tried. If it is a const then it cannot be set via build time flags.

main.go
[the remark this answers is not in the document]
gibizer: Hm `make run` uses `go run ./main.go` so it does not go through the `build` target to get a binary to run.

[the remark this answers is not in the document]
gibizer: OK I can fix the `make run` target to pass the same ldflag there too.

SeanMooney:
> I tried. If it is a const then it cannot be set via build time flags. 

ack

i was building form the command line with make build

when i do make run however its unset.

this not being done when i build in vscode becaust thats currently using go build directly although i coudl alster that. currntly that is fine.
we shoudl avoid using compile time defines like this in general but im fine with it for this one usece case

SeanMooney (approved):
> Hm `make run` uses `go run ./main.go` so it does not go through the `build` target to get a binary to run. 

ok that makes sense its buildign it internally without calling the build target that proably the correct default behavior

i verifed this work by doing this

[stack@crc nova-operator]$ make build
test -s /home/stack/repos/nova-operator/bin/controller-gen || GOBIN=/home/stack/repos/nova-operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2
/home/stack/repos/nova-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
go build -ldflags "-X main.gitBuildHash=cnosp-34-g6a3f641" -o bin/manager main.go
[stack@crc nova-operator]$ ./bin/manager 
1.663759695392221e+09   INFO    setup   Git build hash cnosp-34-g6a3f641
so im ok with htis as proposed

gibizer:
fixed up the `make run` path

Makefile
```diff
@@ -109,11 +111,11 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	go build -ldflags "-X $(BUILD_HASH)" -o bin/manager main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+	go run -ldflags "-X $(BUILD_HASH)" ./main.go
```
SeanMooney, line 118: ok, i guess that's fair.
as long as we don't use -X a lot this is fine i would hate
to have many flagos set by -X that change behaviour

but the build sha is fine

main.go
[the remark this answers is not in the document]
SeanMooney: ack

SeanMooney (dismissed):

main.go
kk7ds, line 43: Not a big deal, but if you default this to the same "<unknown>" here as you do in the '|| echo "<unknown>" in the build, you won't know whether it's unknown because the build couldn't figure it out, or if the build failed to set it.

SeanMooney: hum ya thats a good point i did not consider that

PR #66, opened 2022-09-21
NovaAPI controller logging

gibizer:
Set up clients and logging for the NovaAPI controller. Also I took the opportunity to document some developer guidelines. Let's try to build that doc up as we add the implementation.

Note that this is part of a PR chain. So lets land https://github.com/openstack-k8s-operators/nova-operator/pull/65 first.

### DIFF
--- a/doc/developer.md
+++ b/doc/developer.md
@@ -1,0 +1,12 @@
+
+# Implementation guidelines
+This is a collection of guidelines for developing the nova-operator. The goal
+here is to keep the codebase consistent and help spread good patterns
+between the developers.
+
+Please read the [developer guidelines](https://github.com/openstack-k8s-operators/docs/blob/main/developer.md)
+from openstack-k8s-operator first. Also use that doc for guidelines that are
+true for the whole openstack-k8s-operator development and use this nova
+specific doc for nova specific guidelines.
+
+## \<add the next nova specific guideline here\>
--- a/controllers/novaapi_controller.go
+++ b/controllers/novaapi_controller.go
@@ -20,17 +20,26 @@ import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
-	novav1beta1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
+	"github.com/go-logr/logr"
+	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
+
+	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
+
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // NovaAPIReconciler reconciles a NovaAPI object
 type NovaAPIReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Kclient kubernetes.Interface
+	Scheme  *runtime.Scheme
+	Log     logr.Logger
 }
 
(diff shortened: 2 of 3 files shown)


doc/developer.md
```diff
@@ -0,0 +1,19 @@
+
+# Implementation guidelines
```
stuggi, line 2: could you document this in https://github.com/openstack-k8s-operators/docs and link it here. I think it might be useful to document generic things like this in the broader doc repo and only nova specifics in here?

gibizer: Proposed in docs: https://github.com/openstack-k8s-operators/docs/pull/5

SeanMooney: lets kee this local file too.
there might be things that we want to call out specificaly for the nova operator
as it will be more complext then others

we also my have addtional tool/test that might not be required in other operator that we will want to call out.

i dont know about other but  i would like to continue to use reno for example so we can have release notes in repo.
i would also prefer to use rst for this doc but i guess markdown will do for now. i general i prefer rst to md but there
are pros and cons to both. and i dont wnat to fight that war right now.

as long as we dont have ascci doc or some other format i can live with markdown or rst.

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 RUN if [ ! -f $CACHITO_ENV_FILE ]; then go mod download ; fi
 
 # Build manager
-RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; CGO_ENABLED=0  GO111MODULE=on go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
+RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; export GIT_COMMIT=$(git describe --always --dirty 2>/dev/null || echo "<unknown>") && echo $GIT_COMMIT && CGO_ENABLED=0  GO111MODULE=on go build -ldflags "-X main.gitBuildHash=$GIT_COMMIT" ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
```
SeanMooney, line 27: this feels unrelated but ok

Makefile
```diff
@@ -109,7 +109,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	go build -ldflags "-X main.gitBuildHash=$(shell git describe --always --dirty 2>/dev/null || echo "<unknown>")" -o bin/manager main.go
```
SeanMooney, line 112: i agree with doing this in general just not sure why this is in the logging pr

SeanMooney: later ok its not this is also showin gme the other pr content because you did not mast this off master you based it off pr 65

so you staked them i guess in some case that makes sense but in this case you proably should have rebased this oen commit on master to keep the pr clean

SeanMooney:
ok ill go review these in order but  i guess in future we shoudl only submit prs based on master unless we are breakign up a big feature into smaller mergable chunks so aoivd confusion. unles there is a way for me to select only  whic commits to inlcude in the vscode pr view? ill go review these in order now

gibizer:
> ok ill go review these in order but i guess in future we shoudl only submit prs based on master unless we are breakign up a big feature into smaller mergable chunks so aoivd confusion. unles there is a way for me to select only whic commits to inlcude in the vscode pr view? ill go review these in order now

Sorry I will try better next time. The two commits pulled in here could have been put to master independently. 
But honestly I blame the limitations of github here. In reality implementing the logic of the NovaAPI controller will be one big series as the pieces there each depend on the previous step / commit so I cannot (and does not) want to put them independently to master. I guess we have to accept that github does not help here.

SeanMooney (approved):
yep its fine soe i have reviewd the previous pr and approved it and i think this one looks fine too now that i have filtered out the chagned form the previous commit

gibizer:
Made this independent from #65

doc/developer.md
SeanMooney: yep

SeanMooney (approved):

PR #67, opened 2022-09-21
Add NovaAPITemplate to be used in NovaSpec

gibizer:
We need to split what input is coming from outside (i.e. from the user) and what input is generated by a controller for our service CRs.

This patch create a NovaAPITemplate for the external input and use that in the NovaSpec instead of the NovaAPISpec.


config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -38,21 +38,17 @@ spec:
...
     defaultConfigOverwrite:
-      policy.yaml: >
-        # my custom policy
-    containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:current-tripleo
+      logging.conf: >
+        # my custom logging configuration
+    # TODO(gibi): How can we have a generic repo + tag definition up in Nova CR
+    # so that we can replace this with auto generated image urls from the Nova
+    # controller?
+    containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:zed-tripleo
```
SeanMooney, line 51: for reasons i dont understand they are publishing each release to a different organisation on quay

quay.io/tripleowallabycentos9 has the wallaby images
https://quay.io/repository/tripleomastercentos9 has the master images.

so the tag might stay current-tripleo to get the latest image but the org would chagne to tripleozedcentos9

can you change this quay.io/tripleowallabycentos9/openstack-nova-api:zed-tripleo to
quay.io/tripleomastercentos9/openstack-nova-api:current-tripleo

SeanMooney (changes requested):
im much happier with this as our interface

kk7ds:
Yeah agreed this is looking cool to me. I think the ContainerImage thing is probably something we should table for the moment and just keep it like you have here. I don't understand how we're going to know if we need to do major or minor upgrade type tasks purely based on updates to an otherwise opaque URL, but presumably that's a larger discussion.

config/samples/nova_v1beta1_nova-multi-cell.yaml
gibizer: Based on the discussion https://coreos.slack.com/archives/CQXJFGMK6/p1663752237261069 I think we will keep an non templated `containerImage` both in each Spec and in each Template too. Version tags cannot be used as offline installation needs to refer images by hash.

api/v1beta1/novaapi_types.go
```diff
@@ -19,12 +19,53 @@ package v1beta1
...
+	// ConfigOverwrite - interface to overwrite default config files like e.g. logging.conf
+	// But can also be used to add additional files. Those get added to the service config dir in /etc/<service> .
+	DefaultConfigOverwrite map[string]string `json:"defaultConfigOverwrite,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Resources - Compute Resources required by this service (Limits/Requests).
+	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+}
+
```
gibizer, line 68: We can make KeystoneAuthURL in the Spec required now.

gibizer: Done

config/samples/nova_v1beta1_nova-multi-cell.yaml
SeanMooney: ack i was more pointing out thtat the url you used in the sample was not vaild.

the way tags are used upstream for ooo and downstream fo our product differ so upstream the branch is contoled by the org in the url and the tag is just a a sha or current-tripleo for the head of the branch

so there is no zed-tripleo in tripleowallabycentos9

SeanMooney:
the only nit i have this this is the curren url in the sample is not a real image can you just put it back to the defuatl or use the master url instead.

config/samples/nova_v1beta1_nova-multi-cell.yaml
gibizer: Ah. OK. Fixed it.

SeanMooney (approved):

PR #68, opened 2022-09-22
Split spec template for services in the cell

gibizer:
Depends on https://github.com/openstack-k8s-operators/nova-operator/pull/67


api/v1beta1/novaapi_types.go
```diff
@@ -46,30 +87,15 @@ type NovaAPISpec struct {
...
-	// c) add a validating webhook here that runs only when NovaAPI CR is
-	//    created and does not run when Nova CR is created and make this field
-	//    required via that webhook.
+	// +kubebuilder:validation:Required
 	KeystoneAuthURL string `json:"keystoneAuthURL"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=nova
 	// APIDatabaseUser - username to use when accessing the API DB
 	APIDatabaseUser string `json:"apiDatabaseUser"`
```
SeanMooney, line 96: we shoudl review perhaps taking the db and message bus parmaters as a rendered connection string in the secret in the future but for now thsi makes sesne to require.

api/v1beta1/novacell_types.go
```diff
@@ -24,6 +24,23 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// NovaCellTemplate defines the input parameters specified by the user to
+// create a NovaCell via higher level CRDs.
+type NovaCellTemplate struct {
+	// +kubebuilder:validation:Required
```
SeanMooney, line 30: we could make this optional since the db and message bus will be on the cell.
depending on our default for the replicas of the conductor you woudl get 3 or 1 by defualt unless you passed null or set it to 0 in the template.

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -38,21 +38,14 @@ spec:
...
-    apiMessageBusUser: nova
+    #NOTE(gibi): secret with users is automatically passed by Nova CR
     customServiceConfig: >
       # service config customization
       debug=True
     defaultConfigOverwrite:
-      policy.yaml: >
-        # my custom policy
-    containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:current-tripleo
+      logging.conf: >
```
SeanMooney, line 46: i think i said thie before but we shoudl use | not >

https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics

Values can span multiple lines using | or >. Spanning multiple lines using a “Literal Block Scalar” | will include the newlines and any trailing spaces. Using a “Folded Block Scalar” > will fold newlines to spaces; it’s used to make what would otherwise be a very long line easier to read and edit. In either case the indentation will be ignored. Examples are:

`include_newlines: |
            exactly as you see
            will appear these three
            lines of poetry

fold_newlines: >
            this is really a
            single line of text
            despite appearances
`

this is why the orveride was being collapsed into a singel line when i was testing with customServiceConfig
can you fix this in a new revision

config/samples/nova_v1beta1_novacell1-upcall.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   cellName: cell1
   secret: osp-secret
-  # Here we have o have both api and cell passwords selectors
+  # Here we have to have both api and cell passwords selectors
```
SeanMooney, line 8: ack ill update my config gen patch to enable/disble the upcall based on if we have both dbs or not.

right now im hardcoding it to off.

SeanMooney (changes requested):
over all im continuing to like the direction is is moving in one minor thing in line but over all looking good

api/v1beta1/novaapi_types.go
gibizer: Yeah, this is pending the Secret discussion. If the Nova CRD controller will generate individual Secret objects to lower level CRs then we can put the rendered connection string in that Secret instead of putting the pieces there.

api/v1beta1/novacell_types.go
```diff
@@ -24,6 +24,23 @@ import (
...
+	// ConductorServiceTemplate - defines the cell conductor deployment for the cell
+	ConductorServiceTemplate NovaConductorTemplate `json:"conductorServiceTemplate"`
+
+	// +kubebuilder:validation:Optional
+	// MetadataServiceTemplate - defines the metadata serive dedicated for the cell.
+	MetadataServiceTemplate NovaMetadataSpec `json:"metadataServiceTemplate"`
+
+	// +kubebuilder:validation:Optional
+	// NoVNCProxyServiceTemplate - defines the novvncproxy serive dedicated for
```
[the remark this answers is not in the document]
gibizer: yepp.

api/v1beta1/novacell_types.go
```diff
@@ -24,6 +24,23 @@ import (
...
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// NovaCellTemplate defines the input parameters specified by the user to
+// create a NovaCell via higher level CRDs.
+type NovaCellTemplate struct {
+	// +kubebuilder:validation:Required
+	// ConductorServiceTemplate - defines the cell conductor deployment for the cell
+	ConductorServiceTemplate NovaConductorTemplate `json:"conductorServiceTemplate"`
+
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
gibizer: Yes, It is optional as it either runs on the top or per cell. I think the Nova CR can default to either create it on the top or per cell. I would default it to run on the top level just simply because that is the simpler setup.

api/v1beta1/novacell_types.go
gibizer: Done.

config/samples/nova_v1beta1_nova.yaml
```diff
@@ -12,59 +12,30 @@ spec:
...
-        containerImage: quay.io/tripleowallabycentos9/openstack-nova-conductor:current-tripleo
-      metadataServiceTemplate: null
-      noVNCProxyServiceTemplate:
-        CellName: cell1
-        secret: osp-secret
-        passwordSelectors:
-          cellDatabase: NovaCell1DatabasePassword
-        containerImage: quay.io/tripleowallabycentos9/openstack-nova-novncproxy:current-tripleo
+      # In this example we have metadata at the top level
+      metadataServiceTemplate: null
```
gibizer, line 41: The cell template is broken now. There is no way to specify the CellDB and CellMq.

gibizer: Fixed it with adding new fields to the NovaCellTemplate. Note that CellDatabaseInstance and CellMessageBusInstance already follow the convention to pass depnedencies by CR name instead of label selector. See also https://github.com/openstack-k8s-operators/nova-operator/pull/71

api/v1beta1/novacell_types.go
```diff
@@ -24,6 +24,42 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// NovaCellTemplate defines the input parameters specified by the user to
+// create a NovaCell via higher level CRDs.
+type NovaCellTemplate struct {
+	// +kubebuilder:validation:Required
+	// CellDatabaseInstance is the name of the MariaDB CR to select the DB
+	// Service instance used as the DB of this cell.
+	CellDatabaseInstance string `json:"cellDatabaseInstance"`
```
kk7ds, line 33: I wonder if it makes sense to name these exactly like the top-level and other operators do? Like I wonder if the same argument will be applied here for consistency? So:

```
NovaSpec:
  DatabaseInstance foodb
  MessageBusInstance foomq
  CellTemplates:
    - DatabaseInstance bardb
      MessageBusInstance barmq
    - ...
```

Unless you think it would be confusing to have multiple of the same parameters in the hierarchy, I think maybe we should consider making them the same.

kk7ds (commented):
Just one comment about consistency but not critical. Otherwise I like how this is going :)

api/v1beta1/novacell_types.go
gibizer: Do you mean only differentiate between the API and the Cell DB on the Cell level. 

A cell with upcall support:
```
NovaSpec:
  DatabaseInstance foodb
  MessageBusInstance foomq
  CellTemplates:
      DatabaseInstance bardb
      MessageBusInstance barmq
      APIDatabaseInstance foodb
      APIMessageBusInstance foomq
```

I think in this case it is confusing that the DatabaseInstance on the top means the same as APIDatabaseInstance in the cell.

SeanMooney: i guess you also meant drop the nova prefix form  NovaCellTemplate to just be CellTemplate

which I think makes sense.

if I'm not reading between the line too much 

NovaSpec:
  DatabaseInstance foodb
  MessageBusInstance foomq
  CellTemplates:
    - DatabaseInstance bardb
      MessageBusInstance barmq
    - ...

dan might have been suggesting a list of dictoaries of 
NovaSpec:
  DatabaseInstance foodb
  MessageBusInstance foomq
  CellTemplates:
    - DatabaseInstance apidb
      MessageBusInstance apimq
    - DatabaseInstance celldb
      MessageBusInstance cellmq

if That is what you meant by the ...
i ferfer what we have now as its clear which is which

kk7ds: All I meant was use "DatabaseInstance" anywhere we need one and not call it a "CellDatabaseInstance". If it's inside a CR where it's the only one, then there's no reason to prefix it with "Cell" I think.

kk7ds: > Do you mean only differentiate between the API and the Cell DB on the Cell level.
> 
> A cell with upcall support:
> 
> ```
> NovaSpec:
>   DatabaseInstance foodb
>   MessageBusInstance foomq
>   CellTemplates:
>       DatabaseInstance bardb
>       MessageBusInstance barmq
>       APIDatabaseInstance foodb
>       APIMessageBusInstance foomq
> ```
> 
> I think in this case it is confusing that the DatabaseInstance on the top means the same as APIDatabaseInstance in the cell.

Sorry I just saw this. You're right, it may be somewhat confusing in the spec, although I'm not super concerned about it, because it's kindof a pointer to something outside the cell. It's not "My cell's database instance" it's "that other thing over there".

Anyway, it seems like from the template point of view (i.e. our API) that keeping the names the same would be the better approach for consistency. But, I understand the concern, and it's just a matter of naming preference. I thought it might be a good thing to make us "look" more like the typical operator, but it's not critical :)

SeanMooney (approved):
lets pick up this converstation when dan and i are back form pto next week but i generally agree with the idea of useing

DatabaseInstance anywhere its appropriate

this particular case is less clean and im unsure which is better.

user CellDatabaseInstance for clarity or DatabaseInstance for consitency

since we can simple change this next week with liitle impact im going to appvoed this now since i didnt get to it friday evening and we can chat about this more on monday

SeanMooney (approved):
simple rebase for merge conflict this still looks fine for now so re approving

PR #70, opened 2022-09-22
[NovaAPI] init conditions

gibizer:


controllers/novaapi_controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+	AfterEach(func() {
+		novaAPI := GetNovaAPI(novaAPILookupKey)
+		Expect(k8sClient.Delete(ctx, novaAPI)).Should(Succeed())
+		// We have to wait for the controller to fully delete the instance
+		Eventually(func() bool {
+			err := k8sClient.Get(ctx, novaAPILookupKey, novaAPI)
+			return k8s_errors.IsNotFound(err)
+		}, timeout, interval).Should(BeTrue())
+
+	})
```
SeanMooney, line 109: is see so BeforeEach is the same as setUp and AfterEach is teardown
and they just lambda/anonomus function as a parmater to execute automatical within the test framework.

i will have to look up the docs for this particalar BDD framew work but i assume this will run before and after each When

this seem similar but slightly different then the "Given When Then" i am used to for BDD with frameworks like catch2 https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#bdd-style-test-cases

controllers/novaapi_controller_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+const (
+	Namespace  = "test-namespace"
+	SecretName = "test-secret"
+
+	timeout  = time.Second * 2
+	interval = time.Millisecond * 200
+)
+
+func CreateNamespace(name string) {
```
SeanMooney, line 41: we may want to create an integrated_helpers or test utils module in this operator to centralise simple things like this eventually

but the way i have to triple check this but i also dont belive we are required to keep the tests in the same module/package/folder

i have seen either the opeartorsdk or olm have a top level test directory so we may want to do that
i would prefer to split things up like we do in nova if we can with
funtional/envetes, and unit folders 

right now make test will runa all of them but we can eventually seperate out unit and functional tests into differnt make targets if exectuation time eventually becomes an issue.

anyway this is fine ot have inline for now but i expect all envete test will need this.

controllers/suite_test.go
```diff
@@ -65,6 +74,10 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
+	// NOTE(gibi): Need to add all API schemas our operator can own.
```
SeanMooney, line 77: ill step through this with a debugger but if i understand correctly
this file is setting up the k8s cluster to contain the crds that are part of the operator which is why we have to register the scheme her so that the novaapi_contoller_test.go can run?

so BeforeSuite is basically a global setup that happens before any of the test run to ensure all the resouce we will use are registered correct.

it that is not the case can you run through this with me tomorrow

SeanMooney:
> im not sure this is a good idea.

> the helper contains a refernce to the instance but it will out live the current reconsile loop so that instance cannot be deallocated safely until the helper is replaced the next time the reconsile is called.

> i need to check golangs memory model to say if this is a potenial dangeling pointer or if it will pereform lifetime extention on the instance object but while this is technially not a memory leak it will result in some undesireabl behavior and may be unsafe. 

so its safe because lifetiem exstion does happen as the instance is allcoated on the heap since the go complier cant prove it stays lcoal to the function as part of its escape analaise because it does not in this case since your adding the helper object to the reconsiler 

https://web.archive.org/web/20160923043317/https://www.scvalex.net/posts/29/

for more details ^ 

so if we do this the instance obejct will existh in memory and never be deleted until the next reconsile iterations so let not do this if we can avoid it.

controllers/novaapi_controller.go
```diff
@@ -75,7 +77,7 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
-	helper, err := helper.NewHelper(
+	r.Helper, err = helper.NewHelper(
```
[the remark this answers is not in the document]
gibizer: Thanks for the pointers. I agree that we should not hold this in the Reconciler. I would like to removed then need to pass helper to every function we call from Reconcile() though.

controllers/novaapi_controller.go
```diff
@@ -86,11 +88,40 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		l.Error(err, "Failed to create lib-common Helper", "request", req)
 		return ctrl.Result{}, err
 	}
-	util.LogForObject(helper, "Reconciling", instance)
+	util.LogForObject(r.Helper, "Reconciling", instance)
+
+	// initialize status fields
+	if err = r.initConditions(ctx, instance); err != nil {
```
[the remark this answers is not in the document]
gibizer: Passed it now.

controllers/novaapi_controller_test.go
```diff
@@ -0,0 +1,122 @@
+/*
```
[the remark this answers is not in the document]
gibizer: Yeah I learned these during the summer when build some test to the placement-operator. There are even more patterns there that will appear here too when needed.

controllers/novaapi_controller_test.go
gibizer: Yes. @stuggi already suggested to move some of the common things to a lib-common/modules/test module https://github.com/openstack-k8s-operators/placement-operator/pull/21#issuecomment-1238272659 . CreateNamespace is a good candidate for that too. 

I also agree with you about structuring our tests more. I will push a patch and try to move this to a test package and add some extra structure to it.

controllers/novaapi_controller_test.go
gibizer: Yes. You are right about BeforeEach and AfterEach. Every When() is like a test case and the Descibe() above is like a test class. This style is still new to me so feel free to suggest useful patterns that would make our tests easier to read and adapt to further changes.

controllers/suite_test.go
gibizer: Yes BeforeSuite is the global as far as I see. Also yes, this step is needed so that the k8s api server knows that the give CRD exists and therefore can be instantiated. What I don't know is how can we have to set of tests like one unit that does not need an extensive BeforeSuite and one functional where the current BeforeSuit is needed. I will try to figure this out a bit later.

gibizer (commented):
you are right I removed the helper from the Reconciler that was a bad idea.

controllers/novaapi_controller_test.go
gibizer: This moves the functional test to a separate package https://github.com/openstack-k8s-operators/nova-operator/pull/72

controllers/novaapi_controller_test.go
SeanMooney: yep i read the docs on the test framework last night
really helped me fall asleep :P
but i understand it much better now.

there is an alterinivbe defer apptoch instead of AfterEach which we can use too.

I'm going to get used to this before fully forming an opinion but i think ill find myself gravitating to using Defer for cleanup inside beforeEach but if we group the BeforeEach and After each its also ok.

controllers/suite_test.go
SeanMooney: we can have multiple testSuites so i think  this file only applies to the package its defined in and to tests in that package

i have not tried it so ya we can figure that out as we go

SeanMooney (approved):

PR #71, opened 2022-09-22
Pass dependencies via names instead of labels

gibizer:
The dependencies of the Nova CRD is now passed by the name of the CR representing the dependency.

The agreemen and reasoning is documented in https://github.com/openstack-k8s-operators/docs/pull/6


SeanMooney:
i do like this approch more 
i thik it is reasonable to not have name collisions for these CRs im not even sure k8s allows that so the name is a unique identifyer

gibizer:
> i do like this approch more i thik it is reasonable to not have name collisions for these CRs im not even sure k8s allows that so the name is a unique identifyer

As far as I understand k8s enforces that names should be unique per Namespace and per Kind.

SeanMooney (approved):

PR #72, opened 2022-09-23
Move functional test to separate package

gibizer:
Depends on https://github.com/openstack-k8s-operators/nova-operator/pull/70


SeanMooney (approved):
this looks good to me too honestly

im going to check it out locally and play with it a bit but i like this refactoring

SeanMooney (approved):
thanks for the rebase

PR #75, opened 2022-09-23
precommit

SeanMooney:
- add basic precommit config


SeanMooney:
pre-commit.ci run

dprince:
Would this be needed if we just went ahead and enabled prow to merge commits for nova-operator like we are doing with other openstack-k8s-operators?

SeanMooney:
yes this is unrelated to prow
im adding pre-commit to automatically run things locally before you commit to avoid pushing things that will fail in ci

i also enabled the pre-commit.ic because its free and supper fast.
basically pre-commit is for style enformcne and not forgetting to do thing like run make manifest

prow can also be enable in the future altohg i want to looking what its doing and how its configured so that we can mange the config form this repo ideally.

can you provide an link to the config for prow jobs?

SeanMooney:
just in case you have not used pre-commit before https://pre-commit.com/ 
its "A framework for managing and maintaining multi-language pre-commit hooks."

its got pretty popular a few years ago and we started using it in OpenStack upstream to make sure things like the pep8 style check sere run locally before you pushed to ci to save ci resources by catching those types of things early.

if you just don't install the tool it will have no impact on your day to day dev experience so its just optional.

dprince:
@SeanMooney in that case these things can totally co-exist. If it plays in nova-operator maybe we'll use it everywhere then? LGTM

An example (intial) prow config for cinder:
https://github.com/openshift/release/commit/41b81d9871eed4f8de08e3a3779caed9f90bbeb0

SeanMooney:
thanks for the link :)
ya so this is more often used with python project or node there are glan spcpeic hooks and you can trivially write your own as i did to just rung the make file targets locally.

so if these prove useful to use for nova i was hoping we could port them to other
for context i am skipping some hooks in the pre-commit.ci because  that ci has issue with pulling modules from the golang.org proxy. locally they work fine but id does not seam to like how the proxy tries to swap from udp to TCP if the DNS lookup fails

what motivated me to do this is we were reviewing the inita CRDs and i notice the yaml files and go files where not in sync. so we wanted to catch that going forward locally before we pushed and while I'm at i i wanted to block things like commit private keys and other things that are generic to good git usage like no large files.

bogdando (approved):

gibizer (approved):
Looks good. Runs OK locally with `pre-commit run -a`. 

As a follow up could you add couple of lines of documentation in the README.md about how to set this up locally. I.e. that I need to run `pre-commit install` to have the hooks.

Issue #78, opened 2022-10-03
[envtest]Do not use simulated success when run against existing cluster


gibizer:
When envtest run locally the test case simulate Job and Deployment success by directly updating Status fields of CRs. When envtest runs against an external cluster the test should not simulate the success but actually need to wait for the success to happen automatically by the cluster. So make the simulation steps dependent on USE_EXISTING_CLUSTER ENV variable.

gibizer:
We agreed to use envtest exclusively in simulated mode and use kuttl test against a real deployment. So here I only need to cleanup some envtest cruft about real environment from the helpers and from the makefile.

PR #83, opened 2022-10-04
[envtest]Test refactors

gibizer:
Various improvements on the envtest to prepare for the NovaAPI tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,3 +269,7 @@ golangci: get-ci-tools
 # Run go lint against code
 golint: get-ci-tools
 	export GOWORK=off && PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh
+
+.PHONY: test_crc
+test_crc: manifests generate fmt vet envtest ## Run tests against a deployed CRC cluster
+	USE_EXISTING_CLUSTER=true go test ./... -coverpkg=./controllers -coverprofile cover.out
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -64,33 +64,33 @@ func CreateNovaAPI(namespace string, spec novav1.NovaAPISpec) types.NamespacedNa
 	return types.NamespacedName{Name: novaAPIName, Namespace: namespace}
 }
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -269,3 +269,7 @@ golangci: get-ci-tools
 # Run go lint against code
 golint: get-ci-tools
 	export GOWORK=off && PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh
+
+.PHONY: test_crc
+test_crc: manifests generate fmt vet envtest ## Run tests against a deployed CRC cluster
```
SeanMooney, line 274: this can be fixed in a followup but currently this assumes you have already run make install to create the crds so we can add that to the make env deps in the future

test/functional/base_test.go
```diff
@@ -44,33 +54,43 @@ func CreateNovaAPI(spec novav1.NovaAPISpec) types.NamespacedName {
...
 
-	return types.NamespacedName{Name: novaAPIName, Namespace: Namespace}
+	return types.NamespacedName{Name: novaAPIName, Namespace: namespace}
+}
+
+func DeleteNovaAPI(name types.NamespacedName) {
+	novaAPI := GetNovaAPI(name)
+	Expect(k8sClient.Delete(ctx, novaAPI)).Should(Succeed())
+	// We have to wait for the controller to fully delete the instance
+	Eventually(func(g Gomega) {
```
SeanMooney, line 71: i assume the g Gomega part is what was requried to proplery handel the panics.

SeanMooney (approved):

Makefile
gibizer: I tried with `install` in the list the test ends up hitting the same service account issue when run against an existing cluster as without `install`. So I could add but it would not help. We have #78 as a tracker to fix this, and other things in the test code to allow running `test_crc`

test/functional/base_test.go
gibizer: yes: https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

PR #92, opened 2022-10-11
Envtest improvements

### LINKED ISSUE #90: [envtest]Turn on parallel and randomized execution
To tease out unstable tests run them in parallel and in random order. This might need a change to the `make test` target to call ginkgo directly instead of via `go test`

gibizer:
There are two related envtest improvements:
1. Speed up the execution by reducing the RequeAfter timeouts to 10ms when running in envtest
2. Tease out unstable tests by running the test in random order and in parallel.

Closes #90
Depends-on #87


test/functional/base_test.go
```diff
@@ -40,12 +40,12 @@ const (
...
 
-	timeout  = time.Second * 20
-	interval = time.Millisecond * 50
+	interval = time.Millisecond * 100
+	timeout  = interval * 15
 	// consistencyTimeoutt is the amount of time we use to repeatedly check
 	// that a condition is still valid. This is intendet to be used in
 	// asserts using `Consistently`.
-	consistencyTimeout = time.Second
+	consistencyTimeout = interval * 15
```
SeanMooney, line 48: nit: you have an extra t in the comment above "consistencyTimeoutt"

controllers/utils.go
```diff
@@ -1,49 +1,103 @@
...
+			},
+			"NovaMetadata": &NovaMetadataReconciler{
+				ReconcilerBase: NewReconcilerBase("NovaMetadata", mgr, kclient),
+			},
+			"NovaNoVNCProxy": &NovaNoVNCProxyReconciler{
+				ReconcilerBase: NewReconcilerBase("NovaNoVNCProxy", mgr, kclient),
+			},
+		}}
+}
+
```
SeanMooney, line 85: ok, this improves the type of safety in general.
I'm not sure this file should still be called utils.go at this point but I'm ok with moving them to the controller module. it makes sense that this would be there.

controllers/utils.go
```diff
@@ -1,49 +1,103 @@
...
 		if err = controller.SetupWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", name)
-			os.Exit(1)
+			return err
 		}
 	}
+	return nil
+}
+
+// OverriedRequeueTimeout overrieds the default RequeueTimeout of our reconcilers
```
SeanMooney, line 98: ack so this is just a helper for testing
works for me

SeanMooney (approved):

controllers/utils.go
gibizer: yepp it is only for exposing the RequeueTimeout for testing.

test/functional/base_test.go
gibizer: fixed in the latest push

controllers/utils.go
SeanMooney: yep im ok with that

gibizer: Done in the latest push

SeanMooney (approved):

Issue #91, opened 2022-10-07
Use db.CreateOrPatchDBByName to select the proper MariaDB CR


gibizer:
As https://github.com/openstack-k8s-operators/lib-common/pull/63 merged we can now support different DB Service instance for different DB schemas by passing the MariaDB CR name to lib-common to indicate which DB Service we want to use. So Nova controller needs to be adapted to do so when create nova_api and cell0 DB.

Issue #89, opened 2022-10-06
NovaAPISpec should have Cell0DatabaseHostname and Cell0DatabaseUser


gibizer:
https://github.com/openstack-k8s-operators/nova-operator/blob/6e5ff216df799f08faaa8d3c6696d571a7e2ae73/api/v1beta1/novaapi_types.go#L100

The nova-api service depends on cell0 DB to register itself. So when the NovaAPI is created it should get the connection information.

PR #93, opened 2022-10-11
Support separate MariaDB instance for nova_api and nova cell DBs

gibizer:
[CRD][Nova] Support different DB Services

In a real multi cell deployment it is recommended to use separate DB
services for different cells. This patch adds support to separate the
DB service for the nova_api DB from each nova cell DB (including cell0).

To support this the NovAPISpec now takes Cell0DatabaseHostname as an
input top of the existing APIDatabaseHostname field. When nova is
deployed via the Nova CR the APIDatabaseInstance name can be provided
in NovaSpec while each NovaCellTemplate in NovaSpec.cellsTemplates has
its own CellDatabaseInstance field to specific the name of the DB
service used by the given cell. The Nova controller automatically fills
the Cell0DatabaseHostname of the NovaAPISpec so that the nova-api
service has access not just the nova_api DB but also the nova_cell0 DB
as well.

Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/91
Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/89

Depends-on #87

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -202,16 +202,21 @@ func (r *NovaReconciler) reconcileNormal(
 		"nova-api",
 		instance.Namespace,
 	)
-	result, err := r.reconcileDB(ctx, h, instance, apiDB, novav1.NovaAPIDBReadyCondition)
+	result, err := r.reconcileDB(
+		ctx,
+		h,
+		instance,
+		apiDB,
+		instance.Spec.APIDatabaseInstance,
+		novav1.NovaAPIDBReadyCondition,
+	)
 	if (err != nil || result != ctrl.Result{}) {
 		return result, err
 	}
 
 	cell0DB := database.NewDatabaseWithNamespace(
 		nova.NovaCell0DatabaseName,
-		// TODO(gibi): This should be cell0.CellDatabaseUser or should be
-		// embedded into the Secret we passing down to the cell
-		instance.Spec.APIDatabaseUser,
+		cell0Template.CellDatabaseUser,
 		instance.Spec.Secret,
 		map[string]string{
 			"dbName": cell0Template.CellDatabaseInstance,
@@ -219,7 +224,14 @@ func (r *NovaReconciler) reconcileNormal(
 		"nova-cell0",
 		instance.Namespace,
 	)
-	result, err = r.reconcileDB(ctx, h, instance, cell0DB, novav1.NovaCell0DBReadyCondition)
+	result, err = r.reconcileDB(
+		ctx,
+		h,
+		instance,
+		cell0DB,
+		cell0Template.CellDatabaseInstance,
+		novav1.NovaCell0DBReadyCondition,
+	)
 	if (err != nil || result != ctrl.Result{}) {
 		return result, err
 	}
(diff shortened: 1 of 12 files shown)


controllers/nova_controller.go
```diff
@@ -211,24 +211,36 @@ func (r *NovaReconciler) reconcileNormal(
...
 	)
-	result, err := r.reconcileDB(ctx, h, instance, apiDB, novav1.NovaAPIDBReadyCondition)
+	result, err := r.reconcileDB(
+		ctx,
+		h,
+		instance,
+		apiDB,
+		instance.Spec.APIDatabaseInstance,
+		novav1.NovaAPIDBReadyCondition,
+	)
```
gibizer, line 221: I expect the comment about formatting here :) This goes away in https://github.com/openstack-k8s-operators/nova-operator/pull/104

SeanMooney:
ok that didnt fix it ill leave it to you and re review tomorrow

controllers/nova_controller.go
```diff
@@ -202,24 +202,36 @@ func (r *NovaReconciler) reconcileNormal(
...
```
SeanMooney, line 212: result, err := r.reconcileDB(
		ctx, h,	instance, apiDB, instance.Spec.APIDatabaseInstance, novav1.NovaAPIDBReadyCondition,
	)
:)

pkg/novaapi/deployment.go
```diff
@@ -44,12 +44,9 @@ func Deployment(
...
 		KeystoneServiceUserPasswordSelector: instance.Spec.PasswordSelectors.Service,
-		// TODO(gibi): this should come as a separate input from the Spec
-		// instead for reusing the API DB stuff for cell
-		Cell0DatabaseHostname: instance.Spec.APIDatabaseHostname,
-		// TODO(gibi): ditto. This is a hack now
-		Cell0DatabaseUser: nova.NovaCell0DatabaseName,
-		Cell0DatabaseName: nova.NovaCell0DatabaseName,
+		Cell0DatabaseHostname:               instance.Spec.Cell0DatabaseHostname,
+		Cell0DatabaseUser:                   instance.Spec.Cell0DatabaseUser,
+		Cell0DatabaseName:                   nova.NovaCell0DatabaseName,
```
SeanMooney, line 49: because i need this for the api to start anyway I'm going to rebase my config change on this once this is merged as that will make things simpler

test/functional/base_test.go
```diff
@@ -339,16 +339,21 @@ func NovaConditionGetter(name types.NamespacedName) condition.Conditions {
...
 // CreateDBService creates a k8s Service object that matches with the
 // expectations of lib-common database module as a Service for the MariaDB
-func CreateDBService(namespace string, spec corev1.ServiceSpec) types.NamespacedName {
-	// TODO(gibi): Do we depend on the name?
-	serviceName := "not-openstack"
+func CreateDBService(namespace string, mariadbCRName string, spec corev1.ServiceSpec) types.NamespacedName {
+	// The Name is used as the hostname to access the service. So
+	// we generate something unique for the MariaDB CR it represents
+	// so we can assert that the correct Service is selected.
+	serviceName := "hostname-for-" + mariadbCRName
```
SeanMooney, line 346: oh this is the test i was like cant we just use nova_api ectra but cool we want this to be independent for each test

SeanMooney (approved):

PR #96, opened 2022-10-11
Issue 88 remove normal

### LINKED ISSUE #88: rename reconcileNormal to doRconsile or a similar more descriptive name.
There are 2 issues with the current naming

reconcileNormal is not partially descriptive in what the function is doing
or differentiating it from the public Reconcile function.

it is a private function used to actually implemented the reconsile called within Reconcile
after reconsile is able to satisfy some basic preconditions like retrieving the instance of the
resource that is to be managed and preparing some commen clients/helpers that will be required to
interact with the k8s API

the other issue is that reconcileNormal is not a particularly PC name scheme
there as some unavoidable uses of normal such as normalisation of import or in vector calculus
where using a different term would be incorrect but we do not need to prop use it to describe the happy path ie. the work to be done if setup succeeds.

gibizer:
Rename reconcileNormal to reconcile

Normal does not give any additional information for the reader so it is
removed from the name.

Closes #88


gibizer:
This will be replaced with a patch that flattens out reconcileNormal

PR #100, opened 2022-10-11
Refactor how Specs are created

gibizer:
The NovaConductor case is easy as the only source of input is the
NovaCellSpec. So reduced the input parameter list to a single
NovaCellSpec and move the factory method next to the NovaCellSpec
definition.

For the NovaAPI and NovaCell cases are more complicated as they have
multiple sources of inputs. So there I opted to have the Spec struct
creation inline in the Reconcile code as otherwise I had to pass around
a bunch of parameters to the factory method.


SeanMooney:
i need to review this again when I'm not as tired but this is looking good.
i like the reduction in codesize and consolidation

api/v1beta1/novaconductor_types.go
```diff
@@ -203,3 +203,20 @@ func init() {
 func (s NovaConductorStatus) GetConditions() condition.Conditions {
 	return s.Conditions
 }
+
+// NewNovaConductorSpec constructs a NovaConductorSpec
+func NewNovaConductorSpec(
+	novaCell NovaCellSpec,
```
SeanMooney, line 209: had to re read this but this is taking the cell spec to constuct a conductor spec for that cell.
so ya that shoudl work

SeanMooney (approved):

PR #103, opened 2022-10-11
extract reconciler base

SeanMooney:
- Factor out common reconsiler base
- Refactor reconciler setup
- [WIP] extract contoller creation to utils files

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -1,0 +1,51 @@
+package controllers
+
+import (
+	"os"
+
+	"github.com/go-logr/logr"
+	nova_common "github.com/openstack-k8s-operators/nova-operator/pkg/common"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// SetupReconcilers registers all reconcilers with a provided manager.
+func SetupReconcilers(mgr ctrl.Manager, setupLog logr.Logger, cfg *rest.Config) {
+	kclient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		setupLog.Error(err, "unable to create k8s client")
+		os.Exit(1)
+	}
+	reconcilers := map[string]nova_common.Managable{
+		"Nova": &NovaReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("Nova", mgr, kclient),
+		},
+		"NovaCell": &NovaCellReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaCell", mgr, kclient),
+		},
+		"NovaAPI": &NovaAPIReconciler{
+			ReconcilerBase:        nova_common.NewReconcilerBase("NovaAPI", mgr, kclient),
+			RequeueTimeoutSeconds: 5,
+		},
+		"NovaScheduler": &NovaSchedulerReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaScheduler", mgr, kclient),
+		},
+		"NovaConductor": &NovaConductorReconciler{
+			ReconcilerBase:        nova_common.NewReconcilerBase("NovaConductor", mgr, kclient),
+			RequeueTimeoutSeconds: 5,
+		},
+		"NovaMetadata": &NovaMetadataReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaMetadata", mgr, kclient),
+		},
+		"NovaNoVNCProxy": &NovaNoVNCProxyReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaNoVNCProxy", mgr, kclient),
+		},
+	}
+	for name, controller := range reconcilers {
+		if err = controller.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", name)
+			os.Exit(1)
+		}
+	}
+}
--- a/main.go
+++ b/main.go
@@ -20,7 +20,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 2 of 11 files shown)


controllers/utils.go
```diff
@@ -0,0 +1,53 @@
...
+		setupLog.Error(err, "unable to get config")
+		os.Exit(1)
+	}
+	kclient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		setupLog.Error(err, "unable to create k8s client")
+		os.Exit(1)
+	}
+	reconcilers := map[string]nova_common.Managable{
+		"nova": &NovaReconciler{
```
gibizer, line 25: Why the different casing for nova? All the rest is CameCase and the original reconciler also used "Nova"

test/functional/suite_test.go
```diff
@@ -92,12 +93,18 @@ var _ = BeforeSuite(func() {
...
 	err = (&controllers.NovaAPIReconciler{
-		Client:  k8sManager.GetClient(),
-		Scheme:  k8sManager.GetScheme(),
-		Kclient: kclient,
-		Log:     ctrl.Log.WithName("controllers").WithName("NovaAPI"),
+		ReconcilerBase: nova_common.ReconcilerBase{
+			Client:  k8sManager.GetClient(),
+			Scheme:  k8sManager.GetScheme(),
+			Kclient: kclient,
+			Log:     ctrl.Log.WithName("controllers").WithName("NovaMetadata"),
```
gibizer, line 104: This seems damaged during refactor. The NovaAPIReconciler now has a logger for NovaMetadata

controllers/utils.go
SeanMooney: oh i proably typed this instead of copy paste that is not intentional

main.go
```diff
@@ -107,89 +107,70 @@ func main() {
...
-		setupLog.Error(err, "unable to create controller", "controller", "NovaNoVNCProxy")
-		os.Exit(1)
 	}
-
+	for name, controller := range reconcilers {
+		if err = controller.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", name)
+			os.Exit(1)
+		}
+	}
```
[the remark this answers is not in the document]
SeanMooney: lost the comment but i confirmed that the behavior is to just generate the if block in the main.go as normal because i left the marker comment there.

if you remove the marker it will just omit the generated stub so it looks like we cannot break the code gen by refactoring like this. at least not in this case

test/functional/suite_test.go
```diff
@@ -92,12 +93,18 @@ var _ = BeforeSuite(func() {
 
 	kclient, err := kubernetes.NewForConfig(cfg)
 	Expect(err).ToNot(HaveOccurred(), "failed to create kclient")
+
+	// This should work but causes api db sync job to fail.
+	// controllers.SetupReconcilers(k8sManager, ctrl.Log.WithName("testSetup"))
```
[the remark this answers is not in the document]
SeanMooney: this was a simple case of not passing the config generated by envtest.
so the tests were running against crc and failing as this test could not run against a real env.

i have corrected that in the latest version by passing the config

we can figure out what to do with the test later since that is unrelated to this change.

SeanMooney:
this is going to fail since i have not asress the golint comments
i rebased this on the pre-commit change but forgot to update my hooks locally i bet that is the issue

test/functional/suite_test.go
```diff
@@ -72,9 +71,6 @@ var _ = BeforeSuite(func() {
 	Expect(cfg).NotTo(BeNil())
 
 	// NOTE(gibi): Need to add all API schemas our operator can own.
-	// Keep this in synch with NovaAPIReconciler.SetupWithManager,
-	// otherwise the reconciler loop will silently not start
-	// in the test env.
```
gibizer, line 77: Please keep this or something similar. When we add a new external dependency to nova-operator like mariadb here https://github.com/openstack-k8s-operators/nova-operator/pull/87/files#diff-9e438edde3e826eec37a179369bdcf4a85e4e6822f83bb1cad33d29097f26781R118 then we need to add the schema here otherwise test will start failing without really pointing to the missing schema here.

gibizer (dismissed):
/lgtm

main.go
```diff
@@ -103,74 +103,10 @@ func main() {
...
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "NovaCell")
-		os.Exit(1)
-	}
-	if err = (&controllers.NovaReconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		Kclient:        kclient,
-		Log:            ctrl.Log.WithName("controllers").WithName("Nova"),
-		RequeueTimeout: time.Duration(5) * time.Second,
```
gibizer, line 167: This was lost during the rebase / refactoring.

test/functional/suite_test.go
```diff
@@ -130,44 +128,7 @@ var _ = BeforeSuite(func() {
...
 
-	kclient, err := kubernetes.NewForConfig(cfg)
-	Expect(err).ToNot(HaveOccurred(), "failed to create kclient")
-
-	err = (&controllers.NovaAPIReconciler{
-		Client:                k8sManager.GetClient(),
-		Scheme:                k8sManager.GetScheme(),
-		Kclient:               kclient,
-		Log:                   ctrl.Log.WithName("controllers").WithName("NovaAPI"),
-		RequeueTimeoutSeconds: 1,
```
gibizer, line 141: After the refactor this (and below) is set back to 5 seconds making the overall test execution slower (from 26 sec to 70 sec). I'm OK to merge this as is and I will propose a follow up that moves the RequeueTimeout to the ReconcileBase and also expose the RequeueTimeout configuration for the test env.

gibizer (changes requested):
-1 due to lost RequeueTimeout config from NovaReconciler. 
The test configurability issue can be fixed in a followup

controllers/utils.go
```diff
@@ -0,0 +1,51 @@
...
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaMetadata", mgr, kclient),
+		},
+		"NovaNoVNCProxy": &NovaNoVNCProxyReconciler{
+			ReconcilerBase: nova_common.NewReconcilerBase("NovaNoVNCProxy", mgr, kclient),
+		},
+	}
+	for name, controller := range reconcilers {
+		if err = controller.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", name)
+			os.Exit(1)
```
gibizer, line 48: This is still not good for testing perspective. In the test we should not exit, but raise a panic to trigger cleanup.

gibizer: I will fix this in a followup

main.go
gibizer: actually I'm fixing this as well in a follow up

gibizer (approved):
All my comment is being fixed in a follow up so this can be merged.

PR #104, opened 2022-10-12
Flatten reconcile call and pull out step like functions

### LINKED ISSUE #88: rename reconcileNormal to doRconsile or a similar more descriptive name.

### LINKED ISSUE #97: Check if we can improve readbility be reducing number of function params
The one param per line comes from the python black style to me. I can let that go but what should we do with long lines? I need to wrap it at some point to fit into a screen :) What would be the place of the split? In your example you split at after some parameters in a single line then switched to one line per param for the rest of the params. 

Maybe we have too many params for a single function call and we should wrap the params to a single struct.

_Originally posted by @gibizer in https://github.com/openstack-k8s-operators/nova-operator/pull/87#discussion_r992191441_

gibizer:
1. Get rid of reconcile() and reconcileNormal() by flattening that step of the hierarchy
2. Pull out ensure* like step functions from Reconcile() to make that function a high level description of the steps we need to make the given CR reconciled.

Closes #88 
Closes #97


controllers/nova_controller.go
```diff
@@ -124,8 +123,42 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		return ctrl.Result{}, nil
+	}
+
+	result, err = r.ensureAPI(ctx, h, instance, cell0Template, cell0DB, apiDB)
+	if err != nil {
+		return result, err
+	}
+
+	return ctrl.Result{}, nil
 }
```
SeanMooney, line 162: this looks ok to me.

that said im really tempted to factor this out into a list of tuples [(function poirnt , tuple of args)]
then have a for loop over it which will centralise the cleanup. but its over kill here espcialy since fomr othe the function liek ensureApi are condition on the ready state.

controllers/nova_controller.go
```diff
@@ -174,13 +207,63 @@ func (r *NovaReconciler) initConditions(
...
+			targetCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DBReadyRunningMessage))
+		return ctrlResult, nil
+	}
+
+	instance.Status.Conditions.MarkTrue(targetCondition, condition.DBReadyMessage)
+
+	return ctrl.Result{}, nil
```
SeanMooney, line 263: i also feel like we can factor this out into a loop not sure it will be cleaner.
i think im going to try and revisit this after we have this merged when we have some more time like i did for the contoler regeistartion.

there is defintly common code that could be factored out but it might not be worth the effort so im going to think about this a little more before 
trying to factor it out.

controllers/novaapi_controller.go
```diff
@@ -361,6 +328,56 @@ func (r *NovaAPIReconciler) generateServiceConfigMaps(
...
+	} else if (ctrlResult != ctrl.Result{}) {
+		util.LogForObject(h, "Deployment in progress", instance)
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
+		// It is OK to return success as we are watching for Deployment changes
+		return ctrlResult, nil
+	}
```
SeanMooney, line 362: so one common patteren i am seeeing is we call a function and check if the error is nil and the result is an expected value with and have two data callbacks


so i think we coudl transform this to something liek sthis
```
myVar = HandelError(depl.CreateOrPatch(ctx, h), ctrl.Result{}, instance.Status.Conditions.Set, condition.FalseCondition(...), condition.FalseCondition(...) )

```
where Handel error is  somehting like this
```
HandelError(reult_tuple:(ctrl.Result, error), expected_result: ctrl.Result, error_fuction:interface{}, result_mismatch_payload:condition.FalseCondition, error_payload:condition.FalseCondition)
```
the problem is without thei callling panic or using a goto which is worse i dont see any way to have HandelError return form the calling function which
would still require uss to have it return and error and check that.

so for now i think we are stuck with this.

if i come up with something better ill let you know.

controllers/novaconductor_controller.go
```diff
@@ -127,7 +127,55 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}()
 
-	return r.reconcileNormal(ctx, h, instance)
+	// TODO(gibi): Can we use a simple map[string][string] for hashes?
+	// Collect hashes of all the input we depend on so that we can easily
+	// detect if something is changed.
+	hashes := make(map[string]env.Setter)
```
SeanMooney, line 133: ya i have been wondering the same thing.
im not entirly sure why it needs to be an env.Setter

controllers/novaconductor_controller.go
```diff
@@ -127,7 +127,55 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+	secretHash, result, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		// TODO(gibi): Add the rest of the secret fields here when they are
+		// needed for the controller to reconcile.
+		[]string{
+			// TODO(gibi): This is hardcode here until wil agree one the Secret
+			// handling tracked in issues/79
+			"NovaCell0DatabasePassword",
```
SeanMooney, line 143: ack

SeanMooney (approved):
i have a few comments inlien but nothing to prevent this  merging so approving ill hold on merging this until we sync in our call in a few minutes but if there are no objections or concerns ill merged this afterwards.

PR #107, opened 2022-10-13
config gen

SeanMooney:
- refactor config generation via common base class.


controllers/novaapi_controller.go
```diff
@@ -233,6 +222,27 @@ func (r *NovaAPIReconciler) reconcileNormal(
 		return ctrl.Result{}, err
 	}
 
+	// all our input checks out so report InputReady
+	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
```
gibizer, line 226: This used to be after the ensureSecret and before service config generation. Is it intentional that you moved it after generateConfig?

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
-			Namespace:     instance.Namespace,
-			Type:          util.TemplateTypeConfig,
-			InstanceType:  instance.Kind,
-			CustomData:    customData,
-			ConfigOptions: templateParameters,
-			Labels:        cmLabels,
-		},
-	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+	err := helper.GetClient().Get(ctx, secretName, secret)
```
gibizer, line 314: Let's modify ensureSecret to return the Secret object as it already queries it so you don't have to re-query it here.

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
 		return err
 	}
-
-	return nil
+	configMapVars := make(map[string]env.Setter)
+	templateParameters := map[string]interface{}{
+		"service_name":           "nova-api",
+		"keystone_internal_url":  instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":     instance.Spec.ServiceUser,
+		"nova_keystone_password": string(secret.Data[instance.Spec.PasswordSelectors.Service]),
```
gibizer, line 323: I think secrets are base64 encoded today.

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
+		"api_db_user":            instance.Spec.APIDatabaseUser,
+		"api_db_password":        string(secret.Data[instance.Spec.PasswordSelectors.APIDatabase]),
+		"api_db_adddress":        instance.Spec.APIDatabaseHostname,
+		"api_db_port":            3306,             // TODO lookup from service?
```
gibizer, line 327: I think it is OK to hardcode it now. If we need to make this configurable then we need to:
1. add a NovaDatabasePort to the NovaAPISpec (but not the NovaAPITemplate)
2. extrend the lib-common's Database struct to expose the port from the Service

But we talked about passing down specific secrets from Nova to NovaAPI with the DB user and password so we can actually but the whole DB URLs or the pieces of it there.

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
-	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+	err := helper.GetClient().Get(ctx, secretName, secret)
 	if err != nil {
 		return err
 	}
-
-	return nil
+	configMapVars := make(map[string]env.Setter)
+	templateParameters := map[string]interface{}{
```
gibizer, line 319: can we make this a map[string]string ?

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
-		},
-	}
-	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+	err := helper.GetClient().Get(ctx, secretName, secret)
 	if err != nil {
 		return err
 	}
-
-	return nil
+	configMapVars := make(map[string]env.Setter)
```
gibizer, line 318: This will get populated with the hash of the ConfigMaps created by lib-common we need to store those hashes in instance.Status to know when the config is changed so that when the service needs to be restarted. In the baseline `hashes` was passed in here from reconcile(). So I think we need to do that here as well instead of creating a local map here.

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
+	templateParameters := map[string]interface{}{
+		"service_name":           "nova-api",
+		"keystone_internal_url":  instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":     instance.Spec.ServiceUser,
+		"nova_keystone_password": string(secret.Data[instance.Spec.PasswordSelectors.Service]),
+		"api_db_user":            instance.Spec.APIDatabaseUser,
+		"api_db_password":        string(secret.Data[instance.Spec.PasswordSelectors.APIDatabase]),
```
gibizer, line 325: We need to sync what fields are used here from the secret and what fields are ensured to exists in the Secret by ensureSecret. Otherwise we can silently get an empty string as the value here if the field does not exists in the Secret.

pkg/common/config.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// GenerateConfigMaps helper function to generate config maps
+func (r *ReconcilerBase) GenerateConfigMaps(
+	ctx context.Context, h *helper.Helper,
+	instance client.Object, envVars *map[string]env.Setter,
+	templateParameters *map[string]interface{},
+	extraData *map[string]string,
+) error {
+	r.Log.Info("Reconciling Service")
```
gibizer, line 22: drop this.

pkg/common/config.go
```diff
@@ -0,0 +1,58 @@
...
+	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel("nova"), map[string]string{})
```
gibizer, line 23: This was 
```golang
cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(NovaAPILabelPrefix), map[string]string{})
```
I think  we want to keep the NovaAPI specific labeling. So either we need to move that constant to the ReconcilerBase or pass it as as parameter.

controllers/novaapi_controller.go
gibizer: Hm you even get the `hashes` there via `envVars` parameter. So you can just need to forward that to `GenerateConfigMaps` and not creating a new map here.

controllers/novaapi_controller.go
```diff
@@ -292,74 +302,117 @@ func (r *NovaAPIReconciler) reconcileNormal(
...
+		"api_db_adddress":        instance.Spec.APIDatabaseHostname,
+		"api_db_port":            3306,             // TODO lookup from service?
+		"openstack_cacert":       "",               // fixme
+		"openstack_region_name":  "regionOne",      // fixme
+		"default_project_domain": "Default",        // fixme
+		"default_user_domain":    "Default",        // fixme
+		"transport_url":          "rabbit://fixme", // fixme
+		"metadata_secret":        "42",             // fixme
```
gibizer, line 333: to be able to make nova-api work you need to pass down the cell db related configuration as well.

gibizer (changes requested):
The overall direction looks good. I left some comments that hopefully helps.

controllers/novaapi_controller.go
SeanMooney: we could i guess
i was pasing bools and ints when i was protyping but its going to end up in a config file at the end of the day so sure we can make it strings

controllers/novaapi_controller.go
SeanMooney: ya i did not have any of the hash stuff in my old protyp so this is jus told code i have not updated properly yet.

will do

controllers/novaapi_controller.go
SeanMooney: sure that works.
im already relying on it to verify all the fields i need are there which is why I'm not testing for their present later but by not actually using the one it verified it could have changed so it's also more correct to do that as well as more efficient.

controllers/novaapi_controller.go
SeanMooney: it is yes but when we get it back it's decoded.
that seems to be done transparently somewhere but looking at the data in the debugger and the rendered config it properly decoded. its a byte array which is why I'm calling string to convert it but its not base64 anymore

controllers/novaapi_controller.go
SeanMooney: yes the field im using now are already ensure but i am relying on that checking in ensure

controllers/novaapi_controller.go
SeanMooney: ya i wanted to get the config gen working with the current interface for now but i think when we start spelicing up the secure we should put the rendered URL in the one the internal crds used so we will refactor this later.

controllers/novaapi_controller.go
SeanMooney: ack

pkg/common/config.go
SeanMooney: yep old code will do

pkg/common/config.go
SeanMooney: this is something we will have to discuss in the future I'm not sure if we will want to allow the labels to be configurable but sure i want to revisit the siginture of this function later.

controllers/novaapi_controller.go
SeanMooney: i was minimisng the code that could fail while porting the code so no it has no significats that it moved ill restore it later

config/samples/nova_v1beta1_novaapi.yaml
```diff
@@ -19,22 +19,22 @@ spec:
...
       # my custom policy
   containerImage: quay.io/tripleowallabycentos9/openstack-nova-api:current-tripleo
   replicas: 1
   nodeSelector: {}
   resources:
     limits:
-      cpu: 500m
-      memory: 128Mi
+      cpu: 2.0
+      memory: 256Mi
```
SeanMooney, line 34: this needs to be higher if we are using apache. ill update this later.

gibizer: I'm OK to bump the memory needs if the config gen triggered OMM killer. But do we need 4 times the CPUs as well?

controllers/common.go
```diff
@@ -240,3 +242,36 @@ func (r *Reconcilers) OverriedRequeueTimeout(timeout time.Duration) {
...
+			ConfigOptions:      templateParameters,
+			Labels:             cmLabels,
+			CustomData:         extraData,
+			Annotations:        map[string]string{},
+			AdditionalTemplate: additionalTemplates,
+		},
+	}
+	err := configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
+	if err != nil {
+		return nil
```
SeanMooney: i fixed this in then next commit

SeanMooney (commented):
I am still reworking this in general

i have now got to the point where this is at least somewhat functional but i need to rework the tests.

config/samples/nova_v1beta1_novaapi.yaml
SeanMooney: PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                          
      1 root      20   0   23.9m   5.0m   1.8m S   0.0   0.0   0:12.00 httpd                                                                                                            
     11 apache    20   0   25.1m   4.5m   1.1m S   0.0   0.0   0:00.00  `- httpd                                                                                                        
     12 nova      20   0  265.6m 108.6m   1.3m S   0.0   0.7   0:17.69  `- httpd                                                                                                        
     13 apache    20   0 2109.8m  16.8m   1.3m S   0.0   0.1   1:18.14  `- httpd                                                                                                        
     20 apache    20   0 1981.7m  14.6m   1.3m S   0.0   0.1   1:11.19  `- httpd                                                                                                        
     21 apache    20   0 1917.7m  14.5m   1.3m S   0.0   0.1   1:11.82  `- httpd                                                                                                        
    293 apache    20   0 1917.7m  14.7m   1.4m S   0.0   0.1   1:11.89  `- httpd                                                                                                        
 272510 root      20   0    7.7m   4.0m   3.3m S   0.0   0.0   0:00.11 bash                                                                                                             
 277822 root      20   0   10.2m   3.7m   3.1m S   0.0   0.0   0:01.04  `- top    

ok so apache is using 2G of adress space but only ~70mb of resedent memory and nova then has about 110

so 256MB is tight but should be able to function.

512 is proably safer so ill do that and drop the cpu limit back to 0.5

ill increase teh memory request to 200Mi
that is a better estimate fo the baseline usage.

api/v1beta1/common_types.go
```diff
@@ -97,6 +97,7 @@ type PasswordSelector struct {
 	// password from the Secret
 	APIMessageBus string `json:"apiMessageBus,omitempty"`
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="NovaCell0DatabasePassword"
```
SeanMooney, line 100: this looks a little odd right now but when we have separate secrets this name will be something like CellDatabasePassword

alternatively i could leave this without a default but then change the test code to set it.

everything except a compute service will have this set in the secret and use it

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -44,11 +44,11 @@ spec:
...
       limits:
-        cpu: 500m
-        memory: 128Mi
+        cpu: 0.5
+        memory: 512Mi
       requests:
-        cpu: 10m
-        memory: 64Mi
+        cpu: 0.1
+        memory: 200Mi
```
SeanMooney, line 51: hopefully you are happier with these as a default for now.

honestly however in a real deployment under load i would expect this to be higher.

gibizer (commented):
I have couple of things inline most small cleanups and request for follow up. 

I rebased this top of my series here https://github.com/gibizer/nova-operator/tree/pr-107-sean-config-gen and tried to deploy it. It seems something is wrong with the keystone auth of nova-api as I cannot query flavor list successfully.

```
(.venv) [fedora@gibi-f35 ~]$ openstack flavor list
HttpException: 503: Server Error for url: http://nova-public-openstack.apps-crc.testing/v2.1/flavors/detail?is_public=True, The server is currently unavailable. Please try again at a later time.<br /><br />
The Keystone service is temporarily unavailable.


2022-11-11 09:50:24.934 12 DEBUG nova.api.openstack.wsgi [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] Calling method '<bound method Versions.index of <nova.api.openstack.compute.versions.Versions object at 0x7fed237a9fd0>>' _process_stack /usr/lib/python3.9/site-packages/nova/api/openstack/wsgi.py:513
2022-11-11 09:50:24.935 12 INFO nova.api.openstack.requestlog [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] 192.168.130.1 "GET /" status: 200 len: 415 microversion: - time: 0.001104
2022-11-11 09:50:24.943 12 WARNING keystonemiddleware.auth_token [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] Using the in-process token cache is deprecated as of the 4.2.0 release and may be removed in the 5.0.0 release or the 'O' development cycle. The in-process cache causes inconsistent results and high memory usage. When the feature is removed the auth_token middleware will not cache tokens by default which may result in performance issues. It is recommended to use  memcache for the auth_token token cache by setting the memcached_servers option.
2022-11-11 09:50:24.975 12 WARNING keystonemiddleware.auth_token [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] Identity response: {"error":{"code":401,"message":"The request you have made requires authentication.","title":"Unauthorized"}}
: keystoneauth1.exceptions.http.Unauthorized: The request you have made requires authentication. (HTTP 401) (Request-ID: req-b19458f1-d23e-43f2-a781-314d75480c3e)
2022-11-11 09:50:25.001 12 WARNING keystonemiddleware.auth_token [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] Identity response: {"error":{"code":401,"message":"The request you have made requires authentication.","title":"Unauthorized"}}
: keystoneauth1.exceptions.http.Unauthorized: The request you have made requires authentication. (HTTP 401) (Request-ID: req-11cf9a7c-98e7-471c-af3e-d14aa9c6cd1c)
2022-11-11 09:50:25.006 12 CRITICAL keystonemiddleware.auth_token [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] Unable to validate token: Identity server rejected authorization necessary to fetch token data: keystonemiddleware.auth_token._exceptions.ServiceError: Identity server rejected authorization necessary to fetch token data
2022-11-11 09:50:25.009 12 INFO nova.api.openstack.requestlog [req-7590b5c4-185f-48d7-95aa-baa178d73d3a - - - - -] 192.168.130.1 "GET /v2.1/flavors/detail?is_public=True" status: 503 len: 218 

[fedora@gibi-f35 install_yamls]$ oc rsh -c nova-api-api pod/nova-api-0
sh-5.1# cat /etc/nova
nova/     novajoin/ 
sh-5.1# cat /etc/nova/
api-paste.ini  nova.conf      policy.json    release        rootwrap.conf  
sh-5.1# cat /etc/nova/nova.conf 
[DEFAULT]


state_path = /var/lib/nova

# enable log rotation in oslo config by default
max_logfile_count=5
max_logfile_size_mb=50
log_rotation_type=size

log_file = /var/log/nova/nova-api.log

debug=true
# ensure safe defaults for new hosts

# use predictable instnace names for simpler debuging.
instance_name_template="instance-%(uuid)s"

force_config_drive=True
transport_url=rabbit://fixme


osapi_compute_workers=1
metadata_workers=1


[oslo_concurrency]
lock_path = /var/lib/nova/tmp

[oslo_messaging_rabbit]
amqp_durable_queues=true
amqp_auto_delete=true
# we should consider using quorum queues instead
# rabbit_quorum_queue=true


heartbeat_in_pthread=true



[oslo_policy]
enforce_new_defaults=true









[api]
use_forwarded_for = true
[oslo_middleware]
enable_proxy_headers_parsing = True
[wsgi]
api_paste_config = /etc/nova/api-paste.ini


[upgrade_levels]
compute = auto

driver = noop


[notifications]




[cache]
# always enable caching
enabled = True

# on compute nodes or where memcache is not deployed we should use an in memory
# dict cache
backend = oslo_cache.dict


[workarounds]
disable_fallback_pcpu_query=true
enable_qemu_monitor_announce_self=true
reserve_disk_resource_for_image_cache=true

# never_download_image_if_on_rbd=true
# disable_group_policy_check_upcall=true
# wait_for_vif_plugged_event_during_hard_reboot=["normal"]
# we might want to enable these for upgrades.
# disable_compute_service_check_for_ffu=true
# skip_cpu_compare_on_dest=true
# skip_hypervisor_version_check_on_lm=true






[database]
connection = mysql+pymysql://nova_cell0:12345678@openstack/nova_cell0




[api_database]
connection = mysql+pymysql://nova_api:12345678@openstack/nova_api


[keystone_authtoken]
service_type = compute
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url =  http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = nova
username = nova
password = 12345678
cafile = 
region_name = regionOne

[placement]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url =  http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = nova
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal

[glance]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url = http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = nova
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal


[neutron]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url = http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = nova
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal
metadata_proxy_shared_secret = 42
service_metadata_proxy = true
```

Without the config gen series I have the following config where flavor list works

```
[fedora@gibi-f35 install_yamls]$ oc rsh pod/nova-api-0
Defaulted container "nova-api-api" out of: nova-api-api, init (init)
sh-5.1# egrep -e "^[^#]" /etc/nova/nova.conf
[DEFAULT]
log_config_append = /etc/nova/logging.conf
[api]
auth_strategy = keystone
[api_database]
connection = mysql+pymysql://nova_api:12345678@openstack/nova_api
[barbican]
[cache]
[cinder]
[compute]
[conductor]
[console]
[consoleauth]
[cors]
[cyborg]
[database]
connection = mysql+pymysql://nova_cell0:12345678@openstack/nova_cell0
[devices]
[ephemeral_storage_encryption]
[filter_scheduler]
[glance]
[guestfs]
[healthcheck]
[hyperv]
[image_cache]
[ironic]
[key_manager]
[keystone]
[keystone_authtoken]
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url = http://keystone-public-openstack.apps-crc.testing
auth_type = password
password = 12345678
[libvirt]
[metrics]
[mks]
[neutron]
[notifications]
[oslo_concurrency]
[oslo_messaging_amqp]
[oslo_messaging_kafka]
[oslo_messaging_rabbit]
[oslo_middleware]
[oslo_policy]
[pci]
[placement]
[powervm]
[privsep]
[profiler]
[quota]
[rdp]
[remote_debug]
[scheduler]
[serial_console]
[service_user]
[spice]
[upgrade_levels]
[vault]
[vendordata_dynamic_auth]
[vmware]
[vnc]
[workarounds]
[wsgi]
[zvm]
sh-5.1# 
```

controllers/common.go
```diff
@@ -242,3 +244,31 @@ func (r *Reconcilers) OverriedRequeueTimeout(timeout time.Duration) {
...
 }
+
+// GenerateConfigMaps helper function to generate config maps
+func (r *ReconcilerBase) GenerateConfigMaps(
+	ctx context.Context, h *helper.Helper,
+	instance client.Object, envVars *map[string]env.Setter,
+	templateParameters map[string]interface{},
+	extraData map[string]string, cmLabels map[string]string,
+) error {
+	r.Log.Info("Reconciling Service")
```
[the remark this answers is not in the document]
SeanMooney: oh this is an old old log i can drop that.

this is form my original commit before we merged being able to create the API pods via a deployemnt

pkg/novaapi/deployment.go
```diff
@@ -114,7 +97,7 @@ func StatefulSet(
 	envVars := map[string]env.Setter{}
 	envVars["KOLLA_CONFIG_FILE"] = env.SetValue(MergedServiceConfigPath)
 	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
-	// NOTE(gibi): The stateafulset does not use this hash directly. We store it
+	// NOTE(gibi): The deployment does not use this hash directly. We store it
```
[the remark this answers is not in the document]
SeanMooney: rebase artificat

pkg/novaapi/deployment.go
```diff
@@ -137,12 +120,29 @@ func StatefulSet(
 				},
 				Spec: corev1.PodSpec{
 					ServiceAccountName: nova.ServiceAccount,
-					Volumes: nova.GetVolumes(
-						nova.GetScriptConfigMapName(instance.Name),
+					Volumes: nova.GetOpenstackVolumes(
 						nova.GetServiceConfigConfigMapName(instance.Name),
 					),
-					InitContainers: initContainer(initContainerDetails),
+					// InitContainers: initContainer(initContainerDetails),
```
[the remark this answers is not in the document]
SeanMooney: right we dont that just somethign i forgot to remove will do.

pkg/novaapi/deployment.go
```diff
@@ -137,12 +120,29 @@ func StatefulSet(
...
+								RunAsUser: &runAsUser,
+							},
+							Env:            env,
+							VolumeMounts:   nova.GetOpenstackVolumeMounts(),
+							Resources:      instance.Spec.Resources,
+							ReadinessProbe: readinessProbe,
+							LivenessProbe:  livenessProbe,
+						},
 						{
 							Name: instance.Name + "-api",
```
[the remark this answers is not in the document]
SeanMooney: yep can do

i had considerd just calling it "api" and "logs"

but i forgot to ask you about this.

ill go with what you suggest.

pkg/novaapi/deployment.go
```diff
@@ -120,12 +120,29 @@ func StatefulSet(
 				},
 				Spec: corev1.PodSpec{
 					ServiceAccountName: nova.ServiceAccount,
-					Volumes: nova.GetVolumes(
-						nova.GetScriptConfigMapName(instance.Name),
+					Volumes: nova.GetOpenstackVolumes(
```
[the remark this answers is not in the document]
SeanMooney: ack

pkg/nova/volumes.go
```diff
@@ -106,15 +107,40 @@ func GetServiceVolumeMounts() []corev1.VolumeMount {
...
+			Name:      logVolume,
+			MountPath: "/var/log/nova",
+			ReadOnly:  false,
+		},
+	}
+}
+
+// GetOpenstackVolumes - returns the volumes used for the service deployment and for
+// any jobs needs access for the full service configuration
+func GetOpenstackVolumes(serviceConfigConfigMapName string) []corev1.Volume {
```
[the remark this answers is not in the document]
SeanMooney: ack i can move it back to the refactor commit

templates/novaapi/config/httpd.conf
```diff
@@ -13,33 +13,29 @@ Listen 0.0.0.0:8778
...
 <VirtualHost *:8778>
   <IfVersion >= 2.4>
     ErrorLogFormat "%M"
   </IfVersion>
-  ErrorLog /dev/stdout
   SetEnvIf X-Forwarded-For "^.*\..*\..*\..*" forwarded
+  ErrorLog /dev/stdout
   CustomLog /dev/stdout combined env=!forwarded
   CustomLog /dev/stdout proxy env=forwarded
+  LogLevel debug
```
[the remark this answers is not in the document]
SeanMooney: we actully could remove both i might just do that since this was me attempting to get debug logs to show up via appache before i found out your not meant to log to standard out/error with wsgi.

I'm not sure we need the apache log to be in debug mode

controllers/common.go
```diff
@@ -242,3 +244,31 @@ func (r *Reconcilers) OverriedRequeueTimeout(timeout time.Duration) {
...
+			Type:               util.TemplateTypeConfig,
+			InstanceType:       instance.GetObjectKind().GroupVersionKind().Kind,
+			ConfigOptions:      templateParameters,
+			Labels:             cmLabels,
+			CustomData:         extraData,
+			Annotations:        map[string]string{},
+			AdditionalTemplate: additionalTemplates,
+		},
+	}
+	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
```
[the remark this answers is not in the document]
SeanMooney: ok i can work on that now i was going to do it in a followup because i didn't think that was supported in lib common yet

SeanMooney:
thats just a rebase on top of master i still have to address gibi feedback

templates/novaapi/config/httpd.conf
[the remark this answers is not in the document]
SeanMooney: i have added some comments but to answer your original question the previous log level is the default log level for the apache server 
where as this is scoped to this vhost.

in this case it really does not matter since we are only serving one endpoint
if we were serving multiple websites form the same server/port then line 23 would be what they logged at unless they also overrode it in tehr virtual host file.

controllers/common.go
SeanMooney: im moving things anyway so ill just pull this back too

controllers/common.go
```diff
@@ -242,3 +244,36 @@ func (r *Reconcilers) OverriedRequeueTimeout(timeout time.Duration) {
 		reconciler.SetRequeueTimeout(timeout)
 	}
 }
+
+// GenerateConfigMaps helper function to generate config maps
+func (r *ReconcilerBase) GenerateConfigMaps(
```
SeanMooney, line 249: im goign to rename this to GenerateConfigs
because wether this is stored in a configmap or a secret is an implementation detail that we are planning to change anyway.

controllers/novaapi_controller.go
SeanMooney: I tried to do this but libcommon is using 
map[string]interface{} so i get type error later

I can probably cast it or force it somehow but for not I'm going to keep it as is and we can change it if we update libcommon

pkg/novaapi/deployment.go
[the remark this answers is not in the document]
SeanMooney: fixed i also fixed the typo

stateafulset -> statefulset

templates/novaapi/config/httpd.conf
```diff
@@ -13,33 +13,29 @@ Listen 0.0.0.0:8778
...
 
   ## WSGI configuration
   WSGIProcessGroup nova-api
   WSGIApplicationGroup %{GLOBAL}
   WSGIPassAuthorization On
-  WSGIDaemonProcess nova-api processes=3 threads=1 user=nova group=nova
+  WSGIDaemonProcess nova-api processes=1 threads=1 user=nova group=nova display-name=nova-api
```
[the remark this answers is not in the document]
SeanMooney: nope we give that up if we use apache

we get one request per process/thread
so in this case we only get 1 request per pod.

that why using nova-api directly without apache would perform better or using uwsgi.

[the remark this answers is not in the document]
gibizer: This is OK now, we can get back to this later.

SeanMooney:
so i think i figured out that happened with the keystone auth

in my env the keystone operator did not create the nova user/project
so my config assumes that that will be created 

in my env the endpoints were also not created so i need to redeploy with a newer version fo the keystone operator to confirm but that might be the issue.

my config is assuming we will have a nova project where as you were using service.

so the keystone operator might be adding a nova user to the service project in your env but I'm generating config expecting it to be nova?

ill try and  redeploy with a newer keystone operator but if i create the project/user and grant it the admin role then create the endpoints it worked fine for me locally.

so i think the nova config is correct but its just not in sync with what the keystone operator is doing.

SeanMooney:
ok figured it out

https://github.com/openstack-k8s-operators/keystone-operator/blob/0d929fcf931fb4f3567066a9ad127ffc132dea40/controllers/keystoneservice_controller.go#L414-L422

the keystone operator is currently hardcoded to create all service users in a shared service project and my config gen series assumes each service user will have a separate service project with the same name as the service user i.e. nova user in nova project vs nova user in service project

ill update this and that should resolve the auth issue.

gibizer (approved):
OK. I deployed this and now the nova-api works. So this good to go.

PR #108, opened 2022-10-14
[Conductor]Cleanup DB sync job at the right time

gibizer:
There was a lib-common issue where job.DoJob() deleted the Job right after it is succeeded. Then the controller tried to store the hash of the Job in Status. But if that failed (due to e.g. Conflict) then in the next Reconcile call lib-common would re-create and re-run the Job as the hash of the previous Job and the Job itself is lost.

The lib-common is fixed in a way that the automatic Job deletion is removed from DoJob() and that needs to be now done on the caller side via the new job.DeleteAllSuccededJobs() call.

This patch adapts to the new lib-common version by doing the Job cleanup right after the Status.Update() succeeded in the deferred call at the end of the Reconcile() run.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/77

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,3 +86,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/nova-operator/api => ./api
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/gibizer/lib-common/modules/common v0.0.0-20221014143800-c6662e4a62ff
--- a/controllers/novaconductor_controller.go
+++ b/controllers/novaconductor_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -86,3 +86,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/nova-operator/api => ./api
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/gibizer/lib-common/modules/common v0.0.0-20221014143800-c6662e4a62ff
```
gibizer, line 90: This needs to be dropped when lib-common change lands

gibizer:
So after this fix was still able to reproduce occasional double Job execution. It turned out that we can read stale data from a cache in operator-sdk. So simply moving the deletion logic after the Job hash is persisted is not enough. @abays suggested to simply delay the Job deletion by setting ttlSecondsAfterFinished on the Job. So with a big enough ttl value we can circumvent this issue.

PR #109, opened 2022-10-14
add simple unit test.

SeanMooney:


SeanMooney:
im going to abandon this for now but we can revisit this if we want to explore this later

Issue #106, opened 2022-10-12
[envtest]Every Update() call should have a retry on conflict


gibizer:
We have some flaky test cases where we update Job or Deployment Status to simulate success. However we can race with the nova-operator, that is also updating those entities in parallel. Which can lead to conflict response to in the test. (In the controller too but that is already handled via requeuing the reconciliation)

PR #111, opened 2022-10-15
[envtest] Stabilize flaky tests

gibizer:
We had multiple unstable test case. This series fixes what we can. 

Closes: #106

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -140,6 +140,7 @@ func ExpectCondition(
 	conditionType condition.Type,
 	expectedStatus corev1.ConditionStatus,
 ) {
+	logger.Info("ExpectCondition", "type", conditionType, "expected status", expectedStatus, "on", name)
 	Eventually(func(g Gomega) {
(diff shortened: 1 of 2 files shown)


test/functional/base_test.go
```diff
@@ -94,9 +94,9 @@ func DeleteNovaAPI(name types.NamespacedName) {
 		if k8s_errors.IsNotFound(err) {
 			return
 		}
-		Expect(err).Should(BeNil())
+		g.Expect(err).Should(BeNil())
```
SeanMooney, line 97: if this was nova i woudl wonder if we should have a hacking check to ensure that where we have a Gomega object in scope we use it when adding asserts like this to ensure the failures are captures properly

SeanMooney (approved):

test/functional/base_test.go
gibizer: I had the same thought. I had not time to research if there is a way to extend some of our linters to catch this. Opened a ticket https://github.com/openstack-k8s-operators/nova-operator/issues/121

PR #112, opened 2022-10-15
Reduce status update conflict

gibizer:
Do couple of improvement that makes Status update conflict less likely:
- Only update Status at the end of Reconcile
- Replace Update() with Patch()
- Report error from deferred Status update
- Remove unnecessary RequeueAfter

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -157,15 +157,6 @@ func (r *NovaReconciler) initConditions(
 			),
 		)
 		instance.Status.Conditions.Init(&cl)
-
-		// Register overall status immediately to have an early feedback e.g.
-		// in the cli
-		if err := r.Client.Status().Update(ctx, instance); err != nil {
-			util.LogErrorForObject(
-				h, err, "Failed to initialize Conditions", instance)
-			return err
-		}
-
 	}
 	return nil
 }
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -264,7 +264,9 @@ func (r *NovaReconciler) reconcileNormal(
 	// cell0 is ready as top level services needs cell0 to register in
 	cell0ReadyCond := cell0.Status.Conditions.Get(condition.ReadyCondition)
 	if cell0ReadyCond == nil || cell0ReadyCond.Status != corev1.ConditionTrue {
-		return ctrl.Result{RequeueAfter: r.RequeueTimeout}, nil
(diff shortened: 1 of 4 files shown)


controllers/nova_controller.go
```diff
@@ -102,13 +103,26 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 				condition.ReadyCondition, condition.ReadyMessage)
 		}
-		err := r.Client.Status().Update(ctx, instance)
-		if err != nil && !k8s_errors.IsNotFound(err) {
-			util.LogErrorForObject(
-				h, err, "Failed to update status at the end of reconciliation", instance)
+		if err := h.SetAfter(instance); err != nil {
+			util.LogErrorForObject(h, err, "Set after and calc patch/diff", instance)
+			_err = err
+			return
```
SeanMooney, line 109: note to self since result is never initalised this will have the effect of returnign nil, err

controllers/nova_controller.go
```diff
@@ -255,7 +260,9 @@ func (r *NovaReconciler) reconcileNormal(
 	// cell0 is ready as top level services needs cell0 to register in
 	cell0ReadyCond := cell0.Status.Conditions.Get(condition.ReadyCondition)
 	if cell0ReadyCond == nil || cell0ReadyCond.Status != corev1.ConditionTrue {
-		return ctrl.Result{RequeueAfter: r.RequeueTimeout}, nil
+		// It is OK to return success as NovaCell expected to change to Ready
+		// and we are watching NovaCell
+		return ctrl.Result{}, nil
 	}
```
SeanMooney, line 266: ok so since we are watching nova cell when it updates we will reconsile and run though this again

controllers/nova_controller.go
```diff
@@ -376,7 +383,7 @@ func (r *NovaReconciler) reconcileNovaAPI(
 		},
 	}
 
-	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, api, func() error {
+	op, err := controllerutil.CreateOrPatch(ctx, r.Client, api, func() error {
```
SeanMooney, line 386: im guessing the delta is that patch will only modify the subfiled that have changed rahter then a fully update
yes?

SeanMooney (approved):

controllers/nova_controller.go
gibizer: yes that is the idea.

SeanMooney (approved):

Issue #99, opened 2022-10-11
When PreserveJobs set to False from True then the controller should delete the existing failed Jobs.


gibizer:
im not partically happy with supporting this.
but im glad you at least have tests for it

if we do support this when its set back to false it shoudl delete all jobs that were created while the deleteion is disabeld.

_Originally posted by @SeanMooney in https://github.com/openstack-k8s-operators/nova-operator/pull/87#discussion_r991382721_

PR #115, opened 2022-10-15
[envtest]Verify PreserveJobs behavior

gibizer:
This patch adds a test case to verify that if NovaConductor is configured to preserve jobs at create but then later reconfigured to not preserve them then old jobs are deleted.

Closes: #99

### DIFF
--- a/test/functional/novaconductor_controller_test.go
+++ b/test/functional/novaconductor_controller_test.go
@@ -443,4 +443,54 @@ var _ = Describe("NovaConductor controller", func() {
 		})
 	})
 
+	When("PreserveJobs changed from true to false", func() {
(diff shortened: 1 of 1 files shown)


test/functional/novaconductor_controller_test.go
```diff
@@ -443,4 +443,54 @@ var _ = Describe("NovaConductor controller", func() {
 		})
 	})
 
+	When("PreserveJobs changed from true to false", func() {
+		var jobName types.NamespacedName
+
+		BeforeEach(func() {
```
SeanMooney, line 449: im not really sure you need a before each in this case it could all be in the It but ok

SeanMooney (approved):

PR #118, opened 2022-10-19
Provide a working example Nova CR for install_yamls

gibizer:
The nova support in install_yamls [1] will use nova_v1beta1_nova.yaml to deploy Nova by default. So make that sample correct.

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/27


SeanMooney (approved):
approving for now.
we likely will revert this to using separate db instance in the future although the cell0 db and API db would be ok to share in production.

cell 1 should always be separate in any real deployment so this need to be fix before any release can happen.

PR #120, opened 2022-10-21
[Nova] Support multiple cells

gibizer:
This patch generalize the cell support in the Nova controller. So that we support any number of cells not just cell0.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -32,6 +32,7 @@ import (
 	database "github.com/openstack-k8s-operators/lib-common/modules/database"
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/nova-operator/pkg/nova"
+	corev1 "k8s.io/api/core/v1"
 
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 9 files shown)


controllers/nova_controller.go
```diff
@@ -158,6 +160,49 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return result, err
 	}
 
+	// NOTE(gibi): We do a sequenctial approach to create Cell DBs here. I.e.
+	// we create the DB of the first cell and not move forward to the second
+	// cell DB until the first is created. So if the DB creation fails for the
+	// first cell then we never try to create DB for the second cell.
+	// This is suboptimal but simple.
```
gibizer, line 167: So we agreed on the Today's sync to make this parallel too. The sequence should be:
1. kick off all the DB creation including cell0 in a loop but don't return on error just collect the status of the DBs
2. kick off all the Cell creation including cell0 in a loop if the related DB is ready (if not then skip it over) but don't return on error just collect the status of the Cells.
3. if NovaCell cell0 is ready then create NovaAPI.

gibizer: done

controllers/nova_controller.go
```diff
@@ -130,34 +131,158 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if apiDBStatus == nova.DBFailed {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			novav1.NovaAPIDBReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityError,
+			condition.DBReadyErrorMessage,
+			apiDBError.Error(),
+		))
+	}
+	if apiDBStatus == nova.DBCreating {
```
kk7ds, line 151: This means we started the job successfully right? And the case above means we failed to even start it, and the case below means it is already done from a previous run I guess?

controllers/nova_controller.go
```diff
@@ -130,34 +131,158 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+			novav1.NovaAllCellsDBReadyCreatingMessage,
+			strings.Join(creatingDBs, ",")))
+	} else { // we have no DB in failed or creating status so all DB is ready
+		instance.Status.Conditions.MarkTrue(novav1.NovaAllCellsDBReadyCondition, condition.DBReadyMessage)
 	}
 
-	cell0, result, err := r.ensureCell0(ctx, h, instance, cell0Template, cell0DB, apiDB)
-	if err != nil {
-		return result, err
+	// Kick of the creation of Cells. We skip over those cells where the cell
```
kk7ds, line 199: "off"

controllers/nova_controller.go
gibizer: - DBFailed means we either failed to start the Job or the Job finished with non zero exit
- DBCreating means the Job is started but not finished yet.
- DBCompleted means the Job finished with a zero exit code.

controllers/nova_controller.go
SeanMooney: we chated about this and said we would address this in a followup change at soempoitn and move forward with this. but use it should be off.

SeanMooney (approved):

Issue #76, opened 2022-10-03
NovaAPI should use StatefulSet instead of Deployment


gibizer:
To avoid the need to service de-registration at any Deployment restart event we should use StatefulSet instead to keep the hostnames stable. So that we only need to handle service de-registration at a real scale down event.

PR #122, opened 2022-10-24
[NovaAPI]Switch from Deployment to StatefulSet

gibizer:
To reduce the need of service de-registration when the nova-api
deployment is changed this patch switch to StatefulSet to make the
nova-api replicas "stateful", i.e. ensure they have a stable hostname

Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/76


go.mod
```diff
@@ -86,3 +86,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/nova-operator/api => ./api
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/gibizer/lib-common/modules/common v0.0.0-20221024131128-6f34ff3e8f87
```
gibizer, line 90: Drop this when https://github.com/openstack-k8s-operators/lib-common/pull/82 lands

gibizer: landed. dropped.

gibizer:
I rebased this top of master as it is independent from #120

config/rbac/role.yaml
```diff
@@ -8,7 +8,7 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - deployments
```
SeanMooney, line 11: i wonder if we will endup with both eventually.
probably not in the case of nova since we want to avoid having to deregister them if a pod is killed.

for placement using deployments is actually more correct then stateful sets so we probably wont. make the same change there.

controllers/novaapi_controller.go
```diff
@@ -338,10 +338,10 @@ func (r *NovaAPIReconciler) ensureDeployment(
...
 	}
 
-	depl := deployment.NewDeployment(novaapi.Deployment(instance, inputHash, serviceLabels), 1)
-	depl.SetTimeout(r.RequeueTimeout)
-	ctrlResult, err := depl.CreateOrPatch(ctx, h)
-	if err != nil {
+	ss := statefulset.NewStatefulSet(novaapi.StatefulSet(instance, inputHash, serviceLabels), 1)
+	ss.SetTimeout(r.RequeueTimeout)
+	ctrlResult, err := ss.CreateOrPatch(ctx, h)
+	if err != nil && !k8s_errors.IsNotFound(err) {
```
SeanMooney, line 344: was this a latenet bug before.

we would have taken this path and run the reconcile again.

that kind of explains why i was seeing it sometime run twice locally when i created API CRs

i take it we actually get k8s_errors.IsNotFound(err) when CreateOrPatch creates the statefulset

test/functional/base_test.go
```diff
@@ -559,3 +559,29 @@ func CreateNovaSecret(namespace string, name string) *corev1.Secret {
 	Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
 	return secret
 }
+
+func GetStatefulSet(name types.NamespacedName) *appsv1.StatefulSet {
+	ss := &appsv1.StatefulSet{}
+	Eventually(func(g Gomega) {
```
SeanMooney, line 565: i have been thinking about this by the way

we should not do 
. "github.com/onsi/ginkgo/v2"
. "github.com/onsi/gomega"

when we import them

if do

ginko "github.com/onsi/ginkgo/v2"
gomega "github.com/onsi/gomega"

we can have a hacking rule make sure you never do 
gomega.Expect and instead so that we remember to pass in g Gomega
and do g.Expect to get the correct propacation.

I'm not sure if we want to enforce that but it might be nice to not polite our current scope by importing them unnamed.

just commitng here so you can think about it for the future but i think that might be something we will want to add to the best practices.

so there are kind of two best practices there i think
1 only import named package i.e. never do ". some.package.name"
2 for Gomega always pass use g.<thing> instead of gomega.thing if we have a gomega object in scope

SeanMooney (approved):

config/rbac/role.yaml
gibizer: yeah, probably won't need it. If we need it we can add it though.

PR #124, opened 2022-10-25
Return by pointer for consistency

bogdando:
Signed-off-by: Bogdan Dobrelya <bogdando@yahoo.com>

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -187,20 +187,20 @@ func ExpectConditionWithDetails(
 	logger.Info("ExpectConditionWithDetails succeeded", "type", conditionType, "expected status", expectedStatus, "on", name)
 }
 
-func GetConfigMap(name types.NamespacedName) corev1.ConfigMap {
+func GetConfigMap(name types.NamespacedName) *corev1.ConfigMap {
 	cm := &corev1.ConfigMap{}
 	Eventually(func(g Gomega) {
 		g.Expect(k8sClient.Get(ctx, name, cm)).Should(Succeed())
 	}, timeout, interval).Should(Succeed())
-	return *cm
+	return cm
 }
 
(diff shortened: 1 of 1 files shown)


bogdando:
@SeanMooney @gibizer PTAL

bogdando:
Actually, we must reverse, see the testing
by pointer https://go.godbolt.org/z/jaccozvT6
smart https://go.godbolt.org/z/99Wjf8nqf <<< use THIS

bogdando:
I understand that the example is simple, and for complex structures/interfaces results may vary, but here is that I see for the simple case https://go.godbolt.org/z/Tf8GE5ebK which makes me thinking we should not spend much time on this for test code, but worth coming back to this for other places we use the pattern 1 (in the linked code) instead of 2 or 3. And there is multiple places really @dprince WDYT?

bogdando:
sean-k-mooney: in the value casie its shorter because its  on the stack where as the pointr case is not. but when we call the k8s client we knwo that value will not be stored on the stack

so we can stick with that change, whatever **a consistent** approach looks better for us

gibizer:
Let's merge this to have **consistency** across our test helpers. 

Regarding performance.
1. I don't think we should decide on which is more performant without context. So if we think we need to improve performance then profile the controller (or the test) and see where is the bottleneck and improve on the bottleneck. In all other cases use what is more consistent, easier to read, etc.
2. Note that operator-sdk generates DeepCopy for the structs of our CRDs with a pointer receiver. So in the cases where DeepCopy is needed (e.g. when calling the client provided by the sdk) we are forced to use pointer types.
3. I don't see how stale cache is related to our pointer usage. The cache is hidden from us and the client returns an err. So it is not the case that we cannot differentiate between found or not found resources.

gibizer (approved):
Let's merge this to have consistency across the helpers.

PR #128, opened 2022-11-02
limit test processes

SeanMooney:


Makefile
```diff
@@ -103,9 +103,12 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+PROCS?=$(shell expr $(shell nproc --ignore 2) / 2)
```
gibizer, line 106: this can be 0 as `expr 1 / 2` is 0. I checked, ginkgo does not fail with --procs 0 today but uses a single thread. So it is OK.

gibizer (approved):

Makefile
SeanMooney: ah, good catch ya i probably should have min(1, x/2) if i wanted to harden against that.
im not sure how to do that off the top of my head in bash but if this becomes an issue in the future we can figure it out then.

gibizer: yeah, lets worry about it when ginko hardens the CLI :)

Issue #130, opened 2022-11-03
Start using TransportURL CR to provide message bus access to the nova services


gibizer:
Work is ongoing to have a TransportURL CR https://github.com/openstack-k8s-operators/openstack-operator/pull/27 we can use for rabbit access. After that code is stabilized we can start using them

Issue #129, opened 2022-11-03
Allow passing in placement, neutron, cinder, glance, etc creds to Nova


gibizer:
Nova uses a bunch of other OpenStack services to function. To be able to call those services nova needs their user and password configured. 
- Extend our CRDs to allow passing in the necessary users and passwords
- Pass those values to the different service CRs they need them
- Use those values in the config generation of the nova services

PR #131, opened 2022-11-03
[Conductor]Deploy nova-conductor in a StatefulSet

gibizer:
To have a non crashing nova-conductor we need couple of things:
- nova service user in keystone so that conductor can query endpoints from keystone when starts up. So such user is registered by Nova controller via creating the KeystoneService CR. As KeystoneService CR automatically gets Nova's finalizers (to avoid accidental deletion) Nova controller needs a finalizer on Nova too to be able to remove the finalizer from KeystoneService during Nova deletion.
- placement user and password configured so that conductor can query placement at startup. Today we don't have a way in the nova CRD to pass in the placement user and and password. So the controller hardcodes these values to the know values from install_yamls. This is tracked in #129

The conductor still not have a message bus connection configured. That is a future TODO tracked in #130


controllers/nova_controller.go
```diff
@@ -124,13 +144,46 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		}
 	}()
 
+	if !instance.DeletionTimestamp.IsZero() {
```
SeanMooney, line 147: i guess this is ok but i kind fo feel like this shoudl be one of the first thigns we do.
i woudl prefer if we did this on line 97

is there a resason we are doing this after the status init and defered update calls.

SeanMooney: it is proably safe to run this as is as you are only removing the finaliser and doing the keystone cleanup
so no nova CRs are actully deleted via the reconcilDelete function

controllers/nova_controller.go
```diff
@@ -562,11 +624,121 @@ func (r *NovaReconciler) ensureAPI(
...
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+
+	if k8s_errors.IsNotFound(err) {
+		// Nothing to do as it was never created
+		return nil
+	}
+
+	updated := controllerutil.RemoveFinalizer(service, h.GetFinalizer())
```
SeanMooney, line 680: ok so rather then check if there is one you just delete it
i guess that saves an api call

SeanMooney: oh this is working on our local copy right and the actul remove happens as a result fo the Update call below?

controllers/nova_controller.go
```diff
@@ -562,11 +624,121 @@ func (r *NovaReconciler) ensureAPI(
...
+	util.LogForObject(h, "Reconciled delete successfully", instance)
+	return nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *NovaReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&novav1.Nova{}).
 		Owns(&mariadbv1.MariaDBDatabase{}).
+		Owns(&keystonev1.KeystoneService{}).
```
SeanMooney, line 741: i was kind of expecting the nova CR to own the nova user/project
and the api cr to own the service.

you also declared the api as ownign the service so its proably fine that we bundel the user/project into that but it  would have been nice if both were seperate.

controllers/novaconductor_controller.go
```diff
@@ -289,6 +296,9 @@ func (r *NovaConductorReconciler) generateServiceConfigMaps(
 	templateParameters := make(map[string]interface{})
 	templateParameters["ServiceUser"] = instance.Spec.ServiceUser
 	templateParameters["KeystonePublicURL"] = instance.Spec.KeystoneAuthURL
+	// TODO(gibi): This should come from the Spec / Secret but hardcoded
+	// here now
+	templateParameters["PlacementServiceUser"] = "placement"
```
SeanMooney, line 301: this should also be nova but ill fix that in my config gen serises
the server operators shoudl only use there own keystone user/project to interact with others.

templates/novaconductor/config/nova.conf
```diff
@@ -12,3 +12,21 @@ username = {{ .ServiceUser }}
...
+project_name = service
+user_domain_name = Default
+username = {{ .PlacementServiceUser }}
+auth_url = {{ .KeystonePublicURL }}
+auth_type = password
+
+[service_user]
+send_service_user_token = True
+auth_url = {{ .KeystonePublicURL }}
+project_name = service
```
SeanMooney, line 28: nit: nova

templates/novaconductor/config/nova.conf
```diff
@@ -12,3 +12,21 @@ username = {{ .ServiceUser }}
 www_authenticate_uri = {{ .KeystonePublicURL }}
 auth_url = {{ .KeystonePublicURL }}
 auth_type = password
+
+[placement]
+region_name = regionOne
```
SeanMooney, line 17: im kind of ignoring the project/user/password since this will all be changing in my config gen patch.

with how your are deploying things currntly that looks ok

test/functional/nova_controller_test.go
```diff
@@ -452,4 +448,41 @@ var _ = Describe("Nova controller", func() {
...
+				condition.KeystoneServiceReadyCondition,
+				corev1.ConditionTrue,
+			)
+
+			service := GetKeystoneService(novaKeystoneServiceName)
+			Expect(service.Finalizers).To(ContainElement("Nova"))
+
+			DeleteNova(novaName)
+			service = GetKeystoneService(novaKeystoneServiceName)
+			Expect(service.Finalizers).NotTo(ContainElement("Nova"))
```
SeanMooney, line 485: ack we proably should be asserting the deleteion of the nova keystone user and project oo but we can adress that in a followup

SeanMooney (approved):

Issue #80, opened 2022-10-04
[NovaAPI]Expose the nova-api service


gibizer:
1. expose it via openshift Route or k8s native ingress
2. expose it in keystone as endpoint

PR #132, opened 2022-11-03
[NovaAPI]Expose it in k8s as a Service and in Keystone as Endpoints

gibizer:
This adds both the Service an Route creation and the KeystoneEndpoint creation for nova-api service. After this the Nova API can be used from outside of the kubernetes cluster.

Note that as KeystoneEndpoint automatically gets NovaAPI's finalizer (to avoid accidental deletion) NovaAPI controller needs to add a finalizer to NovaAPI too to have a chance to remove the finalizer from KeystoneEndpoint during NovaAPI deletion.

Closes: #80 
Depends-on: #131


controllers/novaapi_controller.go
```diff
@@ -350,38 +414,175 @@ func (r *NovaAPIReconciler) ensureDeployment(
...
+func (r *NovaAPIReconciler) ensureServiceExposed(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1.NovaAPI,
+) (ctrl.Result, error) {
+	var ports = map[endpoint.Endpoint]endpoint.Data{
+		endpoint.EndpointAdmin:    {Port: novaapi.APIServicePort},
+		endpoint.EndpointPublic:   {Port: novaapi.APIServicePort},
+		endpoint.EndpointInternal: {Port: novaapi.APIServicePort},
+	}
```
SeanMooney, line 454: this is somthing i want to revisit.

i do no think we should be exposing services on multiple ports.

i think we should take the operturing to run everything on port 80/443 using a subdomain per OpenStack service.
so  nova.public.cloud.com, nova.internal.cloud.com, nova.admin.cloud.com

we can run the container on the default ports but we should use the k8s service to convert it to a URL on the standard ports to avoid needing to punch many holes through the firewall.

this is a wider topic then just nova but i think it would be a mistake not to do this in all operators.

can we bring this up next week?

SeanMooney (commented):
not a review just skimed this quickly

controllers/novaapi_controller.go
gibizer: All three endpoints are on the same port so only a single port is used. Also you get something very similar to 
>  nova.public.cloud.com, nova.internal.cloud.com, nova.admin.cloud.com

by this.
```
(.venv) [fedora@gibi-f35 ~]$ openstack endpoint list | grep nova
| 4c9c561b19454b9696ac4dc6f639139c | regionOne | nova         | compute      | True    | internal  | http://nova-internal-openstack.apps-crc.testing      |
| 5ea76a5c1ba84cc3844ffec4c43f36c1 | regionOne | nova         | compute      | True    | admin     | http://nova-admin-openstack.apps-crc.testing         |
| c224a5a944d24b92820c2f3180089db4 | regionOne | nova         | compute      | True    | public    | http://nova-public-openstack.apps-crc.testing        |
```

SeanMooney: ah they have already dont that in the generic code.

we talked about this a few months ago but i never actually checked if we were just using fqdns yet without needing to list the port.

the port you are passing is the port that is used as the target of the service but the service is exposing the endpoint on 80.

and it will get exposed on 443 when we eventually enable tls.

test/functional/novaapi_controller_test.go
```diff
@@ -359,15 +359,92 @@ var _ = Describe("NovaAPI controller", func() {
...
+
+		It("creates KeystoneEndpoint", func() {
+			SimulateStatefulSetReplicaReady(statefulSetName)
+			SimulateKeystoneEndpointReady(types.NamespacedName{Namespace: namespace, Name: "nova"})
+
+			keystoneEndpoint := GetKeystoneEndpoint(types.NamespacedName{Namespace: namespace, Name: "nova"})
+			endpoints := keystoneEndpoint.Spec.Endpoints
+			Expect(endpoints).To(HaveKeyWithValue("public", "http:"))
+			Expect(endpoints).To(HaveKeyWithValue("internal", "http:"))
+			Expect(endpoints).To(HaveKeyWithValue("admin", "http:"))
```
SeanMooney, line 388: ack for now that makes sense but obviouly this will change when we support tls

SeanMooney (approved):

PR #135, opened 2022-11-08
fix test failures form pr 122 and 120 conflict

SeanMooney:
pr 122 intoduced usage of statefulsets for nova api
pr 120 added multicell support

pr 122 and 120 were not in git merge conflict and both passed
all test indepently. however when pr 120 was automatically
rebased when merging it did not trigger another test run...
so it merged with failing tests.

This fixes that.

### DIFF
--- a/test/functional/nova_multicell_test.go
+++ b/test/functional/nova_multicell_test.go
@@ -201,8 +201,12 @@ var _ = Describe("Nova controller", func() {
 			SimulateMariaDBDatabaseCompleted(cell0.MariaDBDatabaseName)
 			SimulateJobSuccess(cell0.CellDBSyncJobName)
 
-			GetNovaAPI(novaAPIName)
-			novaAPIDeployment := GetDeployment(novaAPIdeploymentName)
+			api := GetNovaAPI(novaAPIName)
+			Expect(api.Spec.Replicas).Should(BeEquivalentTo(1))
+			Expect(api.Spec.Cell0DatabaseHostname).Should(BeEquivalentTo("hostname-for-db-for-api"))
+			Expect(api.Spec.Cell0DatabaseHostname).Should(BeEquivalentTo(api.Spec.APIDatabaseHostname))
+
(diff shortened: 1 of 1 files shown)


test/functional/nova_multicell_test.go
```diff
@@ -201,8 +201,9 @@ var _ = Describe("Nova controller", func() {
 			SimulateMariaDBDatabaseCompleted(cell0.MariaDBDatabaseName)
 			SimulateJobSuccess(cell0.CellDBSyncJobName)
 
-			GetNovaAPI(novaAPIName)
-			novaAPIDeployment := GetDeployment(novaAPIdeploymentName)
+			api := GetNovaAPI(novaAPIName)
+			println(api)
```
SeanMooney, line 205: oh ill fix this

test/functional/nova_multicell_test.go
```diff
@@ -201,8 +201,9 @@ var _ = Describe("Nova controller", func() {
 			SimulateMariaDBDatabaseCompleted(cell0.MariaDBDatabaseName)
 			SimulateJobSuccess(cell0.CellDBSyncJobName)
 
-			GetNovaAPI(novaAPIName)
-			novaAPIDeployment := GetDeployment(novaAPIdeploymentName)
+			api := GetNovaAPI(novaAPIName)
+			Expect(api).ShouldNot(BeNil())
```
gibizer, line 205: api will never be nil. GetNovaAPI either fails the test or returns a *NovaAPI struct that was read from k8s.

gibizer (dismissed):

test/functional/nova_multicell_test.go
SeanMooney: _ := GetNovaApi(novaAPIName)

complains there is no variable to the left so ill just add some addtional assert ot make sure the inputs are correct.
does that work for you.

gibizer (approved):

Issue #134, opened 2022-11-03
Properly default struc fields


gibizer:
After may failed attempt I think @stuggi found the solution.

```golang
type SimpleSpec struct {
        // +kubebuilder:validation:Optional
        // +kubebuilder:default={one:Example1, two:Example2}
        Pair Pair `json:"pair,omitempty"`
        // +kubebuilder:validation:Optional
        // +kubebuilder:default="bar"
        Foo string `json:"foo,omitempty"`
}

type Pair struct {

        // +kubebuilder:validation:Optional
        // +kubebuilder:default=Example1
        One string `json:"one"`

        // +kubebuilder:validation:Optional
        // +kubebuilder:default=Example2
        Two string `json:"two"`
}
```

- `Pair` needs to have a full default defined so that when `pair` is missing from the input this default will be used as the value
- `One` and `Two` needs to have its own default defined at the field level, so that when only one of them is defined in the input the other can use the default value defined at the field level.

PR #141, opened 2022-11-16
Default Nova.cellTemplates to create cell0

gibizer:
1. Default cellTemplates to have cell0 deployed if nothing is defined.
2. Made extra defaulting to further reduce minimum sample needed to define Nova with cell0 or even with multiple cells. 

As a result:
- Minimal config that results in a Nova deployment with cell0: [config/samples/nova_v1beta1_nova_only_cell0.yaml](https://github.com/gibizer/nova-operator/blob/2c5f332a0547460b79710ae1e327e0ba7be5bd72/config/samples/nova_v1beta1_nova_only_cell0.yaml)
- Minimal config that results in a Nova deployment with both cell0 and cell1: [config/samples/nova_v1beta1_nova.yaml](https://github.com/gibizer/nova-operator/blob/2c5f332a0547460b79710ae1e327e0ba7be5bd72/config/samples/nova_v1beta1_nova.yaml)

Partially fixes: #134

### DIFF
--- a/api/v1beta1/nova_types.go
+++ b/api/v1beta1/nova_types.go
@@ -29,12 +29,14 @@ type NovaSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=keystone
 	// KeystoneInstance to name of the KeystoneAPI CR to select the Service
 	// instance used by the Nova services to authenticate.
 	KeystoneInstance string `json:"keystoneInstance"`
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=openstack
 	// APIDatabaseInstance is the name of the MariaDB CR to select the DB
 	// Service instance used for the Nova API DB.
 	APIDatabaseInstance string `json:"apiDatabaseInstance"`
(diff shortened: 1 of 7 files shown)


api/v1beta1/nova_types.go
```diff
@@ -29,23 +29,27 @@ type NovaSpec struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=unused
 	// APIMessageBusInstance is the name of the RabbitMqCluster CR to select
 	// the Message Bus Service instance used by the Nova top level services to
 	// communicate.
 	APIMessageBusInstance string `json:"apiMessageBusInstance"`
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={cell0: {cellDatabaseInstance: openstack, cellDatabaseUser: nova_cell0, cellMessageBusInstance: unused, hasAPIAccess: true, conductorServiceTemplate: {containerImage: "quay.io/tripleowallabycentos9/openstack-nova-conductor:current-tripleo"}}}
```
SeanMooney, line 52: ill need to include this in the default image change pr but that's ok

api/v1beta1/nova_types.go
```diff
@@ -84,18 +89,18 @@ type NovaSpec struct {
 	Debug Debug `json:"debug,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:type=NovaAPITemplate
+	// +kubebuilder:default={containerImage: "quay.io/tripleowallabycentos9/openstack-nova-api:current-tripleo"}
```
SeanMooney, line 92: and again here

it would hav been nicer to just have 

efault={}
and have this core form the template default but this is workable

config/crd/bases/nova.openstack.org_nova.yaml
```diff
@@ -343,9 +343,18 @@ spec:
...
                   - cellDatabaseInstance
+                  - cellDatabaseUser
                   - cellMessageBusInstance
                   - hasAPIAccess
                   type: object
+                default:
+                  cell0:
+                    cellDatabaseInstance: openstack
+                    cellDatabaseUser: nova_cell0
+                    cellMessageBusInstance: unused
```
SeanMooney, line 359: unused works but im tempted to use null here instead.

SeanMooney (approved):
some nits inline but i think im ok with this
lets sync on it tomorrow

PR #145, opened 2022-11-21
[NovaAPI]Fix config template

gibizer:
The metadata and osapi worker config is changed to 1 as that is the minimum by the oslo config definition.

The enforce_new_defaults policy is set to False to keep allowing the global admin to list services. This is done to ease testing.

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -36,10 +36,10 @@ force_config_drive=True
 {{if eq .service_name "nova-api"}}
 # scaling should be done by running more pods
 osapi_compute_workers=1
-metadata_workers=0
+metadata_workers=1
 {{else if eq .service_name "nova-metadata"}}
 # scaling should be done by running more pods
-osapi_compute_workers=0
+osapi_compute_workers=1
 metadata_workers=1
 {{end}}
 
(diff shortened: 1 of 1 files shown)


templates/nova.conf
```diff
@@ -36,10 +36,10 @@ force_config_drive=True
 {{if eq .service_name "nova-api"}}
 # scaling should be done by running more pods
 osapi_compute_workers=1
-metadata_workers=0
+metadata_workers=1
```
SeanMooney, line 39: your right although i was seting them to 0 to disabel then endpoint

i actually should have been setting 

https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.enabled_apis

so nova-api should have 
[DEFAULT]
enabled_apis=['osapi_compute']

and the metadtat confg should have


[DEFAULT]
enabled_apis=['metadata']


when running under apache this is less important since we only load one of the applications directly.

care to also fix that or will i do that in a follwoup

SeanMooney: i have filed an issue for this #156 so lets just fix this now and come back to it.

SeanMooney (approved):

templates/nova.conf
```diff
@@ -60,7 +60,7 @@ heartbeat_in_pthread=false
 
 {{ if eq .service_name "nova-api"}}
 [oslo_policy]
-enforce_new_defaults=true
+enforce_new_defaults=false
```
SeanMooney, line 63: filed #157 to track setting this back to true

SeanMooney (approved):

PR #146, opened 2022-11-21
Configure message bus connection for the Nova control plane

gibizer:
This introduce the creation of the TransportURL CRD to get access to the message bus in the deployment. Then configures this message bus for NovaAPI and NovaConductor CRs. 

Closes: #130


test/functional/nova_multicell_test.go
```diff
@@ -59,7 +60,20 @@ func NewCell(novaName types.NamespacedName, cell string) Cell {
...
+			Namespace: novaName.Namespace,
+			Name:      cell + "-transport",
+		},
 	}
+
+	if cell == "cell0" {
+		c.TransportURLName = types.NamespacedName{
+			Namespace: novaName.Namespace,
+			Name:      "nova-api-transport",
+		}
```
SeanMooney, line 73: hum... we proably should not be hardcodeing this.

that transprot url should be based on the name of the nova-api cr instance

but for the test code i think this is ok

SeanMooney (commented):
i am happy wiht this over all some this bug nothing that woudl prevent this merging

SeanMooney (approved):

PR #162, opened 2022-11-24
Bump ginkgo to v2.5.1 to sync with lib-common

gibizer:


SeanMooney (approved):
sure sound reasonable to me

PR #163, opened 2022-11-24
Logging: Use ISO8601 timestamps

bogdando:
The default value for the zap logger is using epoc timestamps, but that's not all that useful for us, so we are changing it to using the ISO8601 time format instead.

The zap package from uber is already a dependency from the zap module of the controller-runtime package we already use, so we are not adding a new dependency.

Signed-off-by: Bogdan Dobrelya <bogdando@yahoo.com>


SeanMooney:
oh this i logging of the operator output.
yes i would prefer iso timestamps for that too.
we likely want to do this in all operators but I'm supportive of this change.

gibizer (approved):
There was  a discussion in Slack today that we might want to centralize this to lib-common. But sure that would be another change anyhow and until then this is still useful.

gibizer (approved):
looks good

PR #166, opened 2022-11-30
update default namespace in rbac files

SeanMooney:


dprince:
With OLM these should automatically get swapped out for 'openstack' I think. Were you hitting an issue where you needed to change these locally for development?

SeanMooney:
yes in the latest CRC, and perhaps in older versions, i have not checked there is no system namespace.

so the default we had would not actually work if you did a oc apply -k config/rbac

which we need to do to be able to run the operator without installing it for debugging.

i have been changing these locally as a workaround but i would like it to work out of the box.

gibizer:
I cannot reproduce the problem with
```
$ head  ~/.crc/crc.log
time="2022-11-30T13:37:14+01:00" level=debug msg="CRC version: 2.10.1+426703d\n"
time="2022-11-30T13:37:14+01:00" level=debug msg="OpenShift version: 4.11.7\n"
time="2022-11-30T13:37:14+01:00" level=debug msg="Podman version: 4.2.0\n"
```

After crc is up, I did `make crc_storage` and `make namespace` from install_yamls then I run `make install` and `make run` from nova-operator.

SeanMooney:
make install will install the nova operator with olm
which will create the service account

i explicitly don't  install all the operators with make install to avoid any sidefect of installing nova with olm

so I install and deploy rabit/mariadb  then keystone and placement then do make install form the nova-operator repo to create the crd and add the rbac rules.

gibizer:
> make install will install the nova operator with olm which will create the service account
> 
> i explicitly don't install all the operators with make install to avoid any sidefect of installing nova with olm
> 
> so I install and deploy rabit/mariadb then keystone and placement then do make install form the nova-operator repo to create the crd and add the rbac rules.

In general I'm not sure we want to support two different ways to install nova-operator. That would also mean we need to support two different ways to upgrade nova-operator to later versions.

Specifically about this change I have no problem having a meaningful default in the repo for the namespace.

gibizer:
OK with the help of @SeanMooney I was able to reproduce the issue. If I only use `make install` and `make run` from nova-operator then nobody installs nova's ServiceAccount. Interestingly nova-operator only fails when it tries to create the MariaDBDatabase instance, no error earlier. Then applying the config/rbac fails with the missing system namespace:
```
gibi@bedrock:~/upstream/git/openstack-k8s-operators/nova-operator$ oc kustomize config/rbac/ | oc apply -f -
serviceaccount/nova created
role.rbac.authorization.k8s.io/nova-role created
role.rbac.authorization.k8s.io/leader-election-role created
clusterrole.rbac.authorization.k8s.io/manager-role created
clusterrole.rbac.authorization.k8s.io/metrics-reader created
clusterrole.rbac.authorization.k8s.io/proxy-role created
rolebinding.rbac.authorization.k8s.io/leader-election-rolebinding created
rolebinding.rbac.authorization.k8s.io/nova-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/manager-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/proxy-rolebinding created
Error from server (NotFound): error when creating "STDIN": namespaces "system" not found
Error from server (NotFound): error when creating "STDIN": namespaces "system" not found
```

So I think switching from system to openstack in the rbac defs make sense to enable debugging nova-operator without the need to manually change the rbac rule each time.

mrKisaoLamb:
Do we plan to support this case where the system namespace has not been created, or should we drop this pull request ? @SeanMooney  @gibizer

gibizer:
> Do we plan to support this case where the system namespace has not been created, or should we drop this pull request ? @SeanMooney @gibizer

As per:
> So I think switching from system to openstack in the rbac defs make sense to enable debugging nova-operator without the need to manually change the rbac rule each time.

I'm fine switching it.

config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
...
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: controller
+  newTag: latest
```
gibizer, line 16: This change is generated by doing local build. I'm not sure we want to check this in.

config/manifests/bases/nova-operator.clusterserviceversion.yaml
```diff
@@ -26,6 +26,12 @@ spec:
       kind: NovaConductor
       name: novaconductors.nova.openstack.org
       version: v1beta1
+    - description: NovaExternalCompute is the Schema for the novaexternalcomputes
+        API
+      displayName: Nova External Compute
+      kind: NovaExternalCompute
+      name: novaexternalcomputes.nova.openstack.org
+      version: v1beta1
```
gibizer, line 34: Yepp this was somehow missed in the original external compute commit. I agree that we need to have this here.

config/rbac/service_account.yaml
```diff
@@ -2,13 +2,18 @@ apiVersion: v1
...
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: nova
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name:  nova-operator-nova
+---
```
gibizer, line 16: Do we need both SeviceAccounts?

gibizer: We agreed to keep both for now.

gibizer (commented):
@mrKisaoLamb I only have one thing left inline, but then it is good to goo.

mrKisaoLamb:
> @mrKisaoLamb I only have one thing left inline, but then it is good to go.

@gibizer Thanks, done

PR #168, opened 2022-11-30
Integrate into openstack-operator

gibizer:
So far I only needed to add `Nova.IsReady()` so openstack-operator can use it to propagate nova's status up.
This is needed for https://github.com/openstack-k8s-operators/openstack-operator/pull/89


api/v1beta1/nova_types.go
```diff
@@ -154,3 +155,9 @@ func init() {
 func (s NovaStatus) GetConditions() condition.Conditions {
 	return s.Conditions
 }
+
+// IsReady returns true if Nova reconciled successfully
+func (c Nova) IsReady() bool {
+	readyCond := c.Status.Conditions.Get(condition.ReadyCondition)
+	return readyCond != nil && readyCond.Status == corev1.ConditionTrue
+}
```
SeanMooney, line 163: the operator does not have a golang publicinterface and this is not used in the nova operator so the fact you are adding this say that the meta operator is using our internal interface functions.

the API boundary between the meta operator and nova-operator is the CRDs it should only be interactig with us indirectly via the kubernets datamodel and should not be using any functions in our API module.

gibizer: Everything that is in /api are part of our public interface. The openstack-operator imports the golang types defining our CRD to be able to:
1. defines its own CRD including Nova
2. create an instance of the Nova CRD in runtime

Btw, nova-operator also imports the /api module from both mariadb-operator, keystone-operator to be able to create CRs defined in those modules.

The IsReady() function only operates on our public CRD type defs in /api so I don't think this breaches any public / private boundary.

SeanMooney: ya im aware i didnt like doing that for that either i guess we can keep this pattern for now. i was hoping to put the API boundary at the yaml crd deviciotn not the golang type but i guess this is ok.

SeanMooney (approved):
i think we are ok to proceed with this as is

SeanMooney (approved):

PR #174, opened 2022-12-04
add_service_db_finalizer_logic

pinikomarov:
Ever since  : https://github.com/openstack-k8s-operators/lib-common/pull/87

Podified operators which use service dbs and the db.CreateOrPatchDB method ,
now will get the finalizer set on the db automatically when using the newest lib-common,
but they won't have the code remove it on cleanup i.e reconcileDelete,
which might make the delete operator action hang.

This adds the service Db removal code in the operator delete (reconcileDelete)
function to work with the above.

Also updated lib-common to latest as to not break lib-common's exported functio

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -37,6 +37,7 @@ import (
 
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/nova-operator/pkg/nova"
+	"github.com/openstack-k8s-operators/nova-operator/pkg/novaapi"
 
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
@@ -854,6 +855,36 @@ func (r *NovaReconciler) ensureKeystoneServiceUser(
 	return nil
 }
 
+func (r *NovaReconciler) ensureDBDeletion(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1.Nova,
+) error {
+	// remove finalizers from all of our MariaDBDatabases to ensure they
+	// are deleted
+
+	// initialize a nova dbs list with default db names and add used cells:
+	novaDbs := []string{"nova-api"}
+	for cellName := range instance.Spec.CellTemplates {
+		novaDbs = append(novaDbs, novaapi.ServiceName+"-"+cellName)
+	}
+	// iterate over novaDbs and remove finalizers
+	for _, dbName := range novaDbs {
+		db, err := database.GetDatabaseByName(ctx, h, dbName)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return err
+		}
+		if !k8s_errors.IsNotFound(err) {
+			if err := db.DeleteFinalizer(ctx, h); err != nil {
+				return err
+			}
+		}
+	}
+
+	util.LogForObject(h, "Removed finalizer from MariaDBDatabase CRs", instance, "MariaDBDatabase names", novaDbs)
+	return nil
+}
+
 func (r *NovaReconciler) ensureKeystoneServiceUserDeletion(
 	ctx context.Context,
 	h *helper.Helper,
@@ -912,7 +943,12 @@ func (r *NovaReconciler) reconcileDelete(
 ) error {
 	util.LogForObject(h, "Reconciling delete", instance)
 
-	err := r.ensureKeystoneServiceUserDeletion(ctx, h, instance)
+	err := r.ensureDBDeletion(ctx, h, instance)
+	if err != nil {
+		return err
+	}
+
+	err = r.ensureKeystoneServiceUserDeletion(ctx, h, instance)
 	if err != nil {
 		return err
 	}
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -605,5 +605,21 @@ var _ = Describe("Nova controller", func() {
 			service = GetKeystoneService(novaKeystoneServiceName)
 			Expect(service.Finalizers).NotTo(ContainElement("Nova"))
 		})
+
+		It("removes the finalizers from the nova dbs", func() {
+			SimulateKeystoneServiceReady(novaKeystoneServiceName)
+
+			apiDB := GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
+			Expect(apiDB.Finalizers).To(ContainElement("Nova"))
+			cell0DB := GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
+			Expect(cell0DB.Finalizers).To(ContainElement("Nova"))
+
+			DeleteNova(novaName)
+
+			apiDB = GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
+			Expect(apiDB.Finalizers).NotTo(ContainElement("Nova"))
+			cell0DB = GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
+			Expect(cell0DB.Finalizers).NotTo(ContainElement("Nova"))
+		})
 	})
 })
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/openshift/api v3.9.0+incompatible
 	github.com/openstack-k8s-operators/keystone-operator/api v0.0.0-20221215165910-80274d6445b1
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20230110110231-28368fe6a6a9
-	github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230110110231-28368fe6a6a9
+	github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230126174350-3be54a6d3263
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.0.0-20230110110231-28368fe6a6a9
 	github.com/openstack-k8s-operators/mariadb-operator/api v0.0.0-20221128124656-71e59ad7384d
 	github.com/openstack-k8s-operators/nova-operator/api v0.0.0-20221209164002-f9e6b9363961
--- a/go.sum
+++ b/go.sum
@@ -312,8 +312,8 @@ github.com/openstack-k8s-operators/keystone-operator/api v0.0.0-20221215165910-8
 github.com/openstack-k8s-operators/keystone-operator/api v0.0.0-20221215165910-80274d6445b1/go.mod h1:HfMoeyB6kEPfH6Q5YlLyUwmY5ZIgEsDoIBvArE01Oto=
 github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20230110110231-28368fe6a6a9 h1:Z8+r/5O5AfTOzbdrUxSbGxdiSbhluFwDMU/5c9jMkFA=
 github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20230110110231-28368fe6a6a9/go.mod h1:qV9OlokZRpqbHI3lmeN5EOmIKynWphw6GPl3zP9KOGM=
-github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230110110231-28368fe6a6a9 h1:ho8NL1fPXgJEN0/xcykA6SDVme53S+vI+3BD6KWyT+o=
-github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230110110231-28368fe6a6a9/go.mod h1:rONM/XgvFs6putDIxRHNv9/CTGh2afAvJM5wRP2OywY=
+github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230126174350-3be54a6d3263 h1:upigetLqHvFWeTf+4Tbs9lCjgtpIxsc+ObAQw7OGOmg=
+github.com/openstack-k8s-operators/lib-common/modules/database v0.0.0-20230126174350-3be54a6d3263/go.mod h1:rONM/XgvFs6putDIxRHNv9/CTGh2afAvJM5wRP2OywY=
 github.com/openstack-k8s-operators/lib-common/modules/openstack v0.0.0-20220915080953-f73a201a1da6 h1:MVNEHyqD0ZdO9jiyUSKw5M2T9Lc4l4Wx1pdC2/BSJ5Y=
 github.com/openstack-k8s-operators/lib-common/modules/openstack v0.0.0-20220915080953-f73a201a1da6/go.mod h1:YsqouRH8DoZAjFaxcIErspk59BcwXtVjPxK/yV17Wrc=
 github.com/openstack-k8s-operators/lib-common/modules/test v0.0.0-20230110110231-28368fe6a6a9 h1:kC5Lf5d8U03yyfHfObgR5hBy+gCUjXiH0teU5gDjKoY=


controllers/nova_controller.go
```diff
@@ -829,6 +829,24 @@ func (r *NovaReconciler) reconcileDelete(
 ) error {
 	util.LogForObject(h, "Reconciling delete", instance)
 
+	// initialize a nova dbs list with default db names and add used cells:
+	nova_dbs := []string{"nova-api", "nova_cell0"}
```
pinikomarov, line 833: @gibizer What do you think about this approach ? 
It gets the hardcoded nova dbs and adds any used ones , then loops over them and removes finalizers

gibizer: You don't have to hardcode nova_cell0 that will come from instance.Spec.CellTemplates below. Other then that I think this is much better now. Thanks.

pinikomarov: ok I'll update the list

pinikomarov:
now working on the  It("removes the finalizer from KeystoneService", func() {  test coverage for the different dbs

controllers/nova_controller.go
gibizer: thanks. looks good now.

test/functional/nova_controller_test.go
```diff
@@ -541,5 +546,23 @@ var _ = Describe("Nova controller", func() {
...
 			Expect(service.Finalizers).NotTo(ContainElement("Nova"))
 		})
+
+		It("removes the finalizers from the nova cell dbs", func() {
+			SimulateNovaServiceReady(novaServiceName)
+			ExpectCondition(
+				novaName,
+				conditionGetterFunc(NovaConditionGetter),
+				condition.KeystoneServiceReadyCondition,
+				corev1.ConditionTrue, //not sure about the keystoneservice condition usage
```
pinikomarov, line 556: @gibizer Not sure about what condition to use here , should I add a new NovaServiceReadyCondition condition as https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/condition/conditions.go#L52

test/functional/nova_controller_test.go
```diff
@@ -541,5 +546,23 @@ var _ = Describe("Nova controller", func() {
...
+				// should I add a new NovaServiceReadyCondition condition as https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/condition/conditions.go#L52
+			)
+
+			service := GetNovaService(novaServiceName)
+			Expect(service.Finalizers).To(ContainElement("Nova"))
+
+			DeleteNova(novaName)
+			service = GetNovaService(novaServiceName)
+			Expect(service.Finalizers).NotTo(ContainElement("Nova")) // I think this is correct as it will check for : nova-operator/config/samples/nova_v1beta1_nova.yaml:kind: Nova
```
pinikomarov, line 565: @gibizer 			`Expect(service.Finalizers).NotTo(ContainElement("Nova")) `
 I think this is correct ("Nova") as it will check for : nova-operator/config/samples/nova_v1beta1_nova.yaml:kind: Nova
But not entirely sure

pinikomarov:
@gibizer can you take a look at the test code I added ? I also left some specific questions I wasn't sure about ^^
tnx

test/functional/nova_controller_test.go
pinikomarov: although right now , go test / make test complains about : 
`nova.nova.openstack.org "nova-service" not found`

controllers/nova_controller.go
```diff
@@ -829,6 +829,24 @@ func (r *NovaReconciler) reconcileDelete(
...
+	nova_dbs := []string{"nova-api"}
+	for cellName := range instance.Spec.CellTemplates {
+		nova_dbs = append(nova_dbs, cellName)
+	}
+	// iterate over nova_dbs and remove finalizers
+	for _, db_name := range nova_dbs {
+		db, err := database.GetDatabaseByName(ctx, h, db_name)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return err
+		}
```
gibizer, line 842: It is debatable if we want to exit early on such error or want to continue removing finalizers from the rest of the DBs and only error out at the end of the loop if any finalizer removal failed. Let's keep it as is for now. We can get back to it later

test/functional/nova_controller_test.go
gibizer: As lib-common adds the finalizer to the MariaDBDatabase CR when it is created by Nova we need to trigger the creation of those CRs in the test. The nova-controller creates the API DB at https://github.com/openstack-k8s-operators/nova-operator/blob/72619e749746e66f54d368615f34fa1bc0e5e2ae/controllers/nova_controller.go#L169 and creates every cell DB in the loop at https://github.com/openstack-k8s-operators/nova-operator/blob/72619e749746e66f54d368615f34fa1bc0e5e2ae/controllers/nova_controller.go#L210 . To be able to reach the point of the api DB creation during the reconcile the Keystone service needs to be ready first https://github.com/openstack-k8s-operators/nova-operator/blob/72619e749746e66f54d368615f34fa1bc0e5e2ae/controllers/nova_controller.go#L151-L153 . So you need `SimulateKeystoneServiceReady(novaKeystoneServiceName)`. 

According to the comment https://github.com/openstack-k8s-operators/nova-operator/blob/72619e749746e66f54d368615f34fa1bc0e5e2ae/controllers/nova_controller.go#L162-L168 the nova-controller does not wait for the DBs to be actually ready before moves forward. So as soon as the keystone service is ready both the api DB and the cell DBs creation is kicked off.  Then you can check if the Finalizers is on the MariaDBDatabase CRs, then you can delete the Nova CR and check again that the finalizers are removed from the MariaDBDatabase CRs

```golang
		It("removes the finalizers from the nova dbs", func() {
			SimulateKeystoneServiceReady(novaKeystoneServiceName)

			api_db := GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
			Expect(api_db.Finalizers).To(ContainElement("Nova"))
			cell0_db := GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
			Expect(cell0_db.Finalizers).To(ContainElement("Nova"))

			DeleteNova(novaName)

			api_db = GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
			Expect(api_db.Finalizers).NotTo(ContainElement("Nova"))
			cell0_db = GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
			Expect(cell0_db.Finalizers).NotTo(ContainElement("Nova"))
		})
```

If you want to wait for a condition before you check the finalizers and delete Nova then you can do the following

```golang
		It("removes the finalizers from the nova dbs", func() {
			SimulateKeystoneServiceReady(novaKeystoneServiceName)
			SimulateMariaDBDatabaseCompleted(mariaDBDatabaseNameForAPI)
			ExpectCondition(
				novaName,
				conditionGetterFunc(NovaConditionGetter),
				novav1.NovaAPIDBReadyCondition,
				corev1.ConditionTrue,
			)
			SimulateMariaDBDatabaseCompleted(mariaDBDatabaseNameForCell0)
			ExpectCondition(
				novaName,
				conditionGetterFunc(NovaConditionGetter),
				novav1.NovaAllCellsDBReadyCondition,
				corev1.ConditionTrue,
			)

			api_db := GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
			Expect(api_db.Finalizers).To(ContainElement("Nova"))
			cell0_db := GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
			Expect(cell0_db.Finalizers).To(ContainElement("Nova"))

			DeleteNova(novaName)

			api_db = GetMariaDBDatabase(mariaDBDatabaseNameForAPI)
			Expect(api_db.Finalizers).NotTo(ContainElement("Nova"))
			cell0_db = GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
			Expect(cell0_db.Finalizers).NotTo(ContainElement("Nova"))
		})
```
By marking the MariaDBDatabse CRs ready with `SimulateMariaDBDatabaseCompleted` (i.e. simulating in the test what the mariadb operator does in a deployment) you can trigger a Condition change in Nova and therefore you can wait for that.

test/functional/base_test.go
```diff
@@ -645,6 +645,23 @@ func GetKeystoneAPI(name types.NamespacedName) *keystonev1.KeystoneAPI {
...
+
+func SimulateNovaServiceReady(name types.NamespacedName) {
+	Eventually(func(g Gomega) {
+		service := GetNovaService(name)
+		service.Status.Conditions.MarkTrue(condition.ReadyCondition, "Ready")
+		g.Expect(k8sClient.Status().Update(ctx, service)).To(Succeed())
+	}, timeout, interval).Should(Succeed())
+	logger.Info("Simulated NovaService ready", "on", name)
+}
+
```
gibizer, line 664: You don't need this. We the Nova CR is handled by the system under test. We only simulate things that are not part of the test environment. In this case we need to simulate that Keystone service is ready to allow the nova-controller to move forward with the reconciliation of the Nova CR. And optionally we you can simulate that mariadb-operator made the MariaDBDatabse CR's Ready. (See my comment below for more specifics)

test/functional/nova_controller_test.go
gibizer: "Nova" is the correct finalizer name. lib-common uses the name of the Kind (type) of the object creating the MariaDBDatabase as the name of the finalizer on the MariaDBDatabase CR. 

https://github.com/openstack-k8s-operators/lib-common/blob/4ea36851a5fb0a7f9de4daa24242282e13823bd4/modules/common/helper/helper.go#L120-L123

https://github.com/openstack-k8s-operators/lib-common/blob/4ea36851a5fb0a7f9de4daa24242282e13823bd4/modules/common/helper/helper.go#L71

gibizer (changes requested):

pinikomarov:
Thanks for the detailed comments @gibizer I'll go over them

pinikomarov:
Hi @gibizer I went with the simpler test option you suggested , 
A question about the mechanism : 
I understand that the SimulateKeystoneServiceReady allows the progression of the reconcile loop and the subsequent creation of the apidbs and celldbs.
But I see that if the : 
condition is true her e, then the reconcile loop is exited (because of the return.Result nil) :

[	if !instance.Status.Conditions.IsTrue(condition.KeystoneServiceReadyCondition) {
		return ctrl.Result{}, nil
	}](https://github.com/pinikomarov/nova-operator/blob/add_service_db_finalizer_logic/controllers/nova_controller.go#L182-L183)

How then is the  SimulateKeystoneServiceReady progresses the reconcile loop and how the apidbs are created here ? 
tnx

gibizer:
> Hi @gibizer I went with the simpler test option you suggested , A question about the mechanism : I understand that the SimulateKeystoneServiceReady allows the progression of the reconcile loop and the subsequent creation of the apidbs and celldbs. But I see that if the : condition is true her e, then the reconcile loop is exited (because of the return.Result nil) :
> 
> [ if !instance.Status.Conditions.IsTrue(condition.KeystoneServiceReadyCondition) { return ctrl.Result{}, nil }](https://github.com/pinikomarov/nova-operator/blob/add_service_db_finalizer_logic/controllers/nova_controller.go#L182-L183)
> 
> How then is the SimulateKeystoneServiceReady progresses the reconcile loop and how the apidbs are created here ? tnx

That condition is True if the KeystoneServiceReadyCondition is False as there is a negation in front. See the **!** in front.

gibizer (changes requested):
I think the functional failure is also valid.

What happens is the following:

`GetDatabaseByName` tries to fetch the MariaDBDatabase instance based on the db_name we pass in. In the current code `nova-api` and `cell0` is passed in during the loop as `databaseName`. 

But the operator creates the MariaDBDatabase instance with `NewDatabaseWithNamespace`:

https://github.com/openstack-k8s-operators/nova-operator/blob/b0e3f90e8b15c4009c873181a847c6af7e48721a/controllers/nova_controller.go#L587-L596

https://github.com/openstack-k8s-operators/nova-operator/blob/b0e3f90e8b15c4009c873181a847c6af7e48721a/controllers/nova_controller.go#L558-L568

https://github.com/openstack-k8s-operators/nova-operator/blob/b0e3f90e8b15c4009c873181a847c6af7e48721a/pkg/nova/common.go#L33

https://github.com/openstack-k8s-operators/lib-common/blob/28368fe6a6a9d29e93082a7ddb05df5e89a0ca5d/modules/database/funcs.go#L54-L70

It sets the `databaseName` to `nova_cell0` and `nova_api` respectively. Also it sets the `name` of the MariaDBDatabase CR to `nova-cell0` and `nova-api`.

Also the `GetDatabaseByName` only sets the `databaseName` on the instance it uses to query the MariaDBDatabase CR:
https://github.com/openstack-k8s-operators/lib-common/blob/28368fe6a6a9d29e93082a7ddb05df5e89a0ca5d/modules/database/funcs.go#L318-L320

So `getDBWithName` will default the `name` of the MariaDBDatabase CR to query to `h.GetBeforeObject().GetName()` 

https://github.com/openstack-k8s-operators/lib-common/blob/28368fe6a6a9d29e93082a7ddb05df5e89a0ca5d/modules/database/funcs.go#L268-L280

That will be `nova`. 

So the finalizer removal code will never find the MariaDBDatabase CR based on these parameters.

We have two MariaDBDatabse CRs during the test:
* `name`: `nova-api` , `databaseName`: `nova_api`
* `name`: `nova-cell0`, `databaseName`: `nova_cell0`
But the code queries the MariaDBDatabse CRs with the following data:
* **`name`: `nova`**, `databaseName`: `nova-api`
* **`name`: `nova`**, `databaseName`: `cell0`

We can fix this PR to pass in the proper `databaseName`. **BUT** we cannot pass the `name` of the MariaDBDatabase CR to `GetDatabaseByName` (the name parameter there is the `databaseName`).

So I think you have to modify lib-common. Add a new `GetDatabaseByName` variant that takes both the `name` and the `databaseName` from the caller. OR change the current `getDBWithName` implementation not to try to default the `name` of the MariaDBDatabase CR based on the instance stored in the helper (`h.GetBeforeObject`) but query the MariaDBDatabase CR purely based on the `databaseName`.

pinikomarov:
@gibizer I would go with the first option ,for this question is also for the second one,
  I would have to use something like: 
in lib-commo/funcs/getDBWithName:
```go
	err := h.GetClient().Get( //TODO: how to use Get to get also by the databaseName var ?
		ctx,
		namespace,
		databaseName,
		db)
```
will adding the databaseName ,be enough to get the serviceDb ? 
Not Sure what the "Get" client function here is capable off.. ?

pinikomarov:
also the test uses : 
			cell0_db = GetMariaDBDatabase(mariaDBDatabaseNameForCell0)
which is : 
		mariaDBDatabaseNameForCell0 = types.NamespacedName{
			Namespace: namespace,
			Name:      "nova-cell0",
		}
this is apposed to the nova/common/ which is with an underscore :
	NovaCell0DatabaseName = "nova_cell0"

Is it intentional ?

pinikomarov:
also I see that the "name" is what failing this , since thi```
s passes : 
add used cells:
	novaDbs := []string{"nova-api", "nova-cell0"}
	for cellName := range instance.Spec.CellTemplates
```

**Comment by pinikomarov** [comment]
I believe I have e a handle on the failing now , my local tests are passing fix on nova_controller side adding : 

1)

```go
		novaDbs = append(novaDbs, novaapi.ServiceName+"-"+cellName)
```
so that the correct db name is "nova-cellx" as the correct naming scheme used in the operator.

2)

 on lib common/database/funcs.go
fix for the database name pass from the GetDatabaseByName: 

https://github.com/openstack-k8s-operators/lib-common/pull/146

*So the lib-common fix should be merged before this one , then I'll update this go.mod to reference that (2)

gibizer (changes requested):
Now I only have some refactoring suggestions here but otherwise this looks good and I verified in CRC that it works as expected.

pinikomarov:
updated to he latest lib-common../database@3be54a6 version with my patch, and squashed the commits .
@gibizer you said you have some refactoring suggestions ?

gibizer:
> updated to the latest lib-common../database@3be54a6 version with my patch, and squashed the commits . @gibizer you said you have some refactoring suggestions ?

Hm, I remember typing up my suggestion for refactoring as a comment but it seems I forgot to actually publish that comment. Anyhow instead of typing up the suggestion I just did the refactoring on top of your commit. (Also fixed the go-lint findings). I pushed up these commits here. Feel free to squash them if you agree with them.

pinikomarov:
Sure , the refactoring looks good

gibizer (approved):
Looks good! Thanks @pinikomarov for pushing this forward.

PR #175, opened 2022-12-05
Change make gotest to work with envtest

gibizer:
The current CI env expects to run our tests via the `make gotest` target so this patch makes sure that such target can be used to run our envtest based functional test.


SeanMooney (approved):
did you want to see this run by prow before merging
testing it locally it does work although we have reverted to searching
for tests in multiple locations not just the test folder.

gibizer:
> did you want to see this run by prow before merging testing it locally it does work 
Yes, I would like to see the CI run on it.

> although we have reverted to searching for tests in multiple locations not just the test folder.
Yes that was intentional. If we start adding unit test then we need to split the make target to two steps one running unit tests and one running our envtest. The complication here is that envtest is go test compatible so a simple `go test` call picks them up automatically but does not setup the envtest environment, i.e. does not download the necessary binaries. So when we add unit tests we have to figure out a way to separate the unit test from the envtest.

SeanMooney:
i think the best way to do the split long term will be to have a make target per test type
and refactort the ci to call those as spereate jobs.

so unit, functional, lint, e2e, ...
that's what i started on in https://github.com/openstack-k8s-operators/nova-operator/pull/109

gibizer:
> i think the best way to do the split long term will be to have a make target per test type and refactort the ci to call those as spereate jobs.
> 
> so unit, functional, lint, e2e, ... that's what i started on in #109

Yeah, that approach is OK to me.

Makefile
```diff
@@ -265,9 +269,12 @@ govet: get-ci-tools
 	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/govet.sh ./api
 
 # Run go test against code
-gotest: get-ci-tools
-	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh
-	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh ./api
+# this will be moved to the job definition after the job passes
+# in CI GOFLAGS is set to -mod=vendor but envtest fails to install
```
SeanMooney, line 273: in CI GOFLAGS is set to -mod=vendor

i see ok that makes sense i tough we generally did not want 
GOFLAGS to be set to anything but ok that makes sense that we need to override the ci config

SeanMooney (approved):
nice find gibi

PR #176, opened 2022-12-05
Add top level NodeSelector and wire into CRs

dprince:
We need the ability to set this at the top level and have it trickle down to each service as a default.

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/116 (merged)
Depends-on: #201 (merged)


gibizer:
@dprince don't worry about the functional prow job. It is non voting yet. as I'm still trying to make the envtest work in the CI container: #175. The Same test still runs as github action so we are covered.

api/v1beta1/nova_types.go
```diff
@@ -83,6 +83,10 @@ type NovaSpec struct {
 	// passwords from the Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// NodeSelector to target subset of worker nodes running this service
```
gibizer, line 87: We discussed on slack that we need to document the behavior of this top level field compared to the ones in the service templates. Right now I don't know about a better place to document it than here as this is our API documentation right now.

api/v1beta1/novacell_types.go
```diff
@@ -87,6 +87,10 @@ type NovaCellSpec struct {
 	// Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// NodeSelector to target subset of worker nodes running this service
```
gibizer, line 91: Here we need documentation too.

controllers/nova_controller.go
```diff
@@ -698,7 +699,7 @@ func (r *NovaReconciler) ensureAPI(
 
 	op, err := controllerutil.CreateOrPatch(ctx, r.Client, api, func() error {
 		api.Spec = apiSpec
-
+		api.Spec.NodeSelector = instance.Spec.NodeSelector
```
gibizer, line 702: As we discussed on slack we should only overwrite the service specific NodeSelector if it is unset, I.e.: it has a value of map[string]string{}

gibizer: Also you probably want to do propagation of the global NodeSelector from Nova CR to NovaCell CR and from NovaCell to NovaConductor (and eventually we need to do that for the rest of the Nova services when we implement them)

gibizer (changes requested):
I agree with the general direction to provide a global NodeSelector in Nova CR to allow the user to specify the node constraint for all nova services in a single place, while we keep the per service overwriteabilty of it via the service templates.

gibizer:
/test functional

> @dprince don't worry about the functional prow job. It is non voting yet. as I'm still trying to make the envtest work in the CI container: #175. The Same test still runs as github action so we are covered.

Job is fixed now.

controllers/nova_controller.go
dprince: I updated the logic to account for the conditional logic

gibizer:
I've tried the PR and I see the following behavior:
- Both setting and re-setting `Nova.Spec.NodeSelector` works if the label there is actually selecting nodes that can host the pods. E.g. `kubernetes.io/os: linux` `NodeSelector` works in CRC.
- Setting the `Nova.Spec.NodeSelector` to a label that does not match any worker, e.g. `foo: bar` pushes the pods into `Pending` state. This is expected as no nodes match. But then re-setting the `Nova.Spec.NodeSelector` only propagates the empty `NodeSelector` to the `StatefulSet` but then k8s does not kill / re-schedule the `Pending` pod so the system get stuck. I can get the system unstuck by setting `replicas` of the services to `0` first then setting them back to `1`.

As the `StatefulSet` get configured properly in every case I don't think this is a nova-operator problem. More like a non-intuitive (for me at least) behavior from k8s. I did not find confirmation of this with a quick google search. Did anybody see such behavior before? Is it the expected k8s behavior?

gibizer:
With the help of @SeanMooney I see a possible way forward.

There is a [KEP](https://github.com/kubernetes/enhancements/pull/3562) open that describes the problem we see with the Pending Pods. There is a workaround described there to set the `StatefulSet.Spec.PodManagementPolicy` to `Parallel` instead of the default `OrderedReady`. 

To be able to do that we need:
1. a lib-common change to allow setting `StatefulSet.Spec.PodManagementPolicy` https://github.com/openstack-k8s-operators/lib-common/pull/116
2. set the `StatefulSet.Spec.PodManagementPolicy` in https://github.com/openstack-k8s-operators/nova-operator/blob/f9e6b9363961ce1e2595b16308cbcc8c4b999597/pkg/novaapi/deployment.go#L112-L114 and in https://github.com/openstack-k8s-operators/nova-operator/blob/f9e6b9363961ce1e2595b16308cbcc8c4b999597/pkg/novaconductor/deployment.go#L128-L130 to `Parallel`

This will result in the following behavior based on my tests:
- scaling from 1 to 3 replicas happen in parallel, so 2 new pods are created simultaneously. I think this is expected due to the new policy
- setting global NodeSelector to match no nodes will push the *last* Pod in the StatefulSet to Pending. I would expected to have all replicas to go to Pending in parallel due to the new policy. But I guess this is OK as is and actually closer what we want from the StatefulSet
- re-setting the NodeSelector to empty kills the Pending Pod and start a new one and that reaches Running state properly. This is what we want.
- when NodeSelector is set in a way that is selects a node then Pods in the StatefulSet is restared one by one, not in parallel, to pick up the new NodeSelector.

gibizer (changes requested):
We need to set PodManagementPolicy to Parallel.

gibizer:
I needed to push a rebase as well to pick up a fix for a breaking change from master.

gibizer:
The lib-common dependency landed. But we need #201 to merge first to be able to bring in that lib-common dep update.

gibizer:
All dependency merged. I would like to get some review from @SeanMooney before we land this.

SeanMooney:
sorry i missed this ping when i got back form pto

since this was written we have merge the schduler and we are close to having the metadata api

@dprince if you would like one of us can pick this up and add in the missing servcie and do the rebase. 

im not sure the current semantics are correct. i only looked breifly, while propagation is proably ok if you define a node selector on the api template or conductor that need to take precidnece over the one on the nova CR.

i didnt see any test coverage to assert that and im not sure the current code does that so i would like to see that coverd too before we merge this.

we might be able to squeeze this into q1 but we will see.

gibizer:
> since this was written we have merge the schduler and we are close to having the metadata api
> 

I see the logic applied to scheduler and metadata in the recent version of the PR.

> im not sure the current semantics are correct. i only looked breifly, while propagation is proably ok if you define a node selector on the api template or conductor that need to take precidnece over the one on the nova CR.

As far as I see we have the logic to propagate the template value to the spec value by default, but if it results in an empty selector and then we use the Nova CR level value. I think this is correct.

> 
> i didnt see any test coverage to assert that and im not sure the current code does that so i would like to see that coverd too before we merge this.

We have the test coverage in two cases:
* Nova reconfiguration when global NodeSelector is set is propagated to the pods
* Nova reconfiguration when global NodeSelector is set does not override non empty NodeSelector defined in the service template

gibizer:
> > i didnt see any test coverage to assert that and im not sure the current code does that so i would like to see that coverd too before we merge this.
> 
> We have the test coverage in two cases:
> 
> * Nova reconfiguration when global NodeSelector is set is propagated to the pods
> * Nova reconfiguration when global NodeSelector is set does not override non empty NodeSelector defined in the service template

One amendment to this. We use the NovaAPI service CR in the test we probably want to extend the test to show that the same thing works for NovaScheduler, NovaCell, NovaMetadata CRs as well.

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,104 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				g.Expect(serviceDeployment.Spec.Template.Spec.NodeSelector).To(BeNil())
+
+				for _, cell := range []string{"cell0", "cell1", "cell2"} {
+					cellDeploymentName := NewCell(novaName, cell).ConductorStatefulSetName
+					cellDeployment := th.GetStatefulSet(cellDeploymentName)
+					g.Expect(cellDeployment.Spec.Template.Spec.NodeSelector).To(BeNil())
+				}
+			}, timeout, interval).Should(Succeed())
+		},
+		Entry("is propagated to the nova api pods", "-api"),
```
gibizer, line 375: so this would be something like:
```golang
Entry("is propagated to the nova api pods", types.NamespacedName{Namespace: namespace, Name:      novaName.Name + "-api",}),
// [...snip...]
Entry("is propagated to cell0 conductor", NewCell(novaName, "cell0").ConductorStatefulSetName),
// [...snip...]
```

The name handling is being simplified in the refactor @bogdando trying to rebase in #323

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,104 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				nova.Spec.NodeSelector = newSelector
+
+				err := k8sClient.Update(ctx, nova)
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+
+				novaDeploymentName := types.NamespacedName{
+					Namespace: namespace,
+					Name:      novaName.Name + nameSuffix,
+				}
+				apiDeployment := th.GetStatefulSet(novaDeploymentName)
```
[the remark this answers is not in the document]
bogdando: just to note that this sort of naming things is being rebased in https://github.com/openstack-k8s-operators/nova-operator/pull/323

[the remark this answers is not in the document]
mrKisaoLamb: done and ack!

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,104 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				g.Expect(apiDeployment.Spec.Template.Spec.NodeSelector).To(Equal(newSelector))
+
+				for _, cell := range []string{"cell0", "cell1", "cell2"} {
+					cellDeploymentName := NewCell(novaName, cell).ConductorStatefulSetName
+					cellDeployment := th.GetStatefulSet(cellDeploymentName)
+					g.Expect(cellDeployment.Spec.Template.Spec.NodeSelector).To(Equal(newSelector))
+				}
```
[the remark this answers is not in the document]
mrKisaoLamb: done!

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: done, thanks!

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				g.Expect(schedulerDeployment.Spec.Template.Spec.NodeSelector).To(Equal(apiSelector))
+				metadataDeployment := th.GetStatefulSet(novaMetadataDeploymentName)
+				g.Expect(metadataDeployment.Spec.Template.Spec.NodeSelector).To(Equal(apiSelector))
+
+				// But cell conductors get the global one
+				for _, cell := range []string{"cell0", "cell1", "cell2"} {
+					cellDeploymentName := NewCell(novaName, cell).ConductorStatefulSetName
+					cellDeployment := th.GetStatefulSet(cellDeploymentName)
+					g.Expect(cellDeployment.Spec.Template.Spec.NodeSelector).To(Equal(globalSelector))
+				}
```
gibizer, line 445: We don't need to test again that propagation without the service level selector works as that is tested in previous test case. So we can also check instead that the conductor in the cells can have its own NodeSelector that is not overridden by the top level selector.

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
 		})
 	})
 
+	When("global NodeSelector is set", func() {
+		DescribeTable("is propagated to the pods", func(serviceNameFunc func() types.NamespacedName) {
```
gibizer, line 324: to get an OKish test case naming you need to change the string here to: "it is propagated to the"

gibizer (changes requested):
I noticed one implementation flaw. And I have some test suggestions inline.

api/v1beta1/novacell_types.go
```diff
@@ -88,6 +88,11 @@ type NovaCellSpec struct {
 	// Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
 
+	// +kubebuilder:validation:Optional
+	// NodeSelector to target subset of worker nodes running this service. Setting here overrides
+	// any global NodeSelector settings within the Nova CR.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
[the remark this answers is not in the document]
mrKisaoLamb: Done I added NodeSelector to NovaCellTemplate and set it in nova_controller into NovaCellSpec.NodeSelector

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
...
+			novaSchedulerDeploymentName := types.NamespacedName{
+				Namespace: namespace,
+				Name:      novaName.Name + "-scheduler",
+			}
+			novaMetadataDeploymentName := types.NamespacedName{
+				Namespace: namespace,
+				Name:      novaName.Name + "-metadata",
+			}
+
+			// Set NovaAPI specific NodeSelector first
```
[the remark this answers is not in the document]
mrKisaoLamb: done

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
...
+			// Set the global NodeSelector and assert that it is propagated
+			// except to the NovaAPI
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaName)
+
+				nova.Spec.NodeSelector = globalSelector
+				err := k8sClient.Update(ctx, nova)
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+
+				// NovaAPI's deployment keeps it own selector
```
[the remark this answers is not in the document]
mrKisaoLamb: done

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: done

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
...
+
+				err := k8sClient.Update(ctx, nova)
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+
+				serviceDeploymentName := serviceNameFunc()
+				serviceDeployment := th.GetStatefulSet(serviceDeploymentName)
+				g.Expect(serviceDeployment.Spec.Template.Spec.NodeSelector).To(BeNil())
+			}, timeout, interval).Should(Succeed())
+		},
+			Entry("is propagated to the nova api pods",
```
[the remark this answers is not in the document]
mrKisaoLamb: done

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,130 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				nova := GetNova(novaName)
+
+				newSelector := map[string]string{"foo": "bar"}
+				nova.Spec.NodeSelector = newSelector
+
+				err := k8sClient.Update(ctx, nova)
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+
+				novaDeploymentName := serviceNameFunc()
+				apiDeployment := th.GetStatefulSet(novaDeploymentName)
```
gibizer, line 338: sorry I missed this in the previous round. I would rename this to serviceDeploymentName and serviceDeployment As this is not specific to API any more.

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: thanks done!

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: Thankd, done

test/functional/nova_multicell_test.go
```diff
@@ -212,11 +207,11 @@ var _ = Describe("Nova multi cell", func() {
 			// assert that cell related CRs are created pointing to the API MQ
 			cell := GetNovaCell(cell0.CellName)
 			Expect(cell.Spec.CellMessageBusSecretName).To(Equal("mq-for-api-secret"))
-			conductor := GetNovaConductor(cell0.CellConductorName)
+			conductor := GetNovaConductor(cell0.ConductorStatefulSetName)
```
gibizer, line 210: I think this set of changes are wrong. GetNovaConductor will read a novav1.NovaConductor instances from the API so we should pass the name of the NovaConductor CR to it not a the name of the StatefulSet resource. (It happens that the naming scheme uses the same name for both, but from typing perspective we have mismatch here)

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,136 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				}
+				err := k8sClient.Update(ctx, nova)
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+
+				apiDeployment := th.GetStatefulSet(novaAPIdeploymentName)
+				g.Expect(apiDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
+				schedulerDeployment := th.GetStatefulSet(novaSchedulerDeploymentName)
+				g.Expect(schedulerDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
+				metadataDeployment := th.GetStatefulSet(novaMetadataDeploymentName)
+				g.Expect(metadataDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
```
gibizer, line 425: you can assert as ell that GetStatefulSet(cell[0, 1, 2].ConductorStatefulSetName).Spec.Template.Spec.NodeSelector is set to the serviceSelector.

test/functional/nova_reconfiguration_test.go
```diff
@@ -320,4 +320,136 @@ var _ = Describe("Nova reconfiguration", func() {
...
+
+				// NovaService's deployment keeps it own selector
+				apiDeployment := th.GetStatefulSet(novaAPIdeploymentName)
+				g.Expect(apiDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
+				schedulerDeployment := th.GetStatefulSet(novaSchedulerDeploymentName)
+				g.Expect(schedulerDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
+				metadataDeployment := th.GetStatefulSet(novaMetadataDeploymentName)
+				g.Expect(metadataDeployment.Spec.Template.Spec.NodeSelector).To(Equal(serviceSelector))
+
+				// But cell conductors get the global one
```
gibizer, line 446: This needs to be updated as now you assert that the conductors also kept it service level selector instead of the global one.

gibizer (changes requested):

test/functional/nova_multicell_test.go
mrKisaoLamb: Done

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: Done

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: Done

test/functional/nova_reconfiguration_test.go
```diff
@@ -244,15 +244,15 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				cell1Names.ConductorStatefulSetName,
 				ConditionGetterFunc(NovaConductorConditionGetter),
 				condition.NetworkAttachmentsReadyCondition,
 				corev1.ConditionFalse,
 				condition.RequestedReason,
 				"NetworkAttachment resources missing: internalapi",
 			)
 			th.ExpectConditionWithDetails(
-				cell1Names.CellConductorName,
+				cell1Names.ConductorStatefulSetName,
```
gibizer, line 255: Hups this still needs to be reverted.

gibizer (changes requested):

test/functional/nova_reconfiguration_test.go
mrKisaoLamb: Done. Sorry I missed this

gibizer (approved):
Thanks

PR #179, opened 2022-12-08
Add basic validation to our samples

gibizer:
With envtest we can validate that our CRD samples are well formed and has all the required parameters.

This patch also took the opportunity to clean up the samples.


SeanMooney (approved):
this is an ok starting point but right now its pretty minimal coverage.
that said i think kuttl can be used instead of or in addition to this in the long term so we might as well start with this

PR #180, opened 2022-12-08
Refactor our defer calls

gibizer:
Our defer call at the end of Reconcile does two things:
1. update the Ready condition of the instance if all the other
   conditions are in True state
2. persist the metadata and the status changes of the object into etcd.

The 2 is complicated but common for each of our controllers so it is pulled out and the impl moved to ReconcileBase. Note that this automatically adds metadata patching capabilities to those controllers (e.g. cell) that does not need it today.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -274,3 +274,39 @@ func (r *ReconcilerBase) GenerateConfigs(
 	// to enable unit testing at some point.
 	return configmap.EnsureConfigMaps(ctx, h, instance, cms, envVars)
 }
+
+func (r *ReconcilerBase) patchInstance(ctx context.Context, h *helper.Helper, instance client.Object) error {
+	var err error
+
+	if err = h.SetAfter(instance); err != nil {
+		util.LogErrorForObject(h, err, "Set after and calc patch/diff", instance)
+		return err
+	}
+
+	changes := h.GetChanges()
+	patch := client.MergeFrom(h.GetBeforeObject())
+
+	if changes["metadata"] {
+		err = r.Client.Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			util.LogForObject(h, "Metadata update conflict", instance)
+			return err
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			util.LogErrorForObject(h, err, "Metadate update failed", instance)
+			return err
+		}
+	}
+
+	if changes["status"] {
+		err = r.Client.Status().Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			util.LogForObject(h, "Status update conflict", instance)
+			return err
+
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			util.LogErrorForObject(h, err, "Status update failed", instance)
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


controllers/common.go
```diff
@@ -274,3 +274,39 @@ func (r *ReconcilerBase) GenerateConfigs(
...
+			util.LogForObject(h, "Status update conflict", instance)
+			return err
+
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			util.LogErrorForObject(h, err, "Status update failed", instance)
+			return err
+		}
+	}
+	return nil
+}
```
abays, line 312: Could this be a candidate for inclusion in `lib-common`?

SeanMooney: lib-common does not have a common basclass for all the controller which is what enables us to do this in the nova operator without using generics 
so either ReconcilerBase would have to be ported or you would have to use go generics instead of (r *ReconcilerBase)

SeanMooney: you could also defene a ClientWrapper interface i geuss and define a GetClient function and reimplment this in terms of that to remove the r.Client dependency to avoid generics.

abays: I was thinking something like this: https://github.com/openstack-k8s-operators/lib-common/compare/master...abays:helper_instance_update?expand=1

abays: Hrm, that creates an import cycle between the `helper` and `util` packages, but perhaps it could be arranged differently to avoid this

abays: Pushed a new version that works when tested when imported and used with Keystone operator

SeanMooney: something like that would work although the name is important it should be patchInstance not update instance as update and patch have different behaviour regarding how defaults are handled.

i think the depenices should be simple enough to resolve

stuggi: I like having this in lib-common and update all operators to use it. I think that would make it easier to debug things as all operators use the same update mechanism and it won't require to do deep code check how they do the patching.

bogdando: If we are going to invest into this in lib-common, I'd vote to invest into server side apply instead

stuggi: server-side apply is something we want to look into. last time I checked (months ago) it did only work for custom resources if you use unstructured. maybe that changed. in the short term I think it could be still valid to add something what @abays checks to lib-common

gibizer: In general I'm OK to have this moved to lib-common. I can adapt nova to it when the lib-common PR is proposed. I have two comments:
1. I agree with @SeanMooney to have this called PatchInstance in lib-common to be a bit more precise about what happens. I had the unverified feeling that with Patch instead of Update I saw less conflicts from k8s.
2. Please note that this change has actually two parts. One is `patchInstance` but there is also the change in the `Reconciler`s to update the return value of the Reconciler based on the return value of patchInstance.
```
		err := r.patchInstance(ctx, h, instance)
		if err != nil {
			_err = err
			return
		}
```
While this cannot be moved to lib-common, it is worth to note in somewhere, maybe in the doc of PatchInstance so that it gets applied.

SeanMooney: we chatted about this today and while we are happy to have this in lib-common we would like to merge this for now and refactor it later so with that in mind I'm going to resolve this conversation for now.

SeanMooney (approved):

PR #182, opened 2022-12-08
Improve logging

gibizer:
To help troubleshooting this patch adds log when the controller returns form reconcile either due to waiting for some external condition or because it succeeded. All the other return point have a non nil error causing the controller to print a stacktrace. So with this patch every return either logs a message or throws an error.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -180,6 +180,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 
 	// We have to wait until our service is registered to keystone
 	if !instance.Status.Conditions.IsTrue(condition.KeystoneServiceReadyCondition) {
+		util.LogForObject(h, "Waiting for the KeystoneService to become Ready", instance)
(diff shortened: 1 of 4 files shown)


controllers/nova_controller.go
```diff
@@ -417,6 +418,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// until cell0 is ready as top level services need cell0 to register in
 	if cell0, ok := cells[Cell0Name]; !ok || !cell0.IsReady() {
 		// we need to stop here until cell0 is ready
+		util.LogForObject(h, "Waiting for cell0 to become Ready before creating the top level services", instance)
```
bogdando, line 421: Just to clarify, this is `NovaReconciler.Reconcile` and it is a top scope (nothing calls it but kubernetes)? If so, this message is LGTM. When there are callers, the 2nd part of the msg "before...the top scope..." should be instead generated by that caller side instead of the callee. At least, that's how I got the logging wrapping layers in Golang :)

bogdando (approved):

PR #186, opened 2022-12-09
[NovaAPI]Tune probes for slower environment

gibizer:
It seems that in slower dev env our nova-api probes times out during startup. So this patch introduces a startupProbe and tunes the liveness probe to be more gentle. The nova-conductor probes are not changed as there we only assert that the process exists which is not affected by the startup time.

### DIFF
--- a/pkg/novaapi/deployment.go
+++ b/pkg/novaapi/deployment.go
@@ -37,17 +37,22 @@ func StatefulSet(
 ) *appsv1.StatefulSet {
 	runAsUser := int64(0)
 
+	// This allows the pod to start up slowly. The pod will only be killed
+	// if it does not succeed a probe in 60 seconds.
+	startupProbe := &corev1.Probe{
+		FailureThreshold: 6,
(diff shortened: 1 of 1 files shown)


pkg/novaapi/deployment.go
```diff
@@ -37,17 +37,22 @@ func StatefulSet(
...
+	startupProbe := &corev1.Probe{
+		FailureThreshold: 6,
+		PeriodSeconds:    10,
+	}
+	// After the first successful startupProbe, livenesProbe takes over
 	livenessProbe := &corev1.Probe{
 		// TODO might need tuning
-		TimeoutSeconds:      5,
-		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
```
SeanMooney, line 44: i see so orginally we were using InitialDelaySeconds to mitigate the impact of the first start and now we are suing a dedicated startupProbe to do that instead ya that sound oke to me.

SeanMooney (approved):

PR #187, opened 2022-12-09
Bump k8s.io api and apimachinery versions

gibizer:
The individual dependabot PRs are failing so I try to combine them into one to see if that helps.


gibizer:
/lgtm

This is a collapsed version of automatic dependapot patches. So even though I created this I will approve it as well.

PR #189, opened 2022-12-09
[NovaConductor]allow setting logging to debug

gibizer:
The original logging.conf prevented us to see debug log on stdout.


SeanMooney (approved):
yep we will change this again when we swap to the new way of doing config generation but this is fine for now

gibizer:
/test functional

It is a test instability I trying to track down but failed so far

PR #200, opened 2022-12-15
WIP bumping k8s apis to 0.26.0 and controller runtime to 0.14.0

gibizer:
This does not seem to work due to:

CustomResourceDefinition.apiextensions.k8s.io \"novaapis.nova.openstack.org\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[resources].properties[claims].items.x-kubernetes-map-type: Invalid value: \"null\": must be atomic as item of a list with x-kubernetes-list-type=set"

It seems that x-kubernetes-list-type is added to the schema but without a proper value.


gibizer:
0.26.1 looks better now see the renovate bump PR #224

PR #201, opened 2022-12-16
Bumping operator deps

gibizer:
This patch bumps the versions of lib-common and the other operators (keystone, mariadb, openstack) nova depends on to latest.

Note that from openstack-operator we inherit `k8s.io/apimachinery v0.26.0` and that seems to be OK but we cannot bump the rest of k8s.io to 0.26.0 as it breaks controller-runtime and / or controller-gen.


api/go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/openstack-k8s-operators/nova-operator/api
 go 1.18
 
 require (
-	github.com/openstack-k8s-operators/lib-common/modules/common v0.0.0-20220909175216-e774739df18a
-	k8s.io/api v0.25.5
```
gibizer, line 7: we cannot keep this on 0.25.5 as it causes 
```
/home/gibi/upstream/git/openstack-k8s-operators/nova-operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/gibi/go/pkg/mod/golang.org/x/oauth2@v0.0.0-20220909003341-f21342109be1/google/default.go:17:2: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
	cloud.google.com/go v0.97.0 (/home/gibi/go/pkg/mod/cloud.google.com/go@v0.97.0/compute/metadata)
	cloud.google.com/go/compute v1.7.0 (/home/gibi/go/pkg/mod/cloud.google.com/go/compute@v1.7.0/metadata)
Error: not all generators ran successfully
```

SeanMooney (approved):

PR #203, opened 2022-12-18
Make TransportURL name depend on Nova CR name

gibizer:
So far nova-api-transport as TransportURL name was hardcoded. Also the cell TransportURL names only depended on the name of the cell but did not depend on the name of the Nova CR. Now both issue if fixed.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -240,7 +240,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// message bus is always the same as the top level API message bus so
 	// we create API MQ separately first
 	apiMQSecretName, apiMQStatus, apiMQError := r.ensureMQ(
-		ctx, h, instance, "nova-api-transport", instance.Spec.APIMessageBusInstance)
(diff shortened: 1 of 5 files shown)


bogdando:
I'm having really hard times with rebasing this

gibizer:
> I'm having really hard times with rebasing this

try to remove the [[EnvTest]Factor out name handling](https://github.com/openstack-k8s-operators/nova-operator/pull/203/commits/73bdd7c288d17f5b266ab734a448b82b631e16ab) commit and focus only the other for now that has the real fix. We can get back to the name handling in test at a different time.

bogdando:
Resubmitted as WIP https://github.com/openstack-k8s-operators/nova-operator/pull/323

mrKisaoLamb:
@gibizer so, work is continuing in another PR. Should we close this one and continue there https://github.com/openstack-k8s-operators/nova-operator/pull/323?

gibizer:
> @gibizer so, work is continuing in another PR. Should we close this one and continue there #323?
Sure we can. In the general we should keep the original PR and work in that to keep the context in a single place. But for this particular case I'm fine to close this as most of the recent context is in #323 anyhow.

PR #204, opened 2022-12-18
Use switch when checking status enum values

gibizer:


controllers/nova_controller.go
```diff
@@ -208,12 +209,14 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	for _, cellName := range orderedCellNames {
 		cellTemplate := instance.Spec.CellTemplates[cellName]
 		cellDB, status, err := r.ensureCellDB(ctx, h, instance, cellName, cellTemplate)
-		if err != nil {
+		switch status {
```
bogdando, line 212: this eats the err != nil case handling

controllers/nova_controller.go
```diff
@@ -281,12 +285,14 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			cellMQ, status, err = r.ensureMQ(
 				ctx, h, instance, cellName+"-transport", cellTemplate.CellMessageBusInstance)
 		}
-		if err != nil {
+		switch status {
```
bogdando, line 288: ditto

bogdando (changes requested):

controllers/nova_controller.go
gibizer: `status == DBFailed` and `err != nil` are equivalent based on https://github.com/openstack-k8s-operators/nova-operator/blob/72619e749746e66f54d368615f34fa1bc0e5e2ae/controllers/nova_controller.go#L497-L499

controllers/nova_controller.go
gibizer: ditto :)

bogdando (approved):

PR #205, opened 2022-12-18
Use PatchInstance from lib-common

gibizer:
Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/117 (merged)


go.mod
```diff
@@ -90,3 +90,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/nova-operator/api => ./api
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/abays/lib-common/modules/common v0.0.0-20221215201557-feb92dd06374
```
bogdando, line 94: This is LGTM, although we should not approve it for merge until this replacement goes away and the dependency patch is merged, correct?

gibizer: Yes, we need to land the lib common change first and then remove this replace line

gibizer: I have a PR in flight to catch such replace lines and prevent merging them #213

gibizer:
Updated to the latest lib-common proposal.

bogdando (commented):
the dependency has been merged

gibizer:
> the dependency has been merged

yepp I need to get back to this bump the lib-common dep and remove the replace line from the go.mod

bogdando (approved):

PR #206, opened 2022-12-19
Use the new lib-common/common/log interface

gibizer:
It is a Draft as only the nova controller is switched over to the new interface. 

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/125

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -26,12 +26,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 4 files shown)


bogdando (changes requested):
So if I get the comments for dependency patch correct, we should rework this to just use log.FromContext(ctx) ?

bogdando:
> So if I get the comments for dependency patch correct, we should rework this to just use log.FromContext(ctx) ?

please disregard... that's exaclty what you are already doing here :)

bogdando (approved):

stuggi:
@bogdando not sure if you wanted to merge this ?

bogdando:
No, I've missed that auto-merge feature. All I wanted is to remove the "requires changes" mark set by me earlier :(

bogdando:
Shall we revert or wait for https://github.com/openstack-k8s-operators/lib-common/pull/125 to become merged. Sorry for unexpected auto-merge

PR #207, opened 2022-12-19
Enable renovate bot to manage dep bumps

gibizer:
We do the following grouping to i) reduce noise ii) keep connected things bumped together. This means each group will be bumped together with a single PR. Groups:

* one group for everything under openstack-k8s-operators
* one group for everything under k8s.io and sigs.k8s.io. Note that we keep these pinned under v1.0.0 as based on the doc[1] that is for kubernetes version >= 1.17

The rest of the dependencies are handled independently by our renovate config so each dep will have its own PR if bump is needed.

It runs weekly automatically but it can be manually triggered via the Dependency Dashboard Issue.

The renovate config doc is at: https://docs.renovatebot.com/configuration-options

[1] https://github.com/kubernetes/client-go/blob/master/README.md#client-go

An example dependency dashboard an be seen in my fork https://github.com/gibizer/nova-operator/issues/1 that uses the same config as proposed here.


SeanMooney (approved):
this seams reasonable so let move forward with it

PR #212, opened 2023-01-05
Make pre-commit.ci's gofmt happy

gibizer:
It seems pre-commit.ci changed to go1.19.4 where gofmt does a bit more
compared to the 1.18 version we use. This patch adapts the code style to
conform to 1.19 as well.


gibizer:
It seems pre-commit.ci runs with go1.19.4 and gofmt does new thing there compared to our go1.18

SeanMooney (approved):

PR #213, opened 2023-01-05
Github action to avoid merging go.mod replace lines

gibizer:


gibizer:
The job https://github.com/openstack-k8s-operators/nova-operator/actions/runs/3849617793/jobs/6558823335 shows that this would prevent merging the logging change patch with the replace line in the first place.

As we reverted that commit in the meantime as soon as I rebase this the job will pass again.

bogdando (approved):

PR #215, opened 2023-01-10
Get rid of deprecated [DEFAULT] use_forwarded_for

kajinamit:
This option was deprecated during Zed cycle[1] because it is basically duplicate of the oslo_middleware functionality.

[1] https://review.opendev.org/c/openstack/nova/+/836253

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -99,7 +99,6 @@ ssl_minimum_version=tlsv1_3
 
 {{if eq .service_name "nova-api" "nova-metadata-api"}}
 [api]
-use_forwarded_for = true
 auth_strategy = keystone
 [oslo_middleware]
 enable_proxy_headers_parsing = True


gibizer:
> This option was deprecated during Zed cycle[1] because it is basically duplicate of the oslo_middleware functionality.
> 
> [1] https://review.opendev.org/c/openstack/nova/+/836253

Thanks. I see `http_proxy_to_wsgi` in the api-paste.ini in the Zed nova-api container so removing the config option does not change any behavior.

gibizer (approved):

SeanMooney:
for what its worth we are not currently generating the api-paste.ini so we may not be using that oslo middleware currently.

i will be changing this shortly anyway but i added use_forwarded_for to ensure we had the correct behaviour until i added support for multiple config files to nova-api

SeanMooney:
> > This option was deprecated during Zed cycle[1] because it is basically duplicate of the oslo_middleware functionality.
> > [1] https://review.opendev.org/c/openstack/nova/+/836253
> 
> Thanks. I see `http_proxy_to_wsgi` in the api-paste.ini in the Zed nova-api container so removing the config option does not change any behavior.

we expcitly do not want to rely on any config files provided in the container or rpms as part of this new deployment approach so while this will work for now its not something we should depend on in the future.

gibizer:
> > > This option was deprecated during Zed cycle[1] because it is basically duplicate of the oslo_middleware functionality.
> > > [1] https://review.opendev.org/c/openstack/nova/+/836253
> > 
> > 
> > Thanks. I see `http_proxy_to_wsgi` in the api-paste.ini in the Zed nova-api container so removing the config option does not change any behavior.
> 
> we expcitly do not want to rely on any config files provided in the container or rpms as part of this new deployment approach so while this will work for now its not something we should depend on in the future.

But the what do you suggest? Undeprecate the config option?

SeanMooney:
no i wanted to keep it in our config until i added that api-paste.ini template in a later pr

i just didn't want to depend on the one in the config file initially 

for now we don't need to do anything.

ill include this as part of reworking the nova-api config gen

kajinamit:
Personally I'd advocate using the default paste.ini file provided by the package because usually we can expect consistency between the default nova.conf and the file maintained in the nova repository. As Sean mentioned we can maintain our own copies but that requires we timely sync the file from the nova repo, which does not usually happen based on my past experience.

PR #216, opened 2023-01-10
Drop deprecated ovsdb_interface

kajinamit:
This option was deprecated in os-vif 2.2.0, because the vsctl driver was deprecated and only the native driver will be supported.

[1] https://review.opendev.org/c/openstack/os-vif/+/744816


gibizer:
@SeanMooney: could you check this please?

SeanMooney (approved):

PR #218, opened 2023-01-10
Remove unused www_authenticate_uri option

kajinamit:
This option is used only by keystone middleware so should be put only in the keystone_authtoken middleware section.


gibizer (approved):
/lgtm 
thanks @kajinamit

PR #219, opened 2023-01-10
Fix wrong option to determine cinder endpoint type

kajinamit:
The [cinder] section does not support the interface option and we should use the catalo_info option instead.

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -259,5 +259,5 @@ username = {{ .nova_keystone_user }}
 password = {{ .nova_keystone_password }}
 cafile = {{ .openstack_cacert }}
 region_name = {{ .openstack_region_name }}
-valid_interfaces = internal
+catalog_info = volumev3:cinderv3:internalURL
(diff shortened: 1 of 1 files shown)


templates/nova.conf
```diff
@@ -259,5 +259,5 @@ username = {{ .nova_keystone_user }}
 password = {{ .nova_keystone_password }}
 cafile = {{ .openstack_cacert }}
 region_name = {{ .openstack_region_name }}
-valid_interfaces = internal
+catalog_info = volumev3:cinderv3:internalURL
```
SeanMooney, line 262: i tried to use this at one point and i thought i caused an error but 
i see catalog_info = volumev3:cinderv3:internalURL used in other project so 
this makes sense.

with that said we do not support using cinder yet.
we wont actually test that integration until after compute nodes are supported in the operator which wont be for a few weeks yet.

SeanMooney (approved):

PR #221, opened 2023-01-10
sync go work deps with go mod

SeanMooney:
this change syncs go work and go mod to make sure we
are running with the deps we expect


gibizer (approved):
/lgtm

PR #222, opened 2023-01-11
Fix the bundle build

gibizer:
The PR #179 break the bundle build by renaming sample files that are listed in the samples/kustomization.yaml. This patch restores the bundle build. A follow up PR will introduce a test build job to prevent such break in the future.


SeanMooney (approved):
ack

we spoke about the followup job to catch this so I'm fine with merging this while you look into the long term solution

PR #226, opened 2023-01-12
bump deps

SeanMooney:
- add workflow that uses make to run tests.
- Update k8s.io


gibizer:
I prefer to fix up the existing jobs instead https://github.com/openstack-k8s-operators/nova-operator/pull/227 
See also the full explanation why the test jobs passed before in: https://github.com/openstack-k8s-operators/nova-operator/pull/224#issuecomment-1381548293

SeanMooney:
can we change the prow jobs to use the make targets too
i would prefer to change the prow functional job to call make test

PR #227, opened 2023-01-13
Generate manifests before test run

gibizer:
The envtest uses our CRD definitions during test so we need to run `make manifests` and `make generate` before the test run to always test with an up to date CRD yaml and golang definition.

The `make test` target generates manifests today but the `make gotest` target doesn't. The latter is used in the prow CI job[1]. So this patch make `gotest` an alias for the `test` target.

Also this patch switches the github action based test run to use `make gotest` instead of indirectly calling `go test`.

[1] https://github.com/openshift/release/blob/fe0b5ddd5c96f053a654883ae33ec05933ad6c5c/ci-operator/config/openstack-k8s-operators/nova-operator/openstack-k8s-operators-nova-operator-master.yaml#L31

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ PROC_CMD = --procs ${PROCS}
 
 .PHONY: test
 test: manifests generate fmt vet envtest ginkgo ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" $(GINKGO) --trace --cover --coverpkg=../../pkg/nova,../../pkg/novaapi,../../pkg/novaconductor,../../controllers,../../api/v1beta1 --coverprofile cover.out --covermode=atomic --randomize-all ${PROC_CMD} ./test/...
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) -v debug --bin-dir $(LOCALBIN) use $(ENVTEST_K8S_VERSION) -p path)" $(GINKGO) --trace --cover --coverpkg=../../pkg/nova,../../pkg/novaapi,../../pkg/novaconductor,../../controllers,../../api/v1beta1 --coverprofile cover.out --covermode=atomic --randomize-all ${PROC_CMD} ./test/...
 
 ##@ Build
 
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yaml
```diff
@@ -22,11 +22,8 @@ jobs:
         run: GOWORK=off ./openstack-k8s-operators-ci/test-runner/govet.sh
       - name: Run golint.sh
         run: export GOWORK=off && ./openstack-k8s-operators-ci/test-runner/golint.sh
-      - name: Setup envtest
-        run: make envtest
-      # TODO(gibi): change openstack-k8s-operators-ci/test-runner/gotest.sh to run the make test target directly
-      - name: Run gotest.sh
-        run: GOWORK=off KUBEBUILDER_ASSETS=$(./bin/setup-envtest use 1.23 -p path) ./openstack-k8s-operators-ci/test-runner/gotest.sh
+      - name: run test
+        run: make gotest
```
bogdando, line 26: this prolly is missing GOWORK=off make gotest?

bogdando (changes requested):

.github/workflows/golangci-lint.yaml
gibizer: I'd like to keep GOWORK on if possible. We will see...

gibizer:
/test functional
the modification on prow side landed for this to work

.github/workflows/golangci-lint.yaml
gibizer: and also this workflow goes away in favor of the prow's functional job

gibizer:
OK I had two tweak the test target but now `make gotest` is an alias of `make test`

Makefile
```diff
@@ -272,8 +272,7 @@ govet: get-ci-tools
 	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/govet.sh ./api
 
 # Run go test against code
-gotest: get-ci-tools envtest
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) -v debug --bin-dir $(LOCALBIN) use $(ENVTEST_K8S_VERSION) -p path)" $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh
+gotest: test
```
SeanMooney, line 275: ok so this basicaly just make it an alias as you mentioned in the PR message

i have tested this locally and both seam to work fine so if the ci is happy then
we can proceed with this

SeanMooney (approved):

PR #229, opened 2023-01-13
[CI]Remove duplications

gibizer:
Our main CI runner is Prow. It already runs[1] on each PR:
* make gotest (envtest)
* golangci-lint
* make golint
* make govet
* make gofmt
* pre-commit run --all-files

So we can remove most of our github workflows as they are duplicated.
[1] https://github.com/openshift/release/blob/master/ci-operator/config/openstack-k8s-operators/nova-operator/openstack-k8s-operators-nova-operator-master.yaml


.pre-commit-config.yaml
```diff
@@ -1,12 +1,11 @@
 ci:
   autofix_prs: false
   skip:
-   - 'make-manifests'
```
bogdando, line 4: why we no longer want to run these locally?

gibizer: the `ci` section only defines what runs by pre-commit.ci all the `hooks` form all the `repos` below are run locally.

gibizer:
> /pj-rehearse intermittent connectivity problems

The pre-commit.ci stuff seems not work with external dependencies. But prow already runs pre-commit so we can disable pre-commit.ci and only keep the pre-commit config for the local run.

.pre-commit-config.yaml
SeanMooney: yep as gibi said.

i have disabled the  pre-commit.ci for this repo now and i will clean up this in a followup pr

SeanMooney (approved):
this looks ok to me we might need to manually merge this and other open prf since pre-commit.ci has already run ill see if there is a way to remove a check but i don't see on in the ui.

gibizer:
> this looks ok to me we might need to manually merge this and other open prf since pre-commit.ci has already run ill see if there is a way to remove a check but i don't see on in the ui.

It is enough to push a new version of the PR to get a clean list of required jobs.

SeanMooney:
/lgtm

thanks that makes things simpler i was looking at the github API to see if i could clear it somehow

bogdando (approved):

PR #234, opened 2023-01-16
Switch to test helper more resources operations

bogdando:
Use database, keystone, transport  basic test ops
from test helper common lib.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/171

Signed-off-by: Bohdan Dobrelia <bdobreli@redhat.com>


gibizer:
I think you need to put the `replace` line into the go.mod file instead of the go.work file to get the CI pass

bogdando:
so this probably depends on https://github.com/openstack-k8s-operators/nova-operator/pull/265 and will need a rebase after that

bogdando:
this should pass now as I made it using the lib-common@c8a2b321 which doesn't bring in the most recent changes

test/functional/novaapi_controller_test.go
```diff
@@ -332,16 +332,17 @@ var _ = Describe("NovaAPI controller", func() {
 				condition.ExposeServiceReadyCondition,
 				corev1.ConditionTrue,
 			)
-			GetService(types.NamespacedName{Namespace: namespace, Name: "nova-public"})
```
bogdando, line 335: ouch, those changes look unexpeced, hence the test failure

gibizer (approved):
Looks good. Thanks @bogdando !

PR #235, opened 2023-01-17
Remove and ignore go.work.sum

gibizer:
That sum file is a constant pain as it is not kept up to date by the standard tools and jobs we have. Also it can be regenerated with go work sync if needed, but IDEs like vscode does it automatically anyhow.


SeanMooney (approved):
ya lets nuke this thanks :)

PR #237, opened 2023-01-20
Test logging improvements

gibizer:
Added -v to ginkgo to produce output even it the test passes.
Changed the timestamp format to match with the format of the operator log in a real env.

### DIFF
--- a/test/functional/suite_test.go
+++ b/test/functional/suite_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"go.uber.org/zap/zapcore"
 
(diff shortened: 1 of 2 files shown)


test/functional/suite_test.go
```diff
@@ -69,7 +70,11 @@ func TestAPIs(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	opts := zap.Options{
+		Development: true,
+		TimeEncoder: zapcore.ISO8601TimeEncoder,
```
SeanMooney, line 75: cool i like iso time and ya aliging it might be useful if we start adding repoducer tests ectra and want to match on messages ectra.

SeanMooney (approved):

PR #240, opened 2023-01-21
Remove and ignore go.work

gibizer:
Our build system does not support go workspaces[1]. So even though we have two modules in this repository we should not have a go.work file as that will fail the build. Still there are tools that only work properly in our repo if there is go.work file provided. For example go vet only checks the ./api module if the go.work file lists the ./api as the module of the workspace. So a gowork make target is added to generate the go.work file locally but never committed to the repository.

[1]https://github.com/containerbuildsystem/cachito/commit/c7fbbb13eeb7d7edec3c01643eb5b24381027365


bogdando (commented):
/lgtm

PR #241, opened 2023-01-23
Fix conductor ready with 0 replicas requested

gibizer:
There is a bug supporting collapsed cell deployments. If the cell0 and cell1 deployed with the same message bus and then cell0 conductor replica is set to 0 afterwards then the deployment succeeds. But if the deployment started with:

      cellTemplates:
          cell0:
            cellDatabaseUser: nova_cell0
            conductorServiceTemplate:
              replicas: 0
            hasAPIAccess: true
          cell1:
            cellDatabaseUser: nova_cell1
            conductorServiceTemplate:
              replicas: 1
            hasAPIAccess: true

Then NovaConductor=cell0 never becomes Ready as the deployment ready condition is defined as StatefulSet.Status.ReadyCount > 0 . In this special case when 0 replicas is requested by the user the NovaConductor should be ready after the DB sync job succeeded as the user requested not to have any running service in the deployment.

This patch handles this special case.


SeanMooney (approved):
this should only be needed as a temporary measure but it makes sense.

PR #242, opened 2023-01-23
Switch Nova default to cell separation

gibizer:
Now that OpenStackControlPlane supports deploying multiple RabbitMQCluster the default value of Nova.Spec.CellTemplates is changed to have cell0 and cell1 on different message buses to support proper cell separation. The new default assumes that besides to common bus named "rabbitmq" there is another bus name "rabbitmq-cell1" deployed.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/157 (merged)
Depends-On: #241 (merged)
Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/244 (merged)
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/76 (merged)

### DIFF
--- a/api/v1beta1/nova_types.go
+++ b/api/v1beta1/nova_types.go
@@ -50,7 +50,7 @@ type NovaSpec struct {
 	APIMessageBusInstance string `json:"apiMessageBusInstance"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={cell0: {cellDatabaseInstance: openstack, cellDatabaseUser: nova_cell0, cellMessageBusInstance: unused, hasAPIAccess: true, conductorServiceTemplate: {containerImage: "quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo"}}}
+	// +kubebuilder:default={cell0: {cellDatabaseUser: nova_cell0, hasAPIAccess: true}, cell1: {cellDatabaseUser: nova_cell1, cellMessageBusInstance: rabbitmq-cell1, hasAPIAccess: true}}
 	// Cells is a mapping of cell names to NovaCellTemplate objects defining
 	// the cells in the deployment. The "cell0" cell is a mandatory cell in
 	// every deployment. Moreover any real deployment needs at least one
--- a/config/samples/nova_v1beta1_nova.yaml
+++ b/config/samples/nova_v1beta1_nova.yaml
@@ -4,11 +4,3 @@ metadata:
   name: nova
 spec:
   secret: osp-secret
-  cellTemplates:
-    cell0:
-      cellDatabaseUser: nova_cell0
-      # cell0 always needs API access as it hosts the super conductor
-      hasAPIAccess: true
-    cell1:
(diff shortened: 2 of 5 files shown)


api/v1beta1/nova_types.go
```diff
@@ -50,7 +50,7 @@ type NovaSpec struct {
 	APIMessageBusInstance string `json:"apiMessageBusInstance"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={cell0: {cellDatabaseInstance: openstack, cellDatabaseUser: nova_cell0, cellMessageBusInstance: unused, hasAPIAccess: true, conductorServiceTemplate: {containerImage: "quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo"}}}
+	// +kubebuilder:default={cell0: {cellDatabaseUser: nova_cell0, hasAPIAccess: true}, cell1: {cellDatabaseUser: nova_cell1, cellMessageBusInstance: rabbitmq-cell1, hasAPIAccess: true}}
```
SeanMooney, line 53: cellMessageBusInstance: rabbitmq-cell1

ack so this is the primary difference between the cells

we also don't need to include the container image here os +1

config/samples/nova_v1beta1_nova.yaml
```diff
@@ -4,11 +4,16 @@ metadata:
...
 spec:
   secret: osp-secret
+  # This is a collapsed cell setup where only one message bus is used and
+  # cell1 conductor acts both like the super-conductor and the cell1 conductor
   cellTemplates:
     cell0:
       cellDatabaseUser: nova_cell0
-      # cell0 always needs API access as it hosts the super conductor
+      conductorServiceTemplate:
+        replicas: 0
```
SeanMooney, line 13: i dont think we shoudl default to this in our samples

we can have an additional sample that would use this but it should not be in our base one

gibizer: obviously I can switch the sample files to have the cell separation one in `nova_v1beta1_nova.yaml`. But that means the `make nova_deploy` install_yamls target will not work out of the box any more as it would need a second rabbitmq cluster to be deployed by `make rabbitmq_deploy` which it doesn't. I can propose a change in install_yamls to not use the  `nova_v1beta1_nova.yaml` but use a new sample that has the collapsed setup.

gibizer: I've added couple of Depends-Ons to this PR now to land first, then we can do the switch in nova's sample as you requested.

SeanMooney: ack i approved the first one so that should merge shortly 
we can loop back to this once the install yamll pr has merged

gibizer: the install yamls dep is merged now so nothing blocks this any morew

SeanMooney: ack

SeanMooney (approved):

PR #256, opened 2023-02-07
Pass GITHUB_TOKEN to install_kustomize.sh to avoid rate-limit

gibizer:
We are hitting rate-limt during post merge image build:

  Github rate-limiter failed the request. Either authenticate or wait a couple of minutes.

It seems it is coming from [1] and can be fixed by passing the GITHUB_TOKEN there. So this patch does that to avoid build failures due to rate-limit.

[1] https://github.com/kubernetes-sigs/kustomize/blob/0fd385d7197026b92272dc8688ec66593be91fba/hack/install_kustomize.sh#L145

### DIFF
--- a/.github/workflows/build-nova-operator.yaml
+++ b/.github/workflows/build-nova-operator.yaml
@@ -103,6 +103,7 @@ jobs:
         REGISTRY:  ${{ env.imageregistry }}/${{ env.imagenamespace }}
         GITHUB_SHA: ${{ github.sha }}
         BASE_IMAGE: nova-operator
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Get branch name
       id: branch-name


.github/workflows/build-nova-operator.yaml
```diff
@@ -103,6 +103,7 @@ jobs:
         REGISTRY:  ${{ env.imageregistry }}/${{ env.imagenamespace }}
         GITHUB_SHA: ${{ github.sha }}
         BASE_IMAGE: nova-operator
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
SeanMooney, line 106: so kustomize is alredy in the image we have in ci

https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/kubernetes-tools.sh#L27-L30

so I'm not sure why we need to "Pass GITHUB_TOKEN to install_kustomize.sh"

since we don't need to call install_kustomize.sh at all.

SeanMooney (commented):
the title and the PR content dont really match can you explisn this a little more.

this runs on ubuntu-latest and customize is already installed in that image so we should not be installing it in the job.

.github/workflows/build-nova-operator.yaml
gibizer: Here is an example run hitting the issue https://github.com/openstack-k8s-operators/nova-operator/actions/runs/4112604504/jobs/7097759483 

What happens is that the post merge image build github action calls `./create_bundle.sh` see L100 and that calls `make bundle` at https://github.com/openstack-k8s-operators/nova-operator/blob/b3694975183fae6a7bd6f99bc9af9922eda66aeb/.github/create_bundle.sh#L26 that will call the `kustomize` target https://github.com/openstack-k8s-operators/nova-operator/blob/b3694975183fae6a7bd6f99bc9af9922eda66aeb/Makefile#L214 and that will call https://github.com/openstack-k8s-operators/nova-operator/blob/b3694975183fae6a7bd6f99bc9af9922eda66aeb/Makefile#L193-L195 which calls the install_kustomize.sh https://github.com/openstack-k8s-operators/nova-operator/blob/b3694975183fae6a7bd6f99bc9af9922eda66aeb/Makefile#L191

gibizer: All this is probably due to re-using the makefile targets here. So we need to have that make target run basically anyhwhere and therefore it is downloading the kustomize binary locally.

SeanMooney: ok so should we make the make target smarter instead and have it check if kustomise is available in the current path too

that should not be hard to do

gibizer: Then we might need to do that for the rest of the binaries which the current make targets are downloading locally to ./bin and then we have to start checking the versions of those binaries in the path to avoid having an old version of the binary in the path causing troubles. 

So while in the case of kustomize where we don't have version dependency it is tempting to use the global kustomize install if exists. For the general logic of the make file handling binaries independently from the global installation actually shields us against wrong binary versions almost like a python virtual env.

SeanMooney: We can proceed with this for now but we might want to reevaluate this.

in general, the best practice for GitHub actions is only to install things not already in the base image but we may have other requirements that super seed that.

i don't think we actually have any dependency on a specific kustomize version today.
we install one but i did not see that as an upper constraint or lower constraint.

this should not be harmful so ill approve as is.

PR #258, opened 2023-02-08
Pass GITHUB_TOKEN to install_kustomize.sh to avoid rate-limit

gibizer:
The PR #256  fixed this to the post merge image build but missed it to the pre merge image build job. Now it is fixed there.


SeanMooney (approved):
ack

i kind of feel like this should be set in the env of all the jobs but yes this makes sense

PR #233, opened 2023-01-16
ansible compute

SeanMooney:
- add example ansible playbooks

PR #259, opened 2023-02-09
Define NovaExternalCompute CRD

gibizer:
This will be the integration point with the dataplane-operator. That operator will instantiate `NovaExternalCompute` CRs based on the  `OpenStackDataPlaneNode` CRs it reconciles. Then nova-operator will do the nova part of the external data plane deployment on that node based on the information in the CR.

So this PR:
- [x] Defines the NovaExternalCompute CRD
    - [x] with ContainerImage fields to define the images for `openstack-nova-compute` and `openstack-nova-libvirt`
    - [x] with ContainerImage field to define the image for `openstack-tripleo-ansible-ee`
- [x] Ensures that the `Spec.InventoryConfigMapName` points to a proper ConfigMap with an `inventory` key
- [x] Ensures that the `Spec.SSHKeySecretName` poits to a proper Secret with an `ssh-privatekey` key
- [x] Ensures that the `Spec.CellName` refers to an existing `NovaCell` CR and that NovaCell is Ready.


controllers/novaexternalcompute_controller.go
```diff
@@ -109,6 +112,53 @@ func (r *NovaExternalComputeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
+	sshKeyHHash, result, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.SSHKeySecretName},
+		// TODO(gibi): Add the fileds we expect to exists in the SSHKeySecret
+		[]string{},
+		h.GetClient(),
+		&instance.Status.Conditions,
+		r.RequeueTimeout,
+	)
```
slagle, line 142: I suppose DataPlaneNode should probably create the SSH key secret. We could add AnsiblePrivateSSHKey to the Node struct. Are you expecting that?

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -17,25 +17,60 @@ limitations under the License.
...
+
+	// +kubebuilder:validation:Optional
+	// ConfigOverwrite - interface to overwrite default config files like e.g. logging.conf
+	// But can also be used to add additional files. Those get added to the service config dir in /etc/<service> .
+	DefaultConfigOverwrite map[string]string `json:"defaultConfigOverwrite"`
+
+	// +kubebuilder:validation:Required
+	// InventorySecretName is the name of the k8s Secret that contains the ansible inventory information
+	// for this node
+	InventorySecretName string `json:"inventorySecretName"`
```
slagle, line 50: The inventory is currently a configmap in the dataplane-operator. We should probably change that to a secret given the different ansible vars that may contain secret data.

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -17,25 +17,60 @@ limitations under the License.
...
+
+	// +kubebuilder:validation:Required
+	// SSHKeySecretName is the name of the k8s Secret that contains the ssh keys to access the node
+	SSHKeySecretName string `json:"sshKeysSecretName"`
```
slagle, line 54: I'm thinking dataplane-operator creates the inventory and ssh key secret(s) and then passes them in here. Is that correct?

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -17,25 +17,60 @@ limitations under the License.
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	// Deploy true means the nova-operator is allowed to do changes on the external compute node
+	// if necessary. If set to false then the operator will only generate the pre-requisite data (e.g. config maps)
+	// but does not do any modification on the compute node itself.
+	Deploy bool `json:"deploy"`
```
slagle, line 61: I like this, and need the same concept on the dataplane-operator side. I'll probably add the same type there.

controllers/novaexternalcompute_controller.go
SeanMooney: i would expect the sshkey to be an input to the dataplane operator.

for the pre provisioned and operator provisioned case we should have the end user provide an ssh key to use. we can optionaly add sshkey generation later but for now we should just treat it as an input.

api/v1beta1/novaexternalcompute_types.go
SeanMooney: the inventory should not containe andy sensitive data.

it should remain a configmap.

i expect very little data to be in the inventory in general

we shoudl try to avoid using the inventory to pass information to the roles and instead do that by using role var form within the playbooks.

api/v1beta1/novaexternalcompute_types.go
SeanMooney: yes it would pass them here.

i think genertion could be nice eventually.

i would prefer to avoid using root by the way and ensure we use become:true in ansibale as required.

so we should ideally bootstrap a stack user with passwordless sudo and use that for all operation on the edpm nodes. 

if we want to have the dataplane operator do that and generate a key that is ok but
the iniall ssh key used to create that user should be provided by the end user.

api/v1beta1/novaexternalcompute_types.go
SeanMooney: the intent is that the dataplane operator will create teh CR with this set to false initally. it will then collect any pacakges/repos that are rqruied form the cr status and install those as part of the bootstrap phase. after the bootstrap phase is completed in the datapalne operator it can then set this to true to have the nova operator actully deploy nova.

so this allows the dataplane-operator to orchestrate the execution of the ansible playbooks by the service operators.

api/v1beta1/novaexternalcompute_types.go
gibizer: From nova-operator perspective it works either ways (Secret or ConfigMap). But we need to make a decision about it so I can codify that in here.

api/v1beta1/novaexternalcompute_types.go
gibizer: > I'm thinking dataplane-operator creates the inventory and ssh key secret(s) and then passes them in here. Is that correct?

Yes, that is what I expect.

controllers/novaexternalcompute_controller.go
gibizer: the nova-operator will only use the NovaExteranCompute.Spec (and the NovaCell=<NovaExteranCompute.Spec.CellName>.Spec) to get all the inputs needed to do the nova deployment on the edpm node. So we expect that the Secret Name is provided to us in NovaExteranCompute.Spec

gibizer (commented):
Thanks for the feedback!

api/v1beta1/novaexternalcompute_types.go
slagle: It's not just role vars, we have host vars as well. We put host vars into the inventory for now.

api/v1beta1/novaexternalcompute_types.go
slagle: I've got a PR to create the secret: https://github.com/openstack-k8s-operators/dataplane-operator/pull/30

api/v1beta1/novaexternalcompute_types.go
SeanMooney: we are not expecting the data-plane operator to use the inventory to do any configuration of the role or playbook.

If the data-plane operator needs to influence something, it should be modelled in the CR.

The nova-operator will incorporate that into our playbook as required but we don't want to provide a backchannel for things to be modified via the inventory.

api/v1beta1/novaexternalcompute_types.go
gibizer: @slagle thanks! I adapted this PR to ensure the field exists in the Secret nova-operator gets.

api/v1beta1/novaexternalcompute_types.go
slagle: The way you've described this isn't how the existing roles work. edpm_nova_compute/edpm_nova_libvirt already define what packages they need installed. They do not define repos, nor do any other ansible roles within edpm-ansible (and they won't).

If you are writing your own ansible roles for nova_compute and nova_libvirt and not using edpm-ansible (first I'm learning of this), than those will need to install the packages they need, and be written in a way that allow you to execute (via openatack-ansibleee-operator) different "pieces" (bootstrap, run, etc) separately. The standalone-roles were well socialized and shared, including a lot of input from the Compute DFG. If you choose not to use them now, that is fine, but our time to adapt how dataplane-operator will execute with the role interfaces we already expect has passed for Q1 already.

We already created the roles in edpm-ansible to work such that bootstrap/configure/etc can be run separately (and even omitted if needed in the case of configure.yml). I'm not aware of how you intend the roles you are writing to work. But, there's no concept of an open ended interface to install any list of packages via dataplane-operator. That needs to be expressed within an ansible role. Again, we already have that in edpm-ansible, but if you are doing your own thing, then it needs to be in the role you're maintaining.

gibizer: I clearly see a disagreement about what ansible code nova-operator runs on an edpm role. I think this discussion belongs to the PR https://github.com/openstack-k8s-operators/nova-operator/pull/233 . This PR in the current form does not prescribe any of the proposed directions.

SeanMooney: The standalone roles were desigined for use with tripleo and we let you know that we expected them to be adapted sginifcalty before they were reused with the operators and expected to refactor the interface. for example all of the use of puppet of config generation was to be removed or made optional and disabled. there were also part fo the roles that need to be separated out into other roles.

since then the edpm-ansibel repo has been created with the winding down of tripelo investment we can add a package install role if we wanted too but we discussed that we did not want to filelock or similar to do synconsiation between different Ansible EE executions to ensure only one ran at a time. This was discussed in design doc which we shared with you for the integration between the podifed control plan and edpm nodes. i thought we had agreement form you on that design before we started this.

if we want too i can delete the existing edpm_nova_compute/edpm_nova_libvirt roles and replace them with my much simpler ones however as I express that will break the current support in install_yamls which I asked not to be written last year.
the reason I asked for people not to work on automating compute node deployment for the end-of-year demo was we knew we were not going to use the roles as they were and did not want people to invest time in trying to deploy computes with them.

gibi is correct however this is not related to this PR since this is the go part but the other pr #233.

#233 is now feature complete and im now refactoring it to prepare for integrating with this PR.
i will be deleting un needed files and rebasing it today as well as squashing most of the commits.

gibizer: @slagle @SeanMooney : would it work for both of you to temporary have the nova and libvirt roles in the nova-operator repo to make progress. When we have the nova-operator part ready and working then I think we can move those roles to the edpm-ansible repo. I believe @SeanMooney has concerns about the current edpm_nova_compute/edpm_nova_libvirt roles there. So when we have Sean's proposal working we can compare the roles and discuss the reason for the differences.

slagle: Yea, it's fine, I don't have a problem with that if you want to maintain the roles here. I cannot understand what is being proposed beyond that with the other needed packages/repos. The existing roles from edpm-ansible for compute/libvirt don't have any packages to install, and no roles in edpm-ansible enable repos.

I suggest you just manage that within the roles you're developing. I hesitate to add an open ended interface in dataplane-operator to run one off tasks to install packages or enable repos or whatever else might come along. Those tasks need to be in a role before dataplane-operator will execute it. It'd be a hard coded list of packages anyway wouldn't it? Again, I'm unable to follow the written description, but if we can exaplain what packages, what's the use case, etc, it might make more sense.

We will eventually have a rhsm/subscription-manager role for repo management in edpm-ansible for product deployments, that will already have run on dataplane nodes by the time NovaExternalCompute is created. I don't know if you intend to enable repos a different way.

SeanMooney: no we just intened to have a way to pass two list back

a list of repos to enable via rhsm and a set of packages to install/remove before the nova operator runs ansibel.

right now we don't need this mechanism at all since nova requires no packages to be installed on the host.

eventually we may deconstiarnis libvirt in which case we have two options.
1 just have the dataplane operate install libvirt by using a role in edpm-ansible
2 have the nova operator install it.

in both option 1 and 2 either the nova operator or datapalne operator  could configure libvirt

what we were trying to solve with this design was 1 ensure there is an entity for ensuring packages are installed or absent, 2 ensuring we don't have race conditions between services, 3 ensuring the repos/subscripts are managed properly by something.

we had planned to decontainerise libvirt before then end of Q1, we got some push back form pm on that so this will be a q2 topic.

there are questions about also deconstiernising iscsid, multipath just like ovs and tuned are deconstierised today.

so in the future we may have 6 package that need to be installed on edpm nodes
ceph-common,libvirt, multipath,iscsid and ovs.

the only one that is even remotely in the scope of the nova operator is libvirt and that a stretch but we can live with that.

the other way i was thinking of doing this was to have a set of boolean flags on the openstackDataplaneNode CR

deploy_ovs, deploy_libvirt deploy_...

when set to true the dataplane operator could just include the corresponding roles to provisions that service at the rhel level.

if we would prefer to do it that way we can i don't mind writing some of the support for that in the dataplane operator once we complete the nova-compute integration.

api/v1beta1/novaexternalcompute_types.go
slagle: Can we switch this to ConfigMap? The inventory is a ConfigMap on the dataplane-operator side: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/controllers/openstackdataplanenode_controller.go#L212

gibizer: I can switch it for sure. I just waited for the discussion to settle if this has sensitive data or not.

api/v1beta1/novaexternalcompute_types.go
bogdando: > The standalone roles were desigined for use with tripleo
this at is not entirely true. Just some naming rules and variables defaults assume the roles might have some use for tripleo (as well), but not exclusivley

api/v1beta1/novaexternalcompute_types.go
bogdando: it should not. otherwise we should consider such infor as a bug

pkg/nova/common.go
```diff
@@ -83,3 +88,25 @@ type MessageBus struct {
...
+
+// AddIndexers adds field indexers to the cache used by the client for Get and
+// List calls.
+// By default the cache can only be filtered by namespace so if the
+// want to filter on other fields then such field needs and indexer added here
+func AddIndexers(mgr manager.Manager) error {
+	cache := mgr.GetCache()
+	ctx := context.TODO()
+
+	err := cache.IndexField(
```
bogdando, line 100: this notes: "If you want
compatibility with the Kubernetes API server, only return one key"

I hope this is fine that we return a slice in its func()?

gibizer: I'm not sure I follow. Anyhow the need of querying NovaCell with a field matcher went away so I dropped the extra index too.

gibizer:
I consider this as ready to land. Sean is working on the adding config gen and ansible execution in #268

gibizer:
fixed the lint failrue

gibizer:
We need to allow more time to pre-commit to run in CI

```
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
[...]
[36mINFO[0m[2023-03-02T15:41:28Z] Ran for 13m46s                               
[31mERRO[0m[2023-03-02T15:41:28Z] Some steps failed:                           
[31mERRO[0m[2023-03-02T15:41:28Z] 
  * could not run steps: step precommit-check failed: test "precommit-check" failed: the pod ci-op-zmlw1qfs/precommit-check failed after 8m49s (failed containers: test): ContainerFailed one or more containers exited
  ```

File: controllers/novaexternalcompute_controller.go [hunk 6]
```diff
@@ -0,0 +1,329 @@
...
+
+// NovaExternalComputeReconciler reconciles a NovaExternalCompute object
+type NovaExternalComputeReconciler struct {
+	ReconcilerBase
+}
+
+//+kubebuilder:rbac:groups=nova.openstack.org,resources=novaexternalcomputes,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=nova.openstack.org,resources=novaexternalcomputes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=nova.openstack.org,resources=novaexternalcomputes/finalizers,verbs=update
+
```
gibizer, line 45: some rbac rules are missing from here like Secret and ConfigMap

controllers/novaexternalcompute_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	// Always update the instance status when exiting this function so we can
+	// persist any changes happend during the current reconciliation.
+	defer func() {
+		// update the overall status condition if service is ready
+		if allSubConditionIsTrue(instance.Status) {
+			instance.Status.Conditions.MarkTrue(
+				condition.ReadyCondition, condition.ReadyMessage)
+		}
```
gibizer, line 94: probably make sense to add the else branch from https://github.com/openstack-k8s-operators/nova-operator/pull/271 here too

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo"
+	// NovaLibvirtContainerImage is the Container Image URL for the nova-libvirt container
+	NovaLibvirtContainerImage string `json:"novaLibvirtContainerImage"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
+	// AnsibleEEContainerImage is the Container Image URL for the ansible execution environment
+	AnsibleEEContainerImage string `json:"ansibleEEContainerImage"`
```
SeanMooney, line 82: we have merged the support for the network-attachments since you started on this and the AEE CR support them so it would be good to add them to the CR so i can pass them forward.

gibizer:
> We need to allow more time to pre-commit to run in CI
> 
> ```
> level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> [...]
> �[36mINFO�[0m[2023-03-02T15:41:28Z] Ran for 13m46s                               
> �[31mERRO�[0m[2023-03-02T15:41:28Z] Some steps failed:                           
> �[31mERRO�[0m[2023-03-02T15:41:28Z] 
>   * could not run steps: step precommit-check failed: test "precommit-check" failed: the pod ci-op-zmlw1qfs/precommit-check failed after 8m49s (failed containers: test): ContainerFailed one or more containers exited
> ```
> 
> This is most probably a fallout of the recent change that limited the memory usage of the pre-commit job

This is being fixed in https://github.com/openstack-k8s-operators/nova-operator/pull/278

SeanMooney (changes requested):
i think we should probably remove the conversion webhook and ca injection webhook until we can implement it properly 

we will need to take a ca secret as an input for the webhook likely this should be passed to the nova CR and then used for all webhooks for all the nova* CR types.

config/crd/patches/cainjection_in_novaexternalcomputes.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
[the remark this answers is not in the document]
SeanMooney: actully this is an exsiting issue for all our CRD definitions

since its not specific to this CRD I'm ok with addressing this in a separate PR.

the  current webhook definition will not really be used so we can clean this up later.

SeanMooney (approved):

config/crd/patches/cainjection_in_novaexternalcomputes.yaml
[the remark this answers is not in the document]
gibizer: Yeah I believe this is coming directly from the operator-sdk scaffolding so I think we can keep that in place. Definitely there will a different PR to introduce the defaulting and validating webhooks to Nova (and the interface of the webhooks in the openstack-operator to use them).

PR #268, opened 2023-02-21
add compute node ansible support

SeanMooney:
This is the inital commit in the support for external compute
hosts managed via ansible. This is created from my poc in PR #233
by taking the final working version as a base.

Following commits will refactor this to be more generic and
relocate the playbooks and configs so that they can be integrated
with the golang operator.


controllers/novaexternalcompute_controller.go
```diff
@@ -119,28 +120,71 @@ func (r *NovaExternalComputeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	// 	[]string{
+	// 		"inventory",
+	// 	},
+	// 	h.GetClient(),
+	// 	&instance.Status.Conditions,
+	// 	r.RequeueTimeout,
+	// )
+
+	// --------- temp hack start -------
+	invConfigmap := &corev1.ConfigMap{}
```
SeanMooney, line 136: this sould just be an ensure configmap function

controllers/novaexternalcompute_controller.go
```diff
@@ -119,28 +120,71 @@ func (r *NovaExternalComputeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	// --------- temp hack end -------
 	hashes[instance.Spec.InventorySecretName] = env.SetValue(inventoryHash)
 
 	sshKeyHHash, result, err := ensureSecret(
 		ctx,
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.SSHKeySecretName},
 		// NOTE(gibi): Add the fields here we expect to exists in the SSHKeySecret
 		[]string{
-			"ssh_private_key",
+			"ssh-privatekey",
```
gibizer, line 187: Where this change is coming from? I synced with the description in https://github.com/openstack-k8s-operators/dataplane-operator/blob/9291ba1ab72a812a33172a5296ea72d393e04697/api/v1beta1/openstackdataplanenode_types.go#L93 and with the implementation here https://github.com/openstack-k8s-operators/dataplane-operator/blob/9291ba1ab72a812a33172a5296ea72d393e04697/pkg/util/ansible_execution.go#L74

SeanMooney: from https://kubernetes.io/docs/concepts/configuration/secret/#use-case-pod-with-ssh-keys

if you create a secrete with  --type 'kubernetes.io/ssh-auth' then the name of the private and public key are validated by the k8s api.

oc create secret generic dataplane-ssh-private-key-compute1  --type 'kubernetes.io/ssh-auth' --from-file=ssh-privatekey=/home/centos/.ssh/id_ed25519_oks --from-file=ssh-publickey=/home/centos/.ssh/id_ed25519_oks.pub

SeanMooney: filed https://github.com/openstack-k8s-operators/dataplane-operator/issues/52 to rack this

controllers/novaexternalcompute_controller.go
```diff
@@ -280,13 +324,10 @@ func (r *NovaExternalComputeReconciler) ensureCell(
 ) (*novav1.NovaCell, ctrl.Result, error) {
 	// We are looking for the NovaCell instance where
 	// NovaCell.Spec.CellName == NovaExternalCompute.Spec.CellName
-	options := []client.ListOption{
-		client.InNamespace(instance.Namespace),
-		client.MatchingFields{"spec.cellName": instance.Spec.CellName},
-	}
-	cells := &novav1.NovaCellList{}
+	cellName := types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.CellName}
```
gibizer, line 327: That is not correct. The NovaCell.Name is not the same as the key in the Nova.Spec.cellTemplate. As NovaCell.Name is a generated from the Nova.Name and the Nova.Spec.cellTemplate key. 
In the other hand the NovaCell.Spec.CellName is the same as NovaExternalCompute.Spec.CellName. Hence my Client.List call with the field filter to look the NovaCell up.

gibizer: We discussed this during our sync call. The agreement there was to:
* add a NovaName field to the NovaExternalCompute CRD and default it to "nova". The same value that is hardcoded in  openstack-operator as the name of the Nova CR it creates. https://github.com/openstack-k8s-operators/openstack-operator/blob/9fd6272787f0625b2be2f995b5ab5e18e08be8d7/pkg/openstack/nova.go#L38
* keep the NovaExternalCompute.Spec.NovaCell as a non prefixed cell name matching the Nova.Spec.cellTemplate keys.
* the reconciler of NovaExternalCompute should do the same logic to generate the name of the NovaCell CR to look up than the Nova reconciler use to generate that name in the first place.

Long term the NovaName can be something that is passed down from the openstack-operator to the dataplane-operator to support other than the hardcoded "nova" string.

gibizer: Implemented this in the newest version of https://github.com/openstack-k8s-operators/nova-operator/pull/259

controllers/novaexternalcompute_controller.go
gibizer: Done in #259

controllers/novaexternalcompute_controller.go
gibizer: fixed in nova-operator in #259

templates/novacompute/nova_compute.json
```diff
@@ -0,0 +1,23 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-compute:current-tripleo",
```
gibizer, line 2: added CRD paramteter for it

templates/libvirt_virtlogd/libvirt_virtlogd.json
```diff
@@ -0,0 +1,19 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo",
```
gibizer, line 2: added a CRD paramter for it

templates/libvirt_virtnodedevd/libvirt_virtnodedevd.json
```diff
@@ -0,0 +1,19 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo",
```
gibizer, line 2: added CRD paramter for it

templates/libvirt_virtproxyd/libvirt_virtproxyd.json
```diff
@@ -0,0 +1,18 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo",
```
gibizer, line 2: added CRD parameter for it

templates/libvirt_virtqemud/libvirt_virtqemud.json
```diff
@@ -0,0 +1,22 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo",
```
gibizer, line 2: added CRD parameter for it

templates/libvirt_virtsecretd/libvirt_virtsecretd.json
```diff
@@ -0,0 +1,18 @@
+{
+    "image": "quay.io/tripleozedcentos9/openstack-nova-libvirt:current-tripleo",
```
gibizer, line 2: added CRD parameter for it

SeanMooney:
that was just a rebase to the top of gibs series but i do not think i will need any further changes in pr #259 ath this point

controllers/novaexternalcompute_controller.go
```diff
@@ -0,0 +1,439 @@
...
+}
+
+func (r *NovaExternalComputeReconciler) generateConfigs(
+	ctx context.Context, h *helper.Helper, instance *novav1.NovaExternalCompute,
+	cell *novav1.NovaCell, hashes *map[string]env.Setter,
+) error {
+	secret := &corev1.Secret{}
+	// we will lookup the cell based on the current namesapce of the cell.
+	// however this should be the same as the current namespace of the instance.
+	// should we assert that?
```
gibizer, line 363: ensureCellReady uses the NovaExternalCompute.Namespace to look for the NovaCell. So it is already ensured that the NovaCell and the NovaExternalCompute is in the same namespace.

controllers/novaexternalcompute_controller.go
```diff
@@ -0,0 +1,439 @@
...
+		Name:      cell.Spec.CellMessageBusSecretName,
+	}
+	err = h.GetClient().Get(ctx, secretName, cellMessageBusSecret)
+	if err != nil {
+		util.LogForObject(
+			h, "Failed reading Secret", instance,
+			"CellMessageBusSecretName", cell.Spec.CellMessageBusSecretName)
+		return err
+	}
+
```
gibizer, line 386: I'm OK to treat all Secret reading errors here as hard errors as we ensured above that the NovaCell is ready so it was able to read these Secrets so they should exists and well structured.

controllers/novaexternalcompute_controller.go
```diff
@@ -0,0 +1,439 @@
...
+	instance.Status.Conditions.MarkTrue(
+		novav1.NovaCellReadyCondition, novav1.NovaCellReadyMessage, cellCRName)
+
+	return cell, ctrl.Result{}, err
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NovaExternalComputeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&novav1beta1.NovaExternalCompute{}).
```
gibizer, line 335: you are generating ConfigMaps so you should state that you Own them here.

SeanMooney: yes and i also need to set the condition to ready

gibizer (commented):
I'm OK with the folder structure in `templates`

controllers/novaexternalcompute_controller.go
```diff
@@ -437,3 +444,72 @@ func (r *NovaExternalComputeReconciler) generateConfigs(
...
+		h, "using playbooks for instance ", instance, "from ", playbookPath,
+	)
+	playbookDirEnteries, err := os.ReadDir(playbookPath)
+	if err != nil {
+		return err
+	}
+	// EnsureConfigMaps is not used as we do not want the templating behavior that adds.
+	configMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        fmt.Sprintf("%s-external-compute-playbooks", instance.Spec.NovaInstance),
```
gibizer, line 472: +1 what we storing in this CM only depends on the version of the nova-operator so it is OK to have a high level name.

controllers/novaexternalcompute_controller.go
```diff
@@ -437,3 +444,72 @@ func (r *NovaExternalComputeReconciler) generateConfigs(
...
+				"Nova ", instance.Spec.NovaInstance,
+			)
+			return err
+		}
+		// we assocate this with the nova cr that manages the external dataplane node because
+		// this shoudl not vary per compute node
+		configMap.Labels = labels.GetLabels(
+			novaInstance, labels.GetGroupLabel(NovaLabelPrefix), map[string]string{},
+		)
+
```
gibizer, line 498: I think if we don't put a controller / owner ref to the CM then we can potentially leak this CM when every Nova related CR is deleted. I think we should put a controller reference to the CM that connects it to the Nova CR with 
`controllerutil.SetControllerReference(novaInstance, configMap, h.GetScheme())`. I think that will not cause any reconcile cycle. But the CM will be deleted when the Nova CR is deleted.

controllers/novaexternalcompute_controller.go
```diff
@@ -82,6 +85,10 @@ func (r *NovaExternalComputeReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
 	l.Info("Reconciling")
 
+	if err = r.ensurePlaybooks(ctx, h, instance); err != nil {
+		return ctrl.Result{}, err
+	}
+
```
gibizer, line 91: right now having this here before initStatus or the creation of the defer func is OK as we don't want to record any status during ensurePlaybooks() but I think to avoid later confusion we should move this below the defer definition.

gibizer (commented):
overall looks good I had some minor comments inline

controllers/novaexternalcompute_controller.go
SeanMooney: ack i wanted to do it before we started doing the heavy lifting but that makes sense i can put it after the defer that's still pretty early.

controllers/novaexternalcompute_controller.go
SeanMooney: ya makes sense

tools/ansible/inventory/bootstrap-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+                  ctlplane_gateway_ip: 10.0.2.1
+                  ctlplane_dns_nameservers:
+                    - 172.30.0.10
+                    - 8.8.8.8
+                  dns_search_domains: []
+                  # this is the ovn sb db service ClusterIP
+                  edpm_ovn_dbs:
+                  - 172.30.251.145
+            vars:
+              edpm_ovn_controller_image: quay.io/tripleomastercentos9/openstack-ovn-controller:current-tripleo
```
bogdando, line 53: how do we use this ovn image in nova config tasks?

gibizer:
When I tried to use the aee/bootstrap.yaml to simulate the work of the dataplane-operator I get

```
$ oc logs -f bootstrap-pbgf9
Identity added: /runner/artifacts/f96dca9a-4751-4b87-b585-4956e613e51b/ssh_key_data (gibi@bedrock)
ERROR! Unable to retrieve file contents
Could not find or access '/usr/share/ansible/edpm-playbooks/deploy-edpm-selinux.yml' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

I'm not sure how that playbook should come from as I don't find a corresponding file either in nova-operator or in edpm-ansible.

tools/ansible/playbooks/bootstrap.yaml
```diff
@@ -0,0 +1,37 @@
...
+          sudo python3 setup.py install
+          popd
+          sudo /usr/local/bin/tripleo-repos current-tripleo-dev
+        creates: /usr/local/bin/tripleo-repos
+# given we were only ment to be reusing the roles this feels a bit wrong but
+# again this should be ok for this bootstrap playbook for now.
+- name: Deploy EDPM facts playbook
+  ansible.builtin.import_playbook: deploy-edpm-facts.yml
+- name: Deploy EDPM SELinux playbook
+  ansible.builtin.import_playbook: deploy-edpm-selinux.yml
```
bogdando, line 25: is the mentioned failure related to this import?

tools/ansible/inventory/bootstrap-configmap.yaml
SeanMooney: we dont as i said in the past the bootstrap part are emulating what the datapalen operator does its not going to be in the repo long term.

its only here so we can actully prep a vm without needing the dataplane operator to be fully finsihed.

tools/ansible/playbooks/bootstrap.yaml
SeanMooney: it was moved/removed by 

https://github.com/openstack-k8s-operators/edpm-ansible/commit/bd427c69d1f11f8df19e1862530be061391f243b

it existed when i first wrote this.

reading the commit it now part of the edpm_bootstrap role.

so yes ill drop it

controllers/novaexternalcompute_controller.go
SeanMooney: done

controllers/novaexternalcompute_controller.go
SeanMooney: done

tools/ansible/playbooks/bootstrap.yaml
SeanMooney: done

controllers/novaexternalcompute_controller.go
```diff
@@ -184,6 +204,47 @@ func (r *NovaExternalComputeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		// then check if they have compelted and requeue if not.
+		if libvirtAEE.Status.JobStatus != "Succeeded" {
+			return ctrl.Result{RequeueAfter: r.RequeueTimeout}, fmt.Errorf(
+				"libvirt deployment %s for %s", libvirtAEE.Status.JobStatus, instance.Name,
+			)
+		}
+		if novaAEE.Status.JobStatus != "Succeeded" {
+			return ctrl.Result{RequeueAfter: r.RequeueTimeout}, fmt.Errorf(
+				"nova deployment %s for %s", novaAEE.Status.JobStatus, instance.Name,
+			)
```
gibizer, line 229: * We don't need to return an error here as this is not an error (and it produces nosy logging)
* We probably don't need to requeue here either as added `Owns(&aee.OpenStackAnsibleEE{}).` to the controller so we get reconciled when the status of the aee.OpenStackAnsibleEE changes.

controllers/novaexternalcompute_controller.go
```diff
@@ -333,5 +392,391 @@ func (r *NovaExternalComputeReconciler) ensureCellReady(
...
+	}
+	err := h.GetClient().Get(ctx, aeeName, ansibleEE)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ansibleEE, err
+	}
+
+	if k8s_errors.IsNotFound(err) {
+		_, err = controllerutil.CreateOrPatch(ctx, h.GetClient(), ansibleEE, func() error {
+			initAEE(instance, ansibleEE, "deploy-nova.yaml")
+
```
gibizer, line 647: you probably want to add a controller reference

controllers/novaexternalcompute_controller.go
```diff
@@ -333,5 +392,391 @@ func (r *NovaExternalComputeReconciler) ensureCellReady(
...
+	}
+	err := h.GetClient().Get(ctx, aeeName, ansibleEE)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ansibleEE, err
+	}
+
+	if k8s_errors.IsNotFound(err) {
+		_, err = controllerutil.CreateOrPatch(ctx, h.GetClient(), ansibleEE, func() error {
+			initAEE(instance, ansibleEE, "deploy-libvirt.yaml")
+
```
gibizer, line 603: you probably want to add a controller reference

controllers/novaexternalcompute_controller.go
```diff
@@ -333,5 +392,391 @@ func (r *NovaExternalComputeReconciler) ensureCellReady(
...
+	return ansibleEE, nil
+}
+
+func initAEE(
+	instance *novav1.NovaExternalCompute, ansibleEE *aee.OpenStackAnsibleEE,
+	playbook string,
+) {
+	ansibleEE.Spec.Image = instance.Spec.AnsibleEEContainerImage
+	// TODO we dont currently have this on the NovaExternalComputeCR
+	// ansibleEE.Spec.NetworkAttachments = instance.Spec.NetworkAttachments
```
gibizer, line 670: we have it now

controllers/novaexternalcompute_controller.go
```diff
@@ -333,5 +392,391 @@ func (r *NovaExternalComputeReconciler) ensureCellReady(
...
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("error create/updating configmap: %w", err)
+	}
+
+	return nil
+}
+
+func (r *NovaExternalComputeReconciler) ensureAEEDeployLibvirt(
```
gibizer, line 576: ensureAEEDeployLibvirt and ensureAEEDeployNova are fairly similar. Can we create a single function? I think only the name of the CR and the name of the playbook file are different.

templates/novacompute/nova-compute.json
```diff
@@ -0,0 +1,36 @@
+{
+    "command": "nova-compute",
+    "config_files": [
+        {
+            "source": "/var/lib/openstack/config/nova-blank.conf",
+            "dest": "/etc/nova/nova.conf",
```
gibizer, line 6: It does not seems to work. At least in my test env the nova_compute container will have /etc/nova/nova.conf filled with a config sample.

gibizer (changes requested):

controllers/novaexternalcompute_controller.go
SeanMooney: yes your right, and ya i noticed the noices i was consideing changing it before but i wasnt sure what to replace this with.

i think i night just return normally.

im wondering if i should should log an info message or not.

controllers/novaexternalcompute_controller.go
SeanMooney: we could yes ill add it to my todo list

templates/novacompute/nova-compute.json
SeanMooney: this really should work but ill try and look into this in a followup pr

https://github.com/openstack/kolla/blob/master/doc/source/admin/kolla_api.rst#format-of-the-configuration-file

the kolla image api is pretty clear that the default behavior should be to replace.

unless we have a bug in the tripleo version of the script.

bogdando: the forked tripleo kolla code for this lives here https://github.com/openstack/tripleo-common/blob/2e59553afa7fa742d93be9f40575ccf7b2983c72/container-images/kolla/base/set_configs.py#L74

I can't see how could this fail to copy the blank file over the existing file...

SeanMooney: it also works for me locally

[centos@nova-compute-1 ~]$ sudo podman exec -it nova_compute bash
bash-5.1$ cat /etc/nova/nova.conf
bash-5.1$

gibizer: I see the issue. In the compute container there is a nova-compute.conf also copied by kolla and that is the one containing the config sample.

```
[root@edpm-compute-0 ~]# podman exec -it nova_compute wc -l /etc/nova/nova.conf
0 /etc/nova/nova.conf
[root@edpm-compute-0 ~]# podman exec -it nova_compute wc -l /etc/nova/nova-compute.conf
726 /etc/nova/nova-compute.conf
[root@edpm-compute-0 ~]# 
```

gibizer: So we need to copy the blank file over nova-compute.conf

test/functional/nova_controller_test.go
```diff
@@ -177,7 +177,13 @@ var _ = Describe("Nova controller", func() {
 					},
 				),
 			)
-			DeferCleanup(th.DeleteKeystoneAPI, th.CreateKeystoneAPI(namespace))
+			keystoneApiName := th.CreateKeystoneAPI(namespace)
+			DeferCleanup(th.DeleteKeystoneAPI, keystoneApiName)
+			keystoneApi := th.GetKeystoneAPI(keystoneApiName)
+			keystoneApi.Status.APIEndpoints["internal"] = "http://keystone-internal-openstack.testing"
```
gibizer, line 183: yepp this needs to move to the CreateKeystoneAPI helper in lib-common

gibizer (commented):
I tested this by built it into openstack-operator.

from install_yamls/devsetup:
* make crc_attach_default_interface
* make edpm_compute

from install_yamls:
* make crc_storage
* make nmstate
* make nncp
* make netattach
* make metallb
* make metallb_config
* OPENSTACK_IMG=quay.io/gibi/openstack-operator-index:v0.0.75 make openstack
* OPENSTACK_CR=../openstack-operator/config/samples/core_v1beta1_openstackcontrolplane_network_isolation.yaml make openstack_deploy

* then in the edpm-compute-0 VM I needed to run `dnf  update` to get a latest podman to avoid crashes

* then followed the ./tools/ansible/readme.md to bootstrap the compute
* then applied the following external compute sample to trigger the nova deploy
```
$ cat config/samples/nova_v1beta1_novaexternalcompute.yaml 
kind: NovaExternalCompute
metadata:
  name: compute1
spec:
  inventoryConfigMapName: compute1-inventory
  sshKeySecretName: dataplane-ansible-ssh-private-key-secret
```
Then I needed to restart the nova-scheduler to pick up cell1 registration
* oc delete pod/nova-scheduler-0

Then I just created the openstack bits to run a VM:
* wget http://download.cirros-cloud.net/0.5.2/cirros-0.5.2-x86_64-disk.img
* openstack image create --file cirros-0.5.2-x86_64-disk.img cirros-0.5.2-x86_64-disk --container-format bare --disk-format qcow2
* openstack flavor create c1 --vcpus 1 --ram 256
* openstack network create private
* openstack subnet create private-subnet --network private --subnet-range 10.0.0.0/26
* openstack --os-compute-api-version 2.93 server create --flavor c1 --image cirros-0.5.2-x86_64-disk --nic net-id=private --use-config-drive vm1

gibizer (approved):
TODOs left for followup:
* do not requeu while waiting for the ansible jobs
* add controller ref to the ansible jobs we create to not leak them if nova is deleted while the job running
* pass NetworkAttachments to the ansiblee CR
* create a common ensureAEEDeploy function from the two similar one
* blank out nova-compute.conf
* add envtest coverage to the negative ansible execution case.

PR #269, opened 2023-02-22
Run pre-commit with make

amolkahat:
Updating pre-commit hook with make. pre-commit-golang repo is no longer maintained[1]. Instead of using
unmaintained repo we can use make to run pre-commit.

[1] https://github.com/dnephin/pre-commit-golang/issues/98

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,21 +1,37 @@
 repos:
-- repo: https://github.com/dnephin/pre-commit-golang
-  rev: v0.5.1
-  hooks:
-    - id: go-fmt
-      exclude: ^vendor
-    - id: go-vet
-    - id: go-mod-tidy
-    - id: go-lint
-
-- repo: https://github.com/golangci/golangci-lint
-  rev: v1.50.1
-  hooks:
-    - id: golangci-lint
-      args: ["--verbose"]
-
 - repo: local
   hooks:
+    - id: golangci-lint
+      name: golangci-lint
+      language: golang
+      types: [go]
+      entry: make
+      args: ["golangci-lint"]
+      pass_filenames: false
+    - id: golint
+      name: golint
+      language: system
+      entry: make
+      args: ["golint"]
+      pass_filenames: false
+    - id: gofmt
+      name: gofmt
+      language: system
+      entry: make
+      args: ["fmt"]
+      pass_filenames: false
+    - id: govet
+      name: govet
+      language: system
+      entry: make
+      args: ["vet"]
+      pass_filenames: false
+    - id: gotidy
+      name: gotidy
+      language: system
+      entry: make
+      args: ["tidy"]
+      pass_filenames: false
     - id: make-manifests
       name: make-manifests
       language: system
@@ -62,4 +78,13 @@ repos:
   rev: 2.1.1
   hooks:
     - id: bashate
-      entry: bashate --error . --ignore=E006,E040,E043
+      entry: bashate --error . --ignore=E006,E040,E020,E012
+      # Run bashate check for all bash scripts
+      # Ignores the following rules:
+      # E006: Line longer than 79 columns (as many scripts use jinja
+      #       templating, this is very difficult)
+      # E040: Syntax error determined using `bash -n` (as many scripts
+      #       use jinja templating, this will often fail and the syntax
+      #       error will be discovered in execution anyway)
+      # E020: Function declaration not in format ^function name {$
+      # E012: here doc didn't end before EOF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,19 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./api/...
 
+.PHONY: tidy
+tidy: fmt
+	go mod tidy; \
+	pushd "$(LOCALBIN)/../api"; \
+	go mod tidy; \
+	popd
+
+.PHONY: golangci-lint
+golangci-lint:
+	# NOTE this will install golangci-lint in to local bin dir
+	test -s $(LOCALBIN)/golangci-lint || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.51.2
+	$(LOCALBIN)/golangci-lint run --fix
+
 PROCS?=$(shell expr $(shell nproc --ignore 2) / 2)
 PROC_CMD = --procs ${PROCS}
 


.pre-commit-config.yaml
```diff
@@ -1,19 +1,31 @@
 repos:
-- repo: https://github.com/dnephin/pre-commit-golang
```
gibizer, line 2: OK this is sunset so we probably don't want to depend on it. There are viable alternatives if we want e.g.: https://github.com/TekWizely/pre-commit-golang

Makefile
```diff
@@ -121,6 +121,15 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./api/...
 
+.PHONY: tidy
+tidy: fmt
+	go mod tidy
```
gibizer, line 126: This will not tidy both go.mod files our repo has.

Makefile
```diff
@@ -121,6 +121,15 @@ vet: gowork ## Run go vet against code.
...
+
+.PHONY: golangci-lint
+golangci-lint:
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.51.2
```
gibizer, line 130: * how this makes sure that the golangci-lint binary actually get installed to $(LOCALBIN) used below?
* I don't like the idea that this downloads and installs the binary at every run

gibizer (changes requested):

.pre-commit-config.yaml
amolkahat: Idea is just use whatever we have there already.. so using pre-commit we can run make commands.

Makefile
amolkahat: Can you suggest something here.. what need to use? I'm open for suggestions.

Makefile
amolkahat: * Script `install.sh` install it in the local directory `./bin/`[1].
* Ok. I'll add some checks for it.
 
[1] https://golangci-lint.run/usage/install/

Makefile
gibizer: what I would do is along the line:
```make
tidy:
    cd ./api
    go mod tidy
    cd ../
    go mod tidy
```

Makefile
gibizer: > * Script `install.sh` install it in the local directory `./bin/`[1].

please add note about it as a comment

> 
>     * Ok. I'll add some checks for it.
> 
> 
> [1] https://golangci-lint.run/usage/install/

Makefile
amolkahat: This also failed in CI
``` 
go mod tidy
go: go.mod file indicates go 1.19, but maximum version supported by tidy is 1.18
make: *** [Makefile:126: tidy] Error 1
```

gibizer: It means the image the job uses don't have go 1.19 installed.

Makefile
amolkahat: Done.

Makefile
amolkahat: Updated with your comments. Resolving discussion.

Makefile
```diff
@@ -121,6 +121,17 @@ vet: gowork ## Run go vet against code.
 	go vet ./...
 	go vet ./api/...
 
+.PHONY: tidy
+tidy: fmt
+	cd ./api; go mod tidy
+	go mod tidy
```
SeanMooney, line 127: i dont think this does what we want it to

you should probably use pushd and popd or similar
i would expect this to do two execution fo go mod tidy in the API dir

SeanMooney: that or revers the order
do it for the top level go mod file then the api

amolkahat: Done.

.pre-commit-config.yaml
```diff
@@ -1,19 +1,31 @@
...
-
 - repo: local
   hooks:
+    - id: golangci-lint
+      name: golangci-lint
+      language: golang
+      types: [go]
+      entry: make
+      args: ["golangci-lint"]
+      pass_filenames: false
```
SeanMooney, line 10: passing false actully feel like a regression here.

we really only want it to check the files you are modifying

SeanMooney: actully i guess it dose not use it currently 
anyway 

https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml

Makefile
```diff
@@ -121,6 +121,19 @@ vet: gowork ## Run go vet against code.
...
+	go mod tidy; \
+	pushd "$(LOCALBIN)/../api"; \
+	go mod tidy; \
+	popd
+
+.PHONY: golangci-lint
+golangci-lint:
+	# NOTE this will install golangci-lint in to local bin dir
+	test -s $(LOCALBIN)/golangci-lint || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.51.2
+	$(LOCALBIN)/golangci-lint run --fix
```
SeanMooney, line 135: we need to include this change

https://github.com/openshift/release/pull/36636/files

    # see tradeoff between speed and memory usage at https://golangci-lint.run/usage/performance/
    export GOGC=10

.pre-commit-config.yaml
```diff
@@ -60,4 +72,4 @@ repos:
   rev: 2.1.1
   hooks:
     - id: bashate
-      entry: bashate --error . --ignore=E006,E040,E043
+      entry: bashate --error . --ignore=E006,E040,E011,E020,E012
```
SeanMooney, line 75: we can fix it later but i would prefer to have a comment explaining what each of the ignored errors are.

Makefile
gibizer: I don't want to include it here. This target is run locally and locally I have enough ram not to get OOM killed. So I would like to keep the speed and only limit the memory usage when it runs from the prow job where memory is limited

.pre-commit-config.yaml
amolkahat: Updated.

gibizer (approved):

SeanMooney:
it failed on go tidy
lets see if a rebase will fix it

SeanMooney (changes requested):
go tidy is still failing


gotidy...................................................................Failed
- hook id: gotidy
- duration: 1.14s
- files were modified by this hook

amolkahat:
@SeanMooney Updated. Thank you.

gibizer:
please consider https://github.com/openstack-k8s-operators/nova-operator/pull/302 here as well

.pre-commit-config.yaml
```diff
@@ -2,20 +2,35 @@ repos:
 - repo: https://github.com/dnephin/pre-commit-golang
   rev: v0.5.1
   hooks:
-    - id: go-fmt
```
amolkahat, line 5: Removed `go-fmt go-vet and go-mod-tidy` which already added in local repo.

SeanMooney:
in general i would prefer to reference the ci logs then local executions
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/269/pull-ci-openstack-k8s-operators-nova-operator-master-precommit-check/1640269815828451328/artifacts/test/build-log.txt
but it also passes there so i think this is good ill review it quickly

.pre-commit-config.yaml
```diff
@@ -2,20 +2,35 @@ repos:
...
-      exclude: ^vendor
-    - id: go-vet
-    - id: go-mod-tidy
     - id: go-lint
```
SeanMooney, line 5: if the intent is to remove the use of this external repo then this should also be ported.
Note it is pinned by the git tag so the deprecation does not actually affect our usage of the repo.

SeanMooney (changes requested):

.pre-commit-config.yaml
amolkahat: Ok. I think removing it would be better and use make to run lint will make seance.

gibizer (commented):
Looks OK to me

.pre-commit-config.yaml
```diff
@@ -62,4 +78,13 @@ repos:
...
+      entry: bashate --error . --ignore=E006,E040,E020,E012
+      # Run bashate check for all bash scripts
+      # Ignores the following rules:
+      # E006: Line longer than 79 columns (as many scripts use jinja
+      #       templating, this is very difficult)
+      # E040: Syntax error determined using `bash -n` (as many scripts
+      #       use jinja templating, this will often fail and the syntax
+      #       error will be discovered in execution anyway)
+      # E020: Function declaration not in format ^function name {$
+      # E012: here doc didn't end before EOF
```
SeanMooney, line 90: # E012: here doc didn't end before EOF

i dont really like skiping this check

Makefile
```diff
@@ -119,6 +119,19 @@ vet: gowork ## Run go vet against code.
...
+tidy: fmt
+	go mod tidy; \
+	pushd "$(LOCALBIN)/../api"; \
+	go mod tidy; \
+	popd
+
+.PHONY: golangci-lint
+golangci-lint:
+	# NOTE this will install golangci-lint in to local bin dir
+	test -s $(LOCALBIN)/golangci-lint || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.51.2
```
SeanMooney, line 132: the correct way to install it would be 
go install golang.org/x/lint/golint

but since we are working on removing this now i guess this is fine for now

SeanMooney (approved):

PR #271, opened 2023-02-23
Reset ready status during reconfiguration

gibizer:
If NovaAPI reaches the Ready True state and then reconfigured then the Ready condition is not reset. So if a reconfiguration step fails then the Status.Conditions inconsistently shows Ready True + a subcondition in False state.

To keep the Status.Conditions consistent the reconciler needs to reset the state of the Ready condition if any of the subcondition is not Ready.

- [x] NovaAPI
- [x] NovaConductor
- [x] NovaCell
- [x] NovaScheduler
- [x] Nova

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -107,10 +107,17 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happend during the current reconciliation.
 	defer func() {
-		// update the overall status condition if service is ready
+		// update the Ready condition based on the sub conditions
 		if allSubConditionIsTrue(instance.Status) {
 			instance.Status.Conditions.MarkTrue(
 				condition.ReadyCondition, condition.ReadyMessage)
+		} else {
+			// something is not ready so reset the Ready condition
+			instance.Status.Conditions.MarkUnknown(
+				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
+			// and recalculate it based on the state of the rest of the conditions
+			instance.Status.Conditions.Set(
+				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
 		err := h.PatchInstance(ctx, instance)
 		if err != nil {
@@ -320,7 +327,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// DB or MQ is not yet created and those which needs API DB access but
 	// cell0 is not ready yet
 	failedCells := []string{}
-	creatingCells := []string{}
+	notReadyCells := []string{}
 	skippedCells := []string{}
 	readyCells := []string{}
 	cells := map[string]*novav1.NovaCell{}
@@ -366,7 +373,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		if err != nil {
 			failedCells = append(failedCells, fmt.Sprintf("%s(%v)", cellName, err.Error()))
 		} else if !cell.IsReady() {
-			creatingCells = append(creatingCells, cellName)
+			notReadyCells = append(notReadyCells, cellName)
 		} else {
 			readyCells = append(readyCells, cellName)
 		}
@@ -375,7 +382,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	}
 	util.LogForObject(
 		h, "Cell statuses", instance, "failed", failedCells,
-		"creating", creatingCells, "waiting", skippedCells,
+		"not ready", notReadyCells, "waiting", skippedCells,
 		"ready", readyCells, "all cells ready", allCellsReady)
 	if len(failedCells) > 0 {
 		instance.Status.Conditions.Set(condition.FalseCondition(
@@ -385,13 +392,13 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			novav1.NovaAllCellsReadyErrorMessage,
 			strings.Join(failedCells, ","),
 		))
-	} else if len(creatingCells) > 0 {
+	} else if len(notReadyCells) > 0 {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			novav1.NovaAllCellsReadyCondition,
 			condition.ErrorReason,
 			condition.SeverityError,
-			novav1.NovaAllCellsReadyCreatingMessage,
-			strings.Join(creatingCells, ","),
+			novav1.NovaAllCellsReadyNotReadyMessage,
+			strings.Join(notReadyCells, ","),
 		))
 	} else if len(skippedCells) > 0 {
 		instance.Status.Conditions.Set(condition.FalseCondition(
--- a/controllers/novaapi_controller.go
+++ b/controllers/novaapi_controller.go
@@ -113,10 +113,17 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happend during the current reconciliation.
 	defer func() {
-		// update the overall status condition if service is ready
+		// update the Ready condition based on the sub conditions
 		if allSubConditionIsTrue(instance.Status) {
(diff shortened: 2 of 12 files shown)


controllers/novaapi_controller.go
```diff
@@ -113,10 +113,17 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
...
 				condition.ReadyCondition, condition.ReadyMessage)
+		} else {
+			// something is not ready so reset the Ready condition
+			instance.Status.Conditions.MarkUnknown(
+				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
+			// and recalculate it based on the state of the rest of the conditions
+			instance.Status.Conditions.Set(
+				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
 		err := h.PatchInstance(ctx, instance)
```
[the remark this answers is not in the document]
bogdando: please disregard. If update fails, then no conditions changes land neither and everything remains consistent

controllers/nova_controller.go
```diff
@@ -107,10 +107,17 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
 		if allSubConditionIsTrue(instance.Status) {
 			instance.Status.Conditions.MarkTrue(
 				condition.ReadyCondition, condition.ReadyMessage)
+		} else {
+			// something is not ready so reset the Ready condition
+			instance.Status.Conditions.MarkUnknown(
+				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
+			// and recalculate it based on the state of the rest of the conditions
+			instance.Status.Conditions.Set(
+				instance.Status.Conditions.Mirror(condition.ReadyCondition))
```
stuggi, line 120: I like using Mirror() for this, but I think there is one issue. the order done by groupOrder() used in Mirror, is False, True, Unknown. With this during the initial service setup when initConditions sets all sub conditions to unknown, we could end up the ReadyCondition to become True when e.g. the first sub condition gets True and all others are still Unknown, right? 

For this we should extend Mirror() with two parameters to:
* exclude a condition status, e.g. the conditions which are True
* allow passing a condition Type list to exclude conditions from being Mirrored, where not all sub conditions are considered for the overall state

gibizer: Thanks for pointing that out. You are right that with the current Mirror implementation the Unknown + True means True. But we don't want to set Ready True in that case here.

I'm curious why the original implementation does the ordering False, True, Unknown? I guess the original use case of Mirror is to reflect the overall status of resource B in an single condition in resource A (that depends on B). So in this use case the behavior of Unknown + True -> True might cause the same confusion. E.g. If Nova depends KeystoneService then Nova can mirror the KeystoneService conditions into a keystone specific condition in the Nova status. But when the KeystoneService conditions are all Unknown with a single True sub condition, then the result of the mirroring is that Nova things the KeystoneService is ready. Or do I misunderstood the original use case of Mirror (and used it wrongly in nova alltogether)?

gibizer: After discussing this with @stuggi over slack we ended up agreeing to change the Mirror() behavior in lib-common. See the fix there https://github.com/openstack-k8s-operators/lib-common/pull/204 and the adaptation PR in nova as #283

bogdando: As this looks so much confusing, would it simplify work with conditions, if we go with alternative frameworks that represent conditions as dictionaries instead of lists?

gibizer: I think the representation of the Conditions type only affects the shape of the implementation of Mirror (and other conditions functions) but not the actual usecase which is about mirroring the most sever, and newest status condition. I believe if conditions would be a map such mirroring need would not change.

gibizer: This issue is resolved in lib-common and nova is being adapted to that new lib-common here https://github.com/openstack-k8s-operators/nova-operator/pull/283

SeanMooney (approved):

test/functional/nova_reconfiguration_test.go
```diff
@@ -220,4 +220,57 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
+			}, timeout, interval).Should(Succeed())
+
+			th.ExpectConditionWithDetails(
+				novaName,
+				ConditionGetterFunc(NovaConditionGetter),
+				condition.ReadyCondition,
+				corev1.ConditionFalse,
+				condition.ErrorReason,
+				"NovaCell cell1 is not Ready",
```
SeanMooney, line 246: we really should not have got that far.

if the network attachment does not exist we should have failed on our input validation before ever creating the cell.
it should not really be failing at the conductor stage.

so this feels like a bug to me.

gibizer: The network attachment that is missing is for the NovaConductor in cell1. 
So the status condition hierarchy will look like:
1.  NovaConductor/nova-cell1-conductor will have NetworkAttachmentsReadyCondition=False (and ReadyCondition=False)
2. NovaCell/nova-cell1 will have NovaConductorReadyCondition=False (and ReadyCondition=False) because the conductor is not ready. The message of this condition will still says "NetworkAttachment resources missing: internalapi"
3. Nova/nova will have NovaAllCellsReadyCondition=False with a message "NovaCell cell1 is not Ready" and that is also mirrored in its ReadyCondition=False condition.

During the mirroring of a condition if we map a single resource status to a single condition of a higher level resource, like in the case when the NovaConductor status is mirrored in the NovaConductorReadyCondition condition of the NovaCell we can copy over the message easily.

But when we represent the status of a set of resource as a single condition, like in the NovaAllCellsReadyCondition where we summarize the status of every cell, including the messages from multiple condition to a single message can result in confusing messages. So instead I implemented the message there to only contain the list of resource names that affected the False condition. Hence the message on Nova level will only mention which cell(s) causing the issue but does not mention the issue specifically. The code for this is here https://github.com/openstack-k8s-operators/nova-operator/blob/8cb955ea2a42f4a5fe67ad8fc7a4d4773f6f8413/controllers/nova_controller.go#L361-L386 I think this is a good compromise to avoid long and confusing condition messages in a multi cell deployment where multiple cells are failing for different reasons.

gibizer: I've added extra assert here now to show each step of the condition propagation in the system.

SeanMooney: i was expecting the nova controller to assert that the attachment exist before calling ensureCell. the cell controller also needs to assert what it does today. that is all correct I'm just saying i was expecting extra validation in the nova_contoler to assert the attachment it was passing to the nova_cell cr was valid before it created it.

SeanMooney: I was not suggesting moving it. I was asserting we should assert it in both places.
anyway, this is not really related to this patch so we can revisit this at another time.

my general perspective on input validation is if we are given a reference to a resource and it does not exist we should report that as early as possible.

gibizer: > my general perspective on input validation is if we are given a reference to a resource and it does not exist we should report that as early as possible.

Yeah that make sense. I even wondering if there is somewhere a resource reference field type in the framework we can reuse to express in the type of the field that what type of reference we expect.

gibizer: bottom line. I can agree that if we have a resource reference field then we should reject the API request if the referenced resource not exists. We can do that as part of the validation webhook work.

gibizer: I looked into having a typed resource reference instead of a single name as a string. 
* There is [v1.ObjectReference](https://github.com/kubernetes/api/blob/19ee583d5657970cfaa63ab11416a0ef5b3679bc/core/v1/types.go#L5902-L5920) but the usage of that is heavily discouraged. 
* We could also say something like
  ```
  ConfigMapRef *metav1.ObjectMeta `json:"configMapRef,omitempty"`
  ```
  But I guess it has very similar unwanted effects than the `v1.ObjectReference`
* What the document at `v1.ObjectReference` suggest is to define its own reference type that is focused to our needs. So I thought about this what we need today is just the name as we always assume the namespace is the same as the referring object's namespace and the kind is hardcoded. 

So I'm not closer to a solution :). I will get back to it once we reach the point where we add the validation webhook

PR #275, opened 2023-02-28
Added logic into NovaMetadata Reconciler

mrKisaoLamb:
Added config templates for novametadata
Modified reconciler
Added tests for NovaMetadata


controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
+	}
+
+	// Always update the instance status when exiting this function so we can
+	// persist any changes happend during the current reconciliation.
+	defer func() {
+		// update the overall status condition if service is ready
+		if allSubConditionIsTrue(instance.Status) {
+			instance.Status.Conditions.MarkTrue(
+				condition.ReadyCondition, condition.ReadyMessage)
+		}
```
gibizer, line 105: I suggest to add the else branch here to handle the case when the Ready condition needs to be set from True to False.  See  https://github.com/openstack-k8s-operators/nova-operator/pull/271

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
+func (r *NovaMetadataReconciler) reconcileDelete(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1beta1.NovaMetadata,
+) error {
+	util.LogForObject(h, "Reconciling delete", instance)
+
+	// Successfully cleaned up everyting. So as the final step let's remove the
+	// finalizer from ourselves to allow the deletion of NovaMetadata CR itself
+	updated := controllerutil.RemoveFinalizer(instance, h.GetFinalizer())
```
gibizer, line 415: As far as I see finalizer is not added to NovaMetadata above. I think that is OK not to be added as there is no need for any manual cleanup. So we don't need the finalizer removal either. In the future we will want to cleanup the service rows in the nova DB when the service is scaled in or deleted. So I'm OK to keep reconcileDelete. Also it is up to you to either remove the `RemoveFinalizer` call from here, or add an `AddFinalizer` call to the Reconcile logic.

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
 
+func (r *NovaMetadataReconciler) initStatus(
+	ctx context.Context, h *helper.Helper, instance *novav1beta1.NovaMetadata,
+) error {
+	if err := r.initConditions(ctx, h, instance); err != nil {
+		return err
+	}
+
+	if instance.Status.Hash == nil {
+		instance.Status.Hash = map[string]string{}
```
gibizer, line 180: you probably want to initialize the NetworkAttachements and the ApiEndpoint Status fields as well here.

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
+	util.LogForObject(h, "Reconciled delete successfully", instance)
+	return nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *NovaMetadataReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&novav1beta1.NovaMetadata{}).
+		Owns(&v1.StatefulSet{}).
+		Owns(&corev1.ConfigMap{}).
```
gibizer, line 429: we probably want to own the Service and the Route as we create those as well via lib-common
```
		Owns(&corev1.Service{}).
		Owns(&routev1.Route{}).
```

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
+	secret := &corev1.Secret{}
+	namespace := instance.GetNamespace()
+	secretName := types.NamespacedName{
+		Namespace: namespace,
+		Name:      instance.Spec.Secret,
+	}
+	err := h.GetClient().Get(ctx, secretName, secret)
+	if err != nil {
+		return err
+	}
```
gibizer, line 249: We probably want to change ensureSecret to return not just the hash of the Secret fields we depend on but the field value, or the whole Secret we depend on, so that here you don't have to read the Secret from the API again and do the error handling again. But this probably need a separate PR anyhow. So this is OK as is now.

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
...
+}
+
+func (r *NovaMetadataReconciler) ensureDeployment(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1beta1.NovaMetadata,
+	inputHash string,
+) (ctrl.Result, error) {
+	serviceLabels := map[string]string{
+		common.AppSelector: NovaConductorLabelPrefix,
```
gibizer, line 309: This should be nova-metadata specific. You added `NovaMetadataLabelPrefix` already so you can use that here

gibizer (changes requested):
I only reviewed the code change and not the test code change yet. I have some comments inline but overall this is moving to the right direction I think.

api/v1beta1/novametadata_types.go
```diff
@@ -92,7 +92,7 @@ type NovaMetadataSpec struct {
 	// keystone
 	ServiceUser string `json:"serviceUser"`
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
mrKisaoLamb: added todo for adding checks

api/v1beta1/novametadata_types.go
```diff
@@ -120,6 +124,12 @@ type NovaMetadataSpec struct {
 	// TODO(gibi): add webhook to validate this CellName constraint
 	CellDatabaseHostname string `json:"cellDatabaseHostname"`
 
+	// +kubebuilder:validation:Required
+	// APIMessageBusSecretName - the name of the Secret conntaining the
+	// transport URL information to use when accessing the API message
+	// bus.
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 6a0883d95be43dcb7d38b7ab110c177ca5862e24

config/samples/nova_v1beta1_nova-multi-cell.yaml
```diff
@@ -76,8 +76,7 @@ spec:
     defaultConfigOverwrite:
       logging.conf: |
         # my custom logging configuration
-    # TODO: what is the name of the container for metadata?
-    containerImage: quay.io/tripleozedcentos9/openstack-nova-metadata-api:current-tripleo
+    containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
```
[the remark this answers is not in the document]
mrKisaoLamb: Yes, in tripleo nova-apui image is run also httpd.conf is one for both services

api/v1beta1/novametadata_types.go
```diff
@@ -63,6 +63,10 @@ type NovaMetadataTemplate struct {
 	// +kubebuilder:validation:Optional
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
 	NetworkAttachments []string `json:"networkAttachments"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints"`
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 6a0883d95be43dcb7d38b7ab110c177ca5862e24

api/v1beta1/novametadata_types.go
```diff
@@ -141,6 +151,10 @@ type NovaMetadataSpec struct {
 	// +kubebuilder:validation:Required
 	// NovaServiceBase specifies the generic fields of the service
 	NovaServiceBase `json:",inline"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints"`
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 6a0883d95be43dcb7d38b7ab110c177ca5862e24

config/crd/bases/nova.openstack.org_novametadata.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 6a0883d95be43dcb7d38b7ab110c177ca5862e24

config/samples/nova_v1beta1_novametadata.yaml
```diff
@@ -3,4 +3,9 @@ kind: NovaMetadata
 metadata:
   name: novametadata-sample
 spec:
-  # TODO(user): Add fields here
+  apiDatabaseHostname: openstack
+  apiMessageBusSecretName: rabbitmq-transport-url-nova-api-transport
+  cell0DatabaseHostname: openstack
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 6a0883d95be43dcb7d38b7ab110c177ca5862e24

controllers/novametadata_controller.go
```diff
@@ -43,17 +57,375 @@ type NovaMetadataReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/reconcile
-func (r *NovaMetadataReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+func (r *NovaMetadataReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
```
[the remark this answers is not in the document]
mrKisaoLamb: done e9fa08c196082d72965fa8541ece713fbd158bc7

controllers/novametadata_controller.go
mrKisaoLamb: done, e9fa08c196082d72965fa8541ece713fbd158bc7

controllers/novametadata_controller.go
mrKisaoLamb: done e9fa08c196082d72965fa8541ece713fbd158bc7

controllers/novametadata_controller.go
mrKisaoLamb: done e9fa08c196082d72965fa8541ece713fbd158bc7

gibizer:
I deployed this PR by building it into a fresh openstack-operator (quay.io/gibi/openstack-operator-index:v0.0.62) then did `make openstack` and `make openstack_deploy` I see that we this PR still needs a bit of code in the nova_controller to automatically create NovaMetadata CRs based on the data in the Nova CR. So I went ahead and created the metadata CR manually with `oc apply -f ../nova-operator/config/samples/nova_v1beta1_novametadata.yaml`. The Statefulset was created but the pod goes to crashloop probably due to failed probes. Unfortunately the only log I see is in the logs is the repeated `2023-03-09 12:20:24.085 12 INFO nova.api.openstack.wsgi_app [None req-3f7608ab-434b-43c2-8b10-3d9276b8edf3 - - - - - -] Global data already initialized, not re-initializing.`
Then I tried manually curling the the url and got a standard error from the httpd:
```
$ oc rsh -c novametadata-sample-metadata  pod/novametadata-sample-0 curl http://localhost:8775/
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 [no address given] to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
</body></html>
```

Then I realized that we have two different containers in the pod and checked the stdout of the httpd one:

```
$ oc logs -f -c novametadata-sample-metadata pod/novametadata-sample-0
INFO:__main__:Loading config file at /var/lib/openstack/config/nova-metadata-config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Copying service configuration files
INFO:__main__:Deleting /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/01-nova.conf to /etc/nova/nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf
INFO:__main__:Deleting /etc/httpd/conf/httpd.conf
INFO:__main__:Copying /var/lib/openstack/config/httpd.conf to /etc/httpd/conf/httpd.conf
INFO:__main__:Setting permission for /etc/httpd/conf/httpd.conf
INFO:__main__:Writing out command to execute
INFO:__main__:Setting permission for /var/log/nova
+ sudo -E kolla_set_configs
sudo: unable to send audit message: Operation not permitted
INFO:__main__:Loading config file at /var/lib/openstack/config/nova-metadata-config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Copying service configuration files
INFO:__main__:Deleting /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/01-nova.conf to /etc/nova/nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf
INFO:__main__:Deleting /etc/httpd/conf/httpd.conf
INFO:__main__:Copying /var/lib/openstack/config/httpd.conf to /etc/httpd/conf/httpd.conf
INFO:__main__:Setting permission for /etc/httpd/conf/httpd.conf
INFO:__main__:Writing out command to execute
INFO:__main__:Setting permission for /var/log/nova
++ cat /run_command
+ CMD='/usr/sbin/httpd -DFOREGROUND'
+ ARGS=
+ [[ ! -n '' ]]
+ . kolla_extend_start
+ echo 'Running command: '\''/usr/sbin/httpd -DFOREGROUND'\'''
Running command: '/usr/sbin/httpd -DFOREGROUND'
+ umask 0022
+ exec /usr/sbin/httpd -DFOREGROUND
mod_wsgi (pid=12): Attach interpreter ''.
mod_wsgi (pid=12): Imported 'mod_wsgi'.
mod_wsgi (pid=12): Started thread 0 in daemon process 'nova-metadata'.
AH01628: authorization result: granted (no directives)
AH01628: authorization result: granted (no directives)
mod_wsgi (pid=12, process='nova-metadata', application=''): Loading Python script file '/usr/bin/nova-metadata-wsgi'.
mod_wsgi (pid=12): Callback registration for signal 12 ignored.
  File "/usr/bin/nova-metadata-wsgi", line 52, in <module>
    application = init_application()
  File "/usr/lib/python3.9/site-packages/nova/api/metadata/wsgi.py", line 20, in init_application
    return wsgi_app.init_application(NAME)
  File "/usr/lib/python3.9/site-packages/nova/api/openstack/wsgi_app.py", line 128, in init_application
    init_global_data(conf_files, name)
  File "/usr/lib/python3.9/site-packages/nova/utils.py", line 1123, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/nova/api/openstack/wsgi_app.py", line 105, in init_global_data
    gmr.TextGuruMeditation.setup_autorun(
  File "/usr/lib/python3.9/site-packages/oslo_reports/guru_meditation_report.py", line 152, in setup_autorun
    cls._setup_signal(signal.SIGUSR2,
  File "/usr/lib/python3.9/site-packages/oslo_reports/guru_meditation_report.py", line 185, in _setup_signal
    signal.signal(signum,
mod_wsgi (pid=12): Failed to exec Python script file '/usr/bin/nova-metadata-wsgi'.
mod_wsgi (pid=12): Exception occurred processing WSGI script '/usr/bin/nova-metadata-wsgi'.
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2716, in _do_get
    return (convert(val), alt_loc)
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2685, in convert
    return self._convert_value(
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2813, in _convert_value
    return opt.type(value)
  File "/usr/lib/python3.9/site-packages/oslo_config/types.py", line 315, in __call__
    raise ValueError('Should be greater than or equal to %g' %
ValueError: Should be greater than or equal to 1
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/bin/nova-metadata-wsgi", line 52, in <module>
    application = init_application()
  File "/usr/lib/python3.9/site-packages/nova/api/metadata/wsgi.py", line 20, in init_application
    return wsgi_app.init_application(NAME)
  File "/usr/lib/python3.9/site-packages/nova/api/openstack/wsgi_app.py", line 128, in init_application
    init_global_data(conf_files, name)
  File "/usr/lib/python3.9/site-packages/nova/utils.py", line 1123, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/nova/api/openstack/wsgi_app.py", line 114, in init_global_data
    CONF.log_opt_values(
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2603, in log_opt_values
    _sanitize(opt, getattr(self, opt_name)))
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2219, in __getattr__
    return self._get(name)
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2653, in _get
    value, loc = self._do_get(name, group, namespace)
  File "/usr/lib/python3.9/site-packages/oslo_config/cfg.py", line 2732, in _do_get
    raise ConfigFileValueError(message)
oslo_config.cfg.ConfigFileValueError: Value for option osapi_compute_workers from LocationInfo(location=<Locations.user: (4, True)>, detail='/etc/nova/nova.conf') is not valid: Should be greater than or equal to 1
```

Which is strange. I thought the we can simply start a metadata API without any `osapi_compute_workers` :/

pkg/novametadata/deployment.go
```diff
@@ -0,0 +1,181 @@
...
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: nova.ServiceAccount,
+					Volumes: nova.GetOpenstackVolumes(
+						nova.GetServiceConfigConfigMapName(instance.Name),
+					),
+					Containers: []corev1.Container{
+						// the first container in a pod is the default selected
+						// by oc log so define the log stream container first.
+						{
```
SeanMooney, line 129: ah good you already have done this

templates/nova.conf
```diff
@@ -36,10 +36,10 @@ force_config_drive=True
 {{if eq .service_name "nova-api"}}
 # scaling should be done by running more pods
 osapi_compute_workers=1
-metadata_workers=1
+metadata_workers=0
 {{else if eq .service_name "nova-metadata"}}
```
SeanMooney, line 40: yep this was something i had ment to fix for a long time

templates/nova.conf
```diff
@@ -36,10 +36,10 @@ force_config_drive=True
...
 # scaling should be done by running more pods
-osapi_compute_workers=1
+osapi_compute_workers=0
 metadata_workers=1
 {{end}}
```
SeanMooney, line 44: and i was oging to do this too but I'm not sure if osapi_compute_workers=0 is strictly required when you use the metatadata wsgi file

i prefer to include it to be explicit in this case however.

config/samples/nova_v1beta1_nova-multi-cell.yaml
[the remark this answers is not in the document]
SeanMooney: it technailly does not mattter as both are installed in it but its kind incorrect.

it does not actually matter however as long as we use the correct wsgi entrypoitn application in the apache cofnig

templates/novametadata/config/httpd.conf
```diff
@@ -0,0 +1,52 @@
...
+  CustomLog /dev/stdout proxy env=forwarded
+  ## set nova vhost log level to debug
+  LogLevel debug
+
+  ## WSGI configuration
+  WSGIProcessGroup nova-metadata
+  WSGIApplicationGroup %{GLOBAL}
+  WSGIPassAuthorization On
+  WSGIDaemonProcess nova-metadata processes=1 threads=1 user=nova group=nova display-name=nova-metadata-api
+  WSGIScriptAlias / /usr/bin/nova-metadata-wsgi
```
SeanMooney, line 42: given this is using the metadta only wsgi entrypoint the fact its a single API image is fine.

SeanMooney (commented):
as a quick review this mostly looks correct but i have not tested it or reviewed this fully.

the broad strokes all look correct however.

api/v1beta1/novametadata_types.go
```diff
@@ -92,9 +97,10 @@ type NovaMetadataSpec struct {
 	// keystone
 	ServiceUser string `json:"serviceUser"`
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
 	// KeystoneAuthURL - the URL that the nova-metadata service can use to talk
 	// to keystone
+	// TODO(ksambor) Add checking if dynamic vendor data is configured
```
gibizer, line 103: the APIDatabaseUser needs to be defaulted to `nova_api` (sorry I cannot comment directly that line as github is :hankey:

templates/nova.conf
gibizer: it is strange that we cannot deploy metadata without osapi_compute_workers

config/samples/nova_v1beta1_novametadata.yaml
```diff
@@ -3,4 +3,8 @@ kind: NovaMetadata
 metadata:
   name: novametadata-sample
 spec:
-  # TODO(user): Add fields here
+  apiDatabaseHostname: openstack
+  apiMessageBusSecretName: rabbitmq-transport-url-nova-api-transport
+  containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
+  keystoneAuthURL: http://keystone-public-openstack.apps-crc.testing
+  secret: osp-secret
```
gibizer, line 10: I actually needed the following sample to get the service to work

```yaml
kind: NovaMetadata
metadata:
  name: novametadata-sample
spec:
  apiDatabaseHostname: openstack
  apiMessageBusSecretName: rabbitmq-transport-url-nova-api-transport
  cellDatabaseHostname: openstack
  cellDatabaseUser: nova_cell0
  containerImage: quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo
  keystoneAuthURL: http://keystone-public-openstack.apps-crc.testing
  secret: osp-secret
  apiDatabaseUser: nova_api
```

gibizer (changes requested):
with a tweaked sample I was able to start the metadata service and it actually reponds to requests:
```
$ oc rsh -c novametadata-sample-metadata  pod/novametadata-sample-0 curl http://localhost:8775/
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
```

We still need code to create the NovaMetadata from Nova but at least the manual create works now. Good progress!

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+		"cell_db_user":           instance.Spec.CellDatabaseUser,
+		"cell_db_password":       string(secret.Data[instance.Spec.PasswordSelectors.CellDatabase]),
+		"cell_db_address":        instance.Spec.CellDatabaseHostname,
+		"cell_db_port":           3306,
+		"openstack_cacert":       "",          // fixme
+		"openstack_region_name":  "regionOne", // fixme
+		"default_project_domain": "Default",   // fixme
+		"default_user_domain":    "Default",   // fixme
+		"metadata_secret":        "42",        // fixme
+		"log_file":               "/var/log/nova/nova-metadata.log",
```
SeanMooney, line 319: oh this is here.
thats kind of bizar it should really be logging to the file

gibizer: it is definitely logging to that file. Like the oslo debug logs are there. But the error stack traces are still end up in the stdout of the httpd container.

api/v1beta1/novametadata_types.go
mrKisaoLamb: Done

api/v1beta1/novametadata_types.go
```diff
@@ -120,6 +124,12 @@ type NovaMetadataSpec struct {
 	// TODO(gibi): add webhook to validate this CellName constraint
 	CellDatabaseHostname string `json:"cellDatabaseHostname"`
 
+	// +kubebuilder:validation:Required
+	// APIMessageBusSecretName - the name of the Secret conntaining the
+	// transport URL information to use when accessing the API message
+	// bus.
+	APIMessageBusSecretName string `json:"apiMessageBusSecretName"`
+
```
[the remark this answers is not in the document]
mrKisaoLamb: Done

config/samples/nova_v1beta1_novametadata.yaml
mrKisaoLamb: Done, file now is updated plus same set of values is used in NovaController

mrKisaoLamb (commented):
Thanks for reviews and pointing out issues and fixes!
I retest with updated Nova cr and now nova-metadata was spawned and api work as expected
`$ oc rsh   pod/nova-metadata-0 curl http://localhost:8775/
Defaulted container "nova-metadata-log" out of: nova-metadata-log, nova-metadata-metadata
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
`

pkg/novametadata/deployment.go
```diff
@@ -0,0 +1,181 @@
...
+	instance *novav1.NovaMetadata,
+	configHash string,
+	labels map[string]string,
+	annotations map[string]string,
+) *appsv1.StatefulSet {
+	runAsUser := int64(0)
+
+	// This allows the pod to start up slowly. The pod will only be killed
+	// if it does not succeed a probe in 60 seconds.
+	startupProbe := &corev1.Probe{
```
bogdando, line 43: @gibizer @SeanMooney so the status que, we keep hardcoding all probes. But what do you see for a longer term interface, as we earlier discussed, to override probes' timeouts (not the commands) in pods specs?

SeanMooney: ya i dont really want to espose this as something the end user can tweak unless we have no other choice.
i consider this to be an implementation detail not part of our API.

I'm reluctant to expsot the limits/requests as it is  

i would like use to set reasonable values and as gibi says relaxt the for longer intervals if needed.

mrKisaoLamb: I also prefer not to expose timeouts, as debugging client instances can be painful when only a portion of dependent intervals is changed. I favor opinionated values in this case, but perhaps we can establish a constant across all StatefulSet deployments within the operator.

controllers/novametadata_controller.go
mrKisaoLamb: I will do follow up then

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+		instance, labels.GetGroupLabel(NovaMetadataLabelPrefix), map[string]string{},
+	)
+
+	err = r.GenerateConfigs(
+		ctx, h, instance, hashes, templateParameters, extraData, cmLabels, map[string]string{},
+	)
+	return err
+}
+
+func (r *NovaMetadataReconciler) ensureDeployment(
```
SeanMooney, line 340: this will always fail with the default container url as it is set to the wrong value

	// +kubebuilder:default="quay.io/tripleozedcentos9/openstack-nova-metadata:current-tripleo"
 we have  ^
 but in tripleo we use the nova-api container for the metadata container.

SeanMooney: i cant commend direclty on the tyrp because github sucks for code review

SeanMooney (changes requested):
ok so two issues.

first the container url is wrong but seting it to use the api container works
second setting replcias 0 never gets to the ready state.


the deployed container does appear to work 
sh-5.1# curl 127.0.0.1:8775
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
latestsh-5.1#

SeanMooney (changes requested):
ok so two issues.

first the container url is wrong but seting it to use the api container works
second setting replcias 0 never gets to the ready state.


the deployed container does appear to work 
sh-5.1# curl 127.0.0.1:8775
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
latestsh-5.1#

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+			condition.NetworkAttachmentsReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.NetworkAttachmentsReadyErrorMessage,
+			err.Error()))
+
+		return ctrl.Result{}, err
+	}
+
+	if instance.Status.ReadyCount > 0 {
```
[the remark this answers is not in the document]
mrKisaoLamb: Done, 33d1090a4ec2a87c2c4c6fa11631931aa8037d99

controllers/novametadata_controller.go
mrKisaoLamb: Nice catch! I appreciate it. I overlooked updating the default value and only changed the CR. Done 33d1090a4ec2a87c2c4c6fa11631931aa8037d99

karelyatin (commented):
shoudn't force_config_drive=True be dropped now from nova.conf and rely on nova defaults as metadata is enabled with this patch?

gibizer:
> shoudn't force_config_drive=True be dropped now from nova.conf and rely on nova defaults as metadata is enabled with this patch?

Does the metadata on the neutron side works together with this patch? If so, then we can drop force_config_drive=True. Unfortunately I had no time to test things together yet, so I hope other have some test results with this e2e. Or I can get back to this after we stabilized the boot-a-VM workflow for the demo.

karelyatin:
> > shoudn't force_config_drive=True be dropped now from nova.conf and rely on nova defaults as metadata is enabled with this patch?
> 
> Does the metadata on the neutron side works together with this patch? If so, then we can drop force_config_drive=True. Unfortunately I had no time to test things together yet, so I hope other have some test results with this e2e. Or I can get back to this after we stabilized the boot-a-VM workflow for the demo.

not tested yet i just realized it, let me test it, will report back.
But anyway config option can be dropped in follow up.

mrKisaoLamb:
AFAIK  we are missing a few configurations to make nova-metadata work with Neutron, such as metadata_proxy_shared_secret and nova_metadata_host. So in current state e2e test will not work

gibizer:
I did a dirty test of this PR. I deployed with openstack-operator master +  install_yamls (https://github.com/openstack-k8s-operators/install_yamls/pull/156) with edpm compute + network isolation. Booted an openstack VM there. Then I replaced the nova-operator with the version from this PR. Run make install to update the nova CRDs from this PR. Then I modified the OpenstackControlPlane CR to have NovaMetadata configured with the necessary parameters. The nova-metadata service was depoyed properly and locally (withing the pod) it servers the metadata response. Then I tried to access the metadata from the openstack VM I booted before but that is failed with:
```
$ curl -v http://169.254.169.254/openstack
* Failed to connect to 169.254.169.254 port 80: Connection timed out
```
so something is definitely missing. 

I've checked the ovn-metadata-agent pod on the compute hosting the VM.

```
+ sudo -E kolla_set_configs
INFO:__main__:Loading config file at /var/lib/kolla/config_files/config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Writing out command to execute
INFO:__main__:Setting permission for /var/log/neutron
INFO:__main__:Setting permission for /var/lib/neutron
INFO:__main__:Setting permission for /var/lib/neutron/.bash_logout
INFO:__main__:Setting permission for /var/lib/neutron/.bash_profile
INFO:__main__:Setting permission for /var/lib/neutron/.bashrc
++ cat /run_command
Running command: '/usr/bin/networking-ovn-metadata-agent --config-file /etc/neutron/neutron.conf --config-file /etc/neutron/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file=/var/log/neutron/ovn-metadata-agent.log'
+ CMD='/usr/bin/networking-ovn-metadata-agent --config-file /etc/neutron/neutron.conf --config-file /etc/neutron/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file=/var/log/neutron/ovn-metadata-agent.log'
+ ARGS=
+ [[ ! -n '' ]]
+ . kolla_extend_start
+ echo 'Running command: '\''/usr/bin/networking-ovn-metadata-agent --config-file /etc/neutron/neutron.conf --config-file /etc/neutron/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file=/var/log/neutron/ovn-metadata-agent.log'\'''
+ umask 0022
+ exec /usr/bin/networking-ovn-metadata-agent --config-file /etc/neutron/neutron.conf --config-file /etc/neutron/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file=/var/log/neutron/ovn-metadata-agent.log
Traceback (most recent call last):
  File "/usr/bin/networking-ovn-metadata-agent", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python3.9/site-packages/neutron/cmd/eventlet/agents/ovn_metadata.py", line 24, in main
    metadata_agent.main()
  File "/usr/lib/python3.9/site-packages/neutron/agent/ovn/metadata_agent.py", line 36, in main
    config.init(sys.argv[1:])
  File "/usr/lib/python3.9/site-packages/neutron/common/config.py", line 102, in init
    n_rpc.init(cfg.CONF)
  File "/usr/lib/python3.9/site-packages/neutron_lib/rpc.py", line 58, in init
    TRANSPORT = oslo_messaging.get_rpc_transport(
  File "/usr/lib/python3.9/site-packages/oslo_messaging/rpc/transport.py", line 47, in get_rpc_transport
    return msg_transport._get_transport(
  File "/usr/lib/python3.9/site-packages/oslo_messaging/transport.py", line 199, in _get_transport
    url = TransportURL.parse(conf, url)
  File "/usr/lib/python3.9/site-packages/oslo_messaging/transport.py", line 464, in parse
    raise InvalidTransportURL(url.geturl(), 'No scheme specified')
oslo_messaging.transport.InvalidTransportURL: No scheme specified
```

So I guess something is not properly configured regarding the ovn agent side of the metadata stack.

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+		return err
+	}
+
+	apiMessageBusSecret := &corev1.Secret{}
+	secretName = types.NamespacedName{
+		Namespace: instance.Namespace,
+		Name:      instance.Spec.APIMessageBusSecretName,
+	}
+	err = h.GetClient().Get(ctx, secretName, apiMessageBusSecret)
+	if err != nil {
```
SeanMooney, line 295: so since we have a hard dep here it means we currently only supprot deploying metadata globally not on a per cell level.

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+	return ctrl.Result{}, nil
+}
+
+func (r *NovaMetadataReconciler) ensureServiceExposed(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1beta1.NovaMetadata,
+) (ctrl.Result, error) {
+	var ports = map[endpoint.Endpoint]endpoint.Data{
+		endpoint.EndpointPublic:   {Port: novametadata.APIServicePort},
```
SeanMooney, line 422: we should not be exposing this a public endpoint
just internal.

templates/novametadata/config/nova-metadata-config.json
```diff
@@ -0,0 +1,24 @@
+{
+  "command": "/usr/sbin/httpd -DFOREGROUND",
+  "config_files": [
+    {
+      "source": "/var/lib/openstack/config/01-nova.conf",
+      "dest": "/etc/nova/nova.conf",
```
SeanMooney, line 6: this is not quite right

we shoudl be copying this to /etc/nova/nova.conf.d/01-nova.conf

and copying a blank file over etc/nova/nova.conf

like we do for the schduler
https://github.com/openstack-k8s-operators/nova-operator/blob/master/templates/novascheduler/config/nova-scheduler-config.json

we we added support for the api we had not backported https://github.com/openstack/nova/commit/73fe84fa0ea6f7c7fa55544f6bce5326d87743a6
to zed i guess this si ok for now and we can fix both the api and metadata api in a follow up

SeanMooney (changes requested):
i had written a nice summeray but vscod decied to drop it so short version.
-1 on exposing the route. that needs to be fixed befoer we merge this
this has no supprot for dpeloying the metadata api per cell but that can be in a followup.
we need to make the api and metadata api both supprot config override. can also be in a followup.

over all code looks ok and test coverage is ok. i did not look at that as closely as i shoudl so we might need to add some negitive tests but over all im mostly happy with this

mrKisaoLamb:
Thank you a lot for tests and revie! Done, public is removed plus now keystoneurl is required. As for the other changes you mentioned, agree that I can work on those in follow-up patches

controllers/novametadata_controller.go
```diff
@@ -18,11 +18,28 @@ package controllers
...
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	routev1 "github.com/openshift/api/route/v1"
+
+	common "github.com/openstack-k8s-operators/lib-common/modules/common"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
```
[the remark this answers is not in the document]
gibizer: I'm not sure what is the reasoning against `import "github.com/openstack-k8s-operators/lib-common/modules/common/condition"` I can see reasoning against the dot import format i.e. `import . "github.com/openstack-k8s-operators/lib-common/modules/common/condition"` as that could lead to name clashes. But `import "github.com/openstack-k8s-operators/lib-common/modules/common/condition"` automatically means the same as `condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"` so I think we are safe here.

controllers/novametadata_controller.go
```diff
@@ -43,17 +67,427 @@ type NovaMetadataReconciler struct {
...
+		}
+	}()
+
+	if !instance.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, r.reconcileDelete(ctx, h, instance)
+	}
+
+	hashes := make(map[string]env.Setter)
+
+	secretHash, result, err := ensureSecret(
```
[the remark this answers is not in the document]
gibizer: ensureSecret is in the `controller` package in common.go as well as this file, hence there is no way to prefix this.

templates/novametadata/config/nova-metadata-config.json
gibizer: > we we added support for the api we had not backported https://github.com/openstack/nova/commit/73fe84fa0ea6f7c7fa55544f6bce5326d87743a6 to zed i guess

We have the fix in zed: https://review.opendev.org/q/I8e3ccd75cbb7f2e132b403cb38022787c2c0a37b

controllers/novametadata_controller.go
gibizer: yepp this was an agreement with @mrKisaoLamb to reduce the scope of this PR

api/v1beta1/novametadata_types.go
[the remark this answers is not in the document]
gibizer: > im not sure we shoudl be takign the cell name as a pramter here this should just be defiend on the cell crd and we should be able to know the cell name form that.

We take the CellName as a parameter in the NovaMetadataSpec (but not in the Template)  (similarly how it is done in NovaConductor) so that the metadata (and the conductor) controller do not need to look up the parent NovaCell instance just to get name of the cell it is in. I think this is good.

test/functional/nova_metadata_controller_test.go
```diff
@@ -0,0 +1,513 @@
...
+						corev1.ConditionTrue,
+					)
+
+					DeleteInstance(GetNovaMetadata(novaMetadataName))
+
+					Eventually(func() []corev1.ConfigMap {
+						return th.ListConfigMaps(novaMetadataName.Name).Items
+					}, timeout, interval).Should(BeEmpty())
+				})
+			})
```
gibizer, line 208: I think this is now a duplicate of L210-L228

test/functional/nova_metadata_controller_test.go
```diff
@@ -0,0 +1,513 @@
...
+						corev1.ConditionFalse,
+						condition.RequestedReason,
+						condition.DeploymentReadyRunningMessage,
+					)
+					th.SimulateStatefulSetReplicaReady(statefulSetName)
+				})
+
+				It("reports that the StatefulSet is ready", func() {
+					ss := th.GetStatefulSet(statefulSetName)
+					_ = ss
```
gibizer, line 293: hm, why you need this?

gibizer (changes requested):
I have couple of requests in the tests. The code looks OK to me.

templates/novametadata/config/nova-metadata-config.json
SeanMooney: we do now yes not when the API support meged.

that way i said we can fix both in a follow up.

controllers/novametadata_controller.go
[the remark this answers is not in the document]
SeanMooney: i tought import "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
put all the module export in the current namespace unprefixed.

if you do 

condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"

then you refer to it via condition.

perhaps the ensureSecret function is because its the same package?

templates/novametadata/config/nova-metadata-config.json
gibizer: let's fix it in a follow up

test/functional/nova_metadata_controller_test.go
```diff
@@ -0,0 +1,513 @@
...
+			})
+
+			It("exposes the service", func() {
+				th.SimulateStatefulSetReplicaReady(statefulSetName)
+				th.ExpectCondition(
+					novaMetadataName,
+					ConditionGetterFunc(NovaMetadataConditionGetter),
+					condition.ExposeServiceReadyCondition,
+					corev1.ConditionTrue,
+				)
```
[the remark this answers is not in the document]
mrKisaoLamb: Done, 9a18b72af3e81aa657fde20a43e0ada89d9f7c3c

test/functional/nova_metadata_controller_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+			th.ExpectCondition(
+				novaMetadataName,
+				ConditionGetterFunc(NovaMetadataConditionGetter),
+				condition.ReadyCondition,
+				corev1.ConditionTrue,
+			)
+		})
+	})
+})
```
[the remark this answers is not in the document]
mrKisaoLamb: done, 9a18b72af3e81aa657fde20a43e0ada89d9f7c3c

test/functional/nova_metadata_controller_test.go
mrKisaoLamb: done, 9a18b72af3e81aa657fde20a43e0ada89d9f7c3c

test/functional/nova_metadata_controller_test.go
mrKisaoLamb: I don't, it is leftover from some debuging, removed 9a18b72af3e81aa657fde20a43e0ada89d9f7c3c

test/functional/nova_controller_test.go
```diff
@@ -358,6 +370,40 @@ var _ = Describe("Nova controller", func() {
...
+				corev1.ConditionTrue,
+			)
+			th.ExpectCondition(
+				novaName,
+				ConditionGetterFunc(NovaConditionGetter),
+				condition.ReadyCondition,
+				corev1.ConditionTrue,
+			)
+		})
 	})
```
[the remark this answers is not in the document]
mrKisaoLamb: done 9a18b72af3e81aa657fde20a43e0ada89d9f7c3c

karelyatin:
> So I guess something is not properly configured regarding the ovn agent side of the metadata stack.
For this issue in edpm deploy pushed https://github.com/openstack-k8s-operators/install_yamls/pull/179 , actual settings can be done once nova-operator, openstack-operator have all required fixes.

api/v1beta1/novametadata_types.go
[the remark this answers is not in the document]
gibizer: yeah, you have to create unique service names so we need some kind of identity passed to the NovaMetadata. We can probably rely on the NovaMetadata.Name if we want. Let's discuss this when we plan the NovaMetadata per cell feature implementation.

PR #278, opened 2023-03-03
Extend the timeout for golangci-lint

gibizer:
The default timeout is 1 minutes that seems to be not enough in CI. This is probably a fallout from the change where we limited the memory usage of golangci-lint in CI as that is expected to decrease the speed of the lint.


SeanMooney (approved):
as with the placement patch this makes sense

PR #283, opened 2023-03-06
Adapt to Conditions.Mirror() change

gibizer:
The lib-common PR https://github.com/openstack-k8s-operators/lib-common/pull/204 fixed the behavior of Conditions.Mirror(). So it now returns any Unknown condition before any True condition (except ReadyCondition True).

It seems that our tests incorrectly assumed that NovaAPI becomes Ready after the StatefulSet is Ready. But in reality NovaAPI creates the KeystoneEndpoint *after* the StatefulSet is ready. So these tests passed only because of the original bug in the Mirror code returned a True condition from NovaAPI when it had DeploymentReady=True and KeystoneEndpointReady=Unknown status. As the bug is fixed now the test needs to make sure the KeystoneEndpoints are ready before it can expect that NovaAPI is ready.


gibizer:
https://github.com/openstack-k8s-operators/lib-common/pull/204 is merged so this is ready to go

SeanMooney (approved):

PR #284, opened 2023-03-07
Refactor tests help function

mrKisaoLamb:
Refactor tests helping function and remove redundunt code

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
@@ -48,12 +49,13 @@ const (
 	consistencyTimeout = timeout
 )
 
-func CreateUnstructured(rawObj map[string]interface{}) {
+func CreateUnstructured(rawObj map[string]interface{}) *unstructured.Unstructured {
 	logger.Info("Creating", "raw", rawObj)
 	unstructuredObj := &unstructured.Unstructured{Object: rawObj}
 	_, err := controllerutil.CreateOrPatch(
 		ctx, k8sClient, unstructuredObj, func() error { return nil })
 	Expect(err).ShouldNot(HaveOccurred())
+	return unstructuredObj
 }
 
 func GetDefaultNovaAPISpec() map[string]interface{} {
@@ -67,7 +69,7 @@ func GetDefaultNovaAPISpec() map[string]interface{} {
 	}
 }
 
-func CreateNovaAPI(namespace string, spec map[string]interface{}) types.NamespacedName {
+func CreateNovaAPI(namespace string, spec map[string]interface{}) client.Object {
 	novaAPIName := uuid.New().String()
 
 	raw := map[string]interface{}{
@@ -79,27 +81,8 @@ func CreateNovaAPI(namespace string, spec map[string]interface{}) types.Namespac
 		},
 		"spec": spec,
 	}
-	CreateUnstructured(raw)
+	return CreateUnstructured(raw)
 
-	return types.NamespacedName{Name: novaAPIName, Namespace: namespace}
-}
-
-func DeleteNovaAPI(name types.NamespacedName) {
-	// We have to wait for the controller to fully delete the instance
-	Eventually(func(g Gomega) {
-		novaAPI := &novav1.NovaAPI{}
-		err := k8sClient.Get(ctx, name, novaAPI)
-		// if it is already gone that is OK
-		if k8s_errors.IsNotFound(err) {
-			return
-		}
-		g.Expect(err).Should(BeNil())
-
-		g.Expect(k8sClient.Delete(ctx, novaAPI)).Should(Succeed())
-
-		err = k8sClient.Get(ctx, name, novaAPI)
-		g.Expect(k8s_errors.IsNotFound(err)).To(BeTrue())
-	}, timeout, interval).Should(Succeed())
 }
 
 func GetNovaAPI(name types.NamespacedName) *novav1.NovaAPI {
@@ -167,7 +150,7 @@ func GetDefaultNovaCellTemplate() map[string]interface{} {
 	}
 }
 
-func CreateNova(name types.NamespacedName, spec map[string]interface{}) {
+func CreateNova(name types.NamespacedName, spec map[string]interface{}) client.Object {
 	raw := map[string]interface{}{
 		"apiVersion": "nova.openstack.org/v1beta1",
 		"kind":       "Nova",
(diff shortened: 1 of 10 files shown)


test/functional/base_test.go
```diff
@@ -85,24 +86,6 @@ func CreateNovaAPI(namespace string, spec map[string]interface{}) types.Namespac
 	return types.NamespacedName{Name: novaAPIName, Namespace: namespace}
 }
 
-func DeleteNovaAPI(name types.NamespacedName) {
```
bogdando, line 88: I think dropping these ``DeleteFoo*`` is ok, as we don't want really test k8s capabilities on removing custom resources here?

test/functional/nova_controller_test.go
```diff
@@ -121,7 +122,7 @@ var _ = Describe("Nova controller", func() {
 	When("Nova CR instance is created without cell0", func() {
 		BeforeEach(func() {
 			CreateNovaWithoutCell0(novaName)
-			DeferCleanup(DeleteNova, novaName)
+			DeferCleanup(DeleteInstance, &novav1.Nova{}, novaName)
```
gibizer, line 125: So instead of passing an empty Nova{} and a novaName separately I think we can pass a single Nova with the metadata filled. Something like: Nova{Name:novaName.Name, Namespace:novaName.Namespace}

gibizer (changes requested):

test/functional/base_test.go
mrKisaoLamb: thanks for review! IMO covering finalizer sound better than testing k8s itself

test/functional/nova_controller_test.go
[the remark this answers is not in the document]
mrKisaoLamb: Probably finalizers or any 'clearing' func that we come up in future is better plase to test for checking if resources are deleted

test/functional/base_test.go
```diff
@@ -213,24 +196,21 @@ func CreateNovaWithCell0(name types.NamespacedName) {
 	CreateUnstructured(rawNova)
 }
 
-func DeleteNova(name types.NamespacedName) {
-	logger.Info("Deleting Nova", "Nova", name)
```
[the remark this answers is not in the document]
mrKisaoLamb: Done, be38a94be3282e4b6d6c16be4bb2896e9c99aca7

test/functional/base_test.go
```diff
@@ -213,24 +196,21 @@ func CreateNovaWithCell0(name types.NamespacedName) {
...
+func DeleteInstance(instance client.Object, name types.NamespacedName) {
```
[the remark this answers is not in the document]
mrKisaoLamb: Done, be38a94be3282e4b6d6c16be4bb2896e9c99aca7

test/functional/nova_controller_test.go
mrKisaoLamb: Done, be38a94be3282e4b6d6c16be4bb2896e9c99aca7

test/functional/base_test.go
```diff
@@ -213,24 +196,23 @@ func CreateNovaWithCell0(name types.NamespacedName) {
 	CreateUnstructured(rawNova)
 }
 
-func DeleteNova(name types.NamespacedName) {
-	logger.Info("Deleting Nova", "Nova", name)
+func DeleteInstance(instance client.Object) {
 	// We have to wait for the controller to fully delete the instance
+	logger.Info("Deleting", "Name", instance.GetName(), "Namespace", instance.GetNamespace(), "Kind", instance.GetObjectKind())
```
gibizer, line 201: This result in a log

```
"Kind": "&TypeMeta{Kind:,APIVersion:,}"}
```

Probably because we are not calling the Delete with an object that has the TypeMeta part filled.

I have a fix for a particular case (see below). And if that applicable for every call then:
```golang
	logger.Info("Deleting", "Name", instance.GetName(), "Namespace", instance.GetNamespace(), "Kind", instance.GetObjectKind().GroupVersionKind().Kind)
```
would be enough here.  (Yeah `GetObjectKind` is more than what we need hence my digging into it`)

test/functional/nova_controller_test.go
```diff
@@ -175,7 +178,7 @@ var _ = Describe("Nova controller", func() {
 			DeferCleanup(th.DeleteKeystoneAPI, th.CreateKeystoneAPI(namespace))
 
 			CreateNovaWithCell0(novaName)
-			DeferCleanup(DeleteNova, novaName)
+			DeferCleanup(DeleteInstance, &novav1.Nova{ObjectMeta: metav1.ObjectMeta{Name: novaName.Name, Namespace: novaName.Namespace}})
```
gibizer, line 181: We can change this to pass the object to DeleteInstance that was created by CreateNovaWithCell0. This solves the logging issue and actually make the code more readable here.

```golang
                       DeferCleanup(DeleteInstance, CreateNovaWithCell0(novaName))
```

This needs additional work in CreateNovaWithCell0 to return the created object. See the whole change I needed to support it:

```diff
diff --git a/test/functional/base_test.go b/test/functional/base_test.go
index 54143bd..4d062b7 100644
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -50,12 +50,13 @@ const (
 	consistencyTimeout = timeout
 )
 
-func CreateUnstructured(rawObj map[string]interface{}) {
+func CreateUnstructured(rawObj map[string]interface{}) *unstructured.Unstructured {
 	logger.Info("Creating", "raw", rawObj)
 	unstructuredObj := &unstructured.Unstructured{Object: rawObj}
 	_, err := controllerutil.CreateOrPatch(
 		ctx, k8sClient, unstructuredObj, func() error { return nil })
 	Expect(err).ShouldNot(HaveOccurred())
+	return unstructuredObj
 }
 
 func GetDefaultNovaAPISpec() map[string]interface{} {
@@ -174,7 +175,7 @@ func CreateNovaWithoutCell0(name types.NamespacedName) {
 	CreateUnstructured(rawNova)
 }
 
-func CreateNovaWithCell0(name types.NamespacedName) {
+func CreateNovaWithCell0(name types.NamespacedName) client.Object {
 	rawNova := map[string]interface{}{
 		"apiVersion": "nova.openstack.org/v1beta1",
 		"kind":       "Nova",
@@ -193,12 +194,12 @@ func CreateNovaWithCell0(name types.NamespacedName) {
 		},
 	}
 
-	CreateUnstructured(rawNova)
+	return CreateUnstructured(rawNova)
 }
 
 func DeleteInstance(instance client.Object) {
 	// We have to wait for the controller to fully delete the instance
-	logger.Info("Deleting", "Name", instance.GetName(), "Namespace", instance.GetNamespace(), "Kind", instance.GetObjectKind())
+	logger.Info("Deleting", "Name", instance.GetName(), "Namespace", instance.GetNamespace(), "Kind", instance.GetObjectKind().GroupVersionKind().Kind)
 	Eventually(func(g Gomega) {
 		name := types.NamespacedName{Name: instance.GetName(), Namespace: instance.GetNamespace()}
 		err := k8sClient.Get(ctx, name, instance)
diff --git a/test/functional/nova_controller_test.go b/test/functional/nova_controller_test.go
index 09a0b32..c0e1fa6 100644
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -177,8 +177,7 @@ var _ = Describe("Nova controller", func() {
 			)
 			DeferCleanup(th.DeleteKeystoneAPI, th.CreateKeystoneAPI(namespace))
 
-			CreateNovaWithCell0(novaName)
-			DeferCleanup(DeleteInstance, &novav1.Nova{ObjectMeta: metav1.ObjectMeta{Name: novaName.Name, Namespace: novaName.Namespace}})
+			DeferCleanup(DeleteInstance, CreateNovaWithCell0(novaName))
 		})
 
 		It("registers nova service to keystone", func() {
```

I'm not 100% sure we can apply this in all DeleteInstance cases but I hope we can

gibizer (changes requested):

test/functional/nova_controller_test.go
mrKisaoLamb: Thanks! Done all CreateFooInstance return client.Object now instead nil or namespace

gibizer: Log looks good now.
```
2023-03-09T09:40:31.765+0100	INFO	---Test---	Deleting	{"Name": "c262c9f3-7a82-441d-9737-bacc0b6ca1f8", "Namespace": "723eaa21-bb60-454c-b094-769914135b30", "Kind": "Nova"}
```

gibizer (approved):
Looks good to me. Thanks for simplifying the test code!

PR #287, opened 2023-03-11
Bump operator-lint to 0.3.0

gibizer:


Makefile
```diff
@@ -305,8 +305,7 @@ golint: get-ci-tools
 
 .PHONY: operator-lint
 operator-lint: $(LOCALBIN) gowork ## Runs operator-lint
-	#TODO(gibi): bump this to v0.2.2 as soon as that version is tagged
-	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@2ffa25b7f1c13fb2bdae5444a3dd1b5bbad529b1
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.3.0
```
mrKisaoLamb, line 308: Great! I wasn't aware that we had our own linter rules. Perhaps we should consider moving this to https://github.com/openstack-k8s-operators?

gibizer: I started working on the linter during my last train trip to Brno as we had some hard learned rules that I wanted to enforce. 

I'm a bit hesitant to move it as the rules there are not openstack specific but operator framework specific. So operators outside of openstack-k8s-operators could benefit from it. But if it helps getting others on board to maintain and extend this linter then I can be convinced to move it.

mrKisaoLamb (approved):
lgtm

Makefile
gibizer: yeah that works for me. So far I'm the only contributor there, but anybody can create issues or PRs to show interest :)

PR #290, opened 2023-03-17
Fix hard-coded playbooks path

kk7ds:
When running with `make run` the hard-coded path to the playbooks is
wrong and makes the operator unable to find them. Running with the
OPERATOR_PLAYBOOKS variable set appropriately *does* work, and changing
this path makes it work for me without as well.


gibizer (approved):
Works for me. Thanks for moving the ENV definition from the test code to the makefile, it is cleaner now.

PR #295, opened 2023-03-20
Gloval beforeeach test

mrKisaoLamb:


test/functional/suite_test.go
```diff
@@ -194,3 +196,18 @@ var _ = AfterSuite(func() {
 	err := testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
 })
+
+var _ = BeforeEach(func() {
+	// NOTE(gibi): We need to create a unique namespace for each test run
+	// as namespaces cannot be deleted in a locally running envtest. See
+	// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+	namespace = uuid.New().String()
```
bogdando, line 204: hm, in a debugger session, it never reaches out the breakpoing that I put here...

gibizer: interesting. I added an extra
```
	logger.Info("!!! global BeforeEach", "namespace", namespace)
```
here and I see that being logged before each test case. So I think the BeforeEach gets called for each test. So it looks promising.

I tried to find the doc you mentioned over slack that promotes this usage but I failed. Could you please point me to it? I just want to make sure this runs by intention and not just by chance so we can rely on it in the long run being supported.

I run gingko outline but that does not pick this up as node, but I think that is a limitation of the tool as it does not pick up the BeforeSuite we have either. Probably it is limited to the given source file.
```
$ ginkgo outline --format indent test/functional/nova_controller_test.go 
Name,Text,Start,End,Spec,Focused,Pending,Labels
Describe,Nova controller,972,24365,false,false,false,""
    BeforeEach,,1673,3601,false,false,false,""
    When,Nova CR instance is created without cell0,3604,4646,false,false,false,""
        BeforeEach,,3665,3756,false,false,false,""
[...]
```

test/functional/suite_test.go
```diff
@@ -194,3 +196,18 @@ var _ = AfterSuite(func() {
...
+	th.CreateNamespace(namespace)
+	// We still request the delete of the Namespace to properly cleanup if
+	// we run the test in an existing cluster.
+	DeferCleanup(th.DeleteNamespace, namespace)
+	// NOTE(gibi): ConfigMap generation looks up the local templates
+	// directory via ENV, so provide it
+	DeferCleanup(os.Setenv, "OPERATOR_TEMPLATES", os.Getenv("OPERATOR_TEMPLATES"))
+	os.Setenv("OPERATOR_TEMPLATES", "../../templates")
```
gibizer, line 212: this is moved to the makefile since then on master.

test/functional/suite_test.go
mrKisaoLamb: I pick up by analogy of aftereach from this part of documentation: https://onsi.github.io/ginkgo/#suite-setup-and-cleanup-beforesuite-and-aftersuite . Aa far as I understand Ginko create top lvl node where we run AfterSuit and BeforeSuit and also there can be run AfterEach and BeforeEach I will give a try on bigger number of tests then

mrKisaoLamb: I tested this one more time with debuger in vscode (plus additional env values for          "OPERATOR_TEMPLATES":"../../templates",  "OPERATOR_PLAYBOOKS":"../../playbooks"). And if ginko is run from lvlv of testsuit debuger work in global BeforeEach as expected. Author of ginko by deasign put posibilities of global beforeeach: https://github.com/onsi/ginkgo/issues/18#issuecomment-29078502 plus like I mentioned in previous comment in documentation pattern for global aftereach is documented. So IMO we are good to go with that solution, OI created proper PR with changes for all tests https://github.com/openstack-k8s-operators/nova-operator/pull/297

PR #300, opened 2023-03-22
enable cinder config gen for all services

SeanMooney:


gibizer:
> enable cinder config gen for all services

It only enables it for metadata, scheduler and compute, conductor and api still uses the old config generation code that rely on a different template.

gibizer (approved):
Looks good

PR #303, opened 2023-03-23
Add GINKGO_ARGS env var to test target

gibizer:
With this change you can pass arbitrary parameters to the ginkgo test executor via the make target.

E.g. if you want to run only a subset of test sequentially failing on the first error then you can do that with:

 `make test GINKGO_ARGS="--focus 'does not create cell1 if cell0 fails as cell1 needs API access' --fail-fast  --procs 1"`


mrKisaoLamb (approved):
lgtm!

PR #304, opened 2023-03-24
Refactor ensureSecret to return secret

mrKisaoLamb:
To reduce the number of requests for secrets to Kubernetes, now return a secret in ensureSecret()


controllers/common.go
```diff
@@ -99,7 +99,7 @@ func ensureSecret(
 	reader client.Reader,
 	conditionUpdater conditionUpdater,
 	requeueTimeout time.Duration,
-) (string, ctrl.Result, error) {
+) (string, ctrl.Result, corev1.Secret, error) {
```
gibizer, line 102: Now that we are touching this I can see a one more step.

There is a hidden contract between `ensureSecret` and `ensureConfigMaps` (and the other users of the secret). 

The `ensureSecret` call ensures that the `Secret` has the fields defined in `expectedFields`. But then the `ensureConfigMaps` can try to read any fields out from that `Secret` without any assurance that those extra fields are present in the Secret. Imagine that in the next feature we need to read out a new field from the Secret. Then the developer can go and change `ensureConfigMaps` to read out the extra field from the `Secret` without realizing that they need to modify `ensureSecret` call for the extra insurance. 

I think we can make this contract explicit and therefor prevent the potential bug described above.

What if (please challenge this idea) instead of returning the whole Secret we just return a view into the Secret.Data that only show the fields that we ensured to exists? As far as I see golang does not have a construct of a "view" so the implementation would be a data copy from Secret.Data to a new `map[string][]byte` based on the keys in `expectedFields`. Then pass that restricted data set to `ensureConfigMaps`. 
This way if a new field is needed from a Secret and the developer just start using the new field in `ensureConfigMaps` then the value will not be available until the call to `ensureSecret` is also modified to ensure the existence of the new field.

One note on the implementation, ensureSecret returns the hash of the used values. That could be moved out from ensureSecret now if we want as what is returned from ensureSecret is exactly the data we want to include in the hash. However there is catch, the map[] does not have a stable order so the hashing needs to use ordered iteration either by sorting, or by using the hardcoded key order in `expectedFields`.

gibizer (commented):
nice refactoring. I have an improvement proposal if we change the interface of ensureSecret anyhow.

controllers/common.go
SeanMooney: so if we are updating this we should deligate the verification to the new libCommon functions VerifySecret.

added in https://github.com/openstack-k8s-operators/lib-common/commit/8d4d1a38b86d970608abfa205c94881d14c9ca86

i would prefer to either update aht to return the secrete and or filter it or just return the secret.

this current ensureSecret implementation dose not actually follow our nameing scheme as the ensure* functions are epexted to also update the state to the desired state in most cases.

as in we have ensure Configmap and ensure deployment type function that take a desire state and creat/update the resource if they don't exist to ensure that state.

that is why i asked for the VerifySecret function to be renamed when it was imported into libcommnet

my counter propals is to add a VerifyAndReturnSecret function to lib common and then call that form VerifySecret and just discatr the returned secrete

that way we can keep the API of VerifySecret intact but then 
we can use the VerifyAndReturnSecret in nova to implement this optimisation.


if we want to support filtering instead of trunting a map[string][]byte

i would add a filter bool to VerifyAndReturnSecret and have it make a new secrete and return that with only the data files listed in the expected fields copied over.

I get where gibi si coming form with the filtering since it prevents you using a field that you have not asserted should be present however
this would still be a runtime error when we do the access into the data or map[string][]byte so i personally don't think we should add that.

it wont help with staticly checking the field access.

SeanMooney (dismissed):
im -1 on the curent patch and gibi's proposal but i have a counter proposal in line

to;dr you should add a [VerifyAndReturnSecret](https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L389-L426) function to lib common

and make https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L389-L426 call and discard the result.

that way we can have both forms without breaking the API compatibility

gibizer:
> im -1 on the curent patch and gibi's proposal but i have a counter proposal in line
> 
> to;dr you should add a [VerifyAndReturnSecret](https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L389-L426) function to lib common
> 
> and make https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L389-L426 call and discard the result.
> 
> that way we can have both forms without breaking the API compatibility

I'm OK to have helpers in lib-common. But there is a catch. When we discussed VerifySecret the agreement was not to add condition setting code to the lib-common. So either we re-discuss that, or we loose the half of the generality of the current unsureSecret in nova-operator as that does the condition setting for us today too.

SeanMooney:
we can keep the nova verient but delegate to the lib_common VerifyAndReturnSecret function and just have the nova function do the condition handling if that works for you.

SeanMooney:
we had debate naming but if we wrapp the lib common function i would prefer to have a slightly different name just to call out it has different behavior so CheckSecret or something like that i would prefer to change it form ensure as to me simple that if the secret does not exist ensure would create it.

gibizer:
I'm OK to rename it and delegate the core logic to lib-common while keeping a wrapper in nova-operator with the generic condition setting logic. 

I agree to drop the idea of returning a filtered map based on expectedFields, as indeed it only helps moving the runtime error around. 

Having a proper static dependency would required some reflection magic to fill the fields of a pre-defined struct with data from a map[] dynamically and that is overly complicated for now.

mrKisaoLamb:
I think we shouldn't set conditions in lib-common. There's no reason to keep anything more than really simple one-purpose functions where only one thing is happening.
So I prefer create func in nova-operator and use GetSecret from lib-common and, based on the error, set conditions. Next check values in secret and set condition based on that. 
The reason for making the function a part of the nova-operator and only using the GetSecret function from lib-common is because when I looked at other operators, they handle all the conditions themselves. That's why the lib-common VerifySecret function is almost never used and does not fit into the flow of operators. So I prefere not add IMO next almost unused function.

Also I have additional question about error handling, should we use WrapErrorForObject instead standard fmt.Errorf ?

controllers/common.go
```diff
@@ -111,6 +111,7 @@ func ensureSecret(
 				fmt.Sprintf(novav1.InputReadyWaitingMessage, "secret/"+secretName.Name)))
 			return "",
 				ctrl.Result{RequeueAfter: requeueTimeout},
+				*secret,
 				fmt.Errorf("Secret %s not found", secretName)
```
gibizer, line 115: good point, we should wrap the err into the new one. I would do:
```golang
fmt.Errorf("Secret %s not found: %w", secretName, err)
```
The lib-common error wrapper does more than what we need as the logger already has all the fields it would just duplicate fields in the log line. I expect that wrapper will go away when we fix our logging across the board.

gibizer:
> I thing because we shouldn't set conditions in lib-common there's no reason to keep anything more than really simple one-purpose functions where only one thing is happening. So I prefer create func in nova-operator and use GetSecret from lib-common and, based on the error, set conditions. Next check values in secret and set condition based on that. The reason for making the function a part of the nova-operator and only using the GetSecret function from lib-common is because when I looked at other operators, they handle all the conditions themselves. That's why the lib-common VerifySecret function is almost never used and does not fit into the flow of operators. So I prefere not add IMO next almost unused function.

I'm OK either way:
* VerifySecret (or a VerifySecret2) returns the Secret object and nova-operator has a wrapper to set Conditions genericly
* nova-operator uses GetSecret and does it's own generic fields checking and condition setting.

> 
> Also I have additional question about error handling, should we use WrapErrorForObject instead standard fmt.Errorf ?

You are right we should wrap the error but I would not sue the lib-common function for it. See inline

SeanMooney:
can we chat about the design tomorrow in our sync.
lest not add a VerifySecret2 I dislike that c convention for abi compatibility.

the reason to introduce VerifyAndReturnSecret is to not break the ansible EE operator which is using VerifySecret and to enable us to deprecate and remove that if it makes sense to do so.

this is just the standard way to do API changes without breaking your clients.
we should treat lib-common as if its a released lib and try not to break API compatiablity.
because go is staticly linked we do not have to worry about abi compatibility the same way as we would if it were dymicaly linked. in other words we only have to do this rename dance if we change the function signature (including the return type)

gibizer:
> can we chat about the design tomorrow in our sync. lest not add a VerifySecret2 I dislike that c convention for abi compatibility.
> 
> the reason to introduce VerifyAndReturnSecret is to not break the ansible EE operator which is using VerifySecret and to enable us to deprecate and remove that if it makes sense to do so.
> 
> this is just the standard way to do API changes without breaking your clients. we should treat lib-common as if its a released lib and try not to break API compatiablity. because go is staticly linked we do not have to worry about abi compatibility the same way as we would if it were dymicaly linked. in other words we only have to do this rename dance if we change the function signature (including the return type)

I'm OK to call it `VerifyAndReturnSecret`

bogdando:
can we unblock this to get an immediate improvement by reducing the number of API calls, then follow-up with the above proposals?

controllers/novaconductor_controller.go
```diff
@@ -137,7 +137,7 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		required_secret_fields = append(required_secret_fields, instance.Spec.PasswordSelectors.APIDatabase)
 	}
 
-	secretHash, result, err := ensureSecret(
+	secretHash, result, _, err := ensureSecret(
```
bogdando, line 140: not sure if I resolved this merge conflict properly, @mrKisaoLamb ?

bogdando:
The "Nova reconfiguration.*when networkAttachemnt is added to a conductor while the definition is missing.*applies new NetworkAttachments configuration to that Conductor" case failure looks floating, I've it 4 times passed and 1 failed of 5 total runs locally

SeanMooney (approved):
approving with teh caveat that this is adding more technical debt since we already had a todo time to remove this function and adopt the one form lib-common

so we still need to extend lib-common with a new function that returns the secrete and then replace this with that.

PR #305, opened 2023-03-24
Limit the conductor workers

gibizer:
By default the number of conductor workers is equal to the number of CPU
on the node. Since we want to scale by number of replicas in a
deployment and not by number of workers in a pod this patch limits the
number of conductor workers to 1 per pod.

This is a stopgap solution until:
- conductor is moved to use the new config gen codepath and use the
  common config template that already has [conductor]workers set to 1
- we find a proper default for this value by running tempest against the
  clusters to see if 1 worker becomes a bottleneck or not

### DIFF
--- a/templates/novaconductor/config/nova.conf
+++ b/templates/novaconductor/config/nova.conf
@@ -1,3 +1,6 @@
+[conductor]
+workers = 1
+
 [api]
 auth_strategy = keystone
 


gibizer:
Let's debase what should be the magic number here

templates/nova.conf
```diff
@@ -60,7 +60,7 @@ enforce_new_defaults=false
 {{ if eq .service_name "nova-conductor"}}
 [conductor]
 # scaling should be done by running more pods
-workers=1
```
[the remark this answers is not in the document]
kk7ds: Consider me suitably RTFM'd.

templates/nova.conf
```diff
@@ -60,7 +60,7 @@ enforce_new_defaults=false
...
+workers=4
```
kk7ds, line 63: So from the chat, sounds like this should go back to 1.

templates/novaconductor/config/nova.conf
```diff
@@ -1,3 +1,6 @@
+[conductor]
+workers = 4
```
kk7ds, line 2: ...and we will set this to 1 for now as well.

gibizer: Done

SeanMooney (approved):
yep this is what we talked about in the sync

PR #306, opened 2023-03-24
Adding operator-sdk markers

fao89:
https://sdk.operatorframework.io/docs/building-operators/golang/references/markers/


gibizer (commented):
This seems harmless but wondering if we have a common doc somewhere describing the startegy. E.g. what print columns we should use.

SeanMooney:
i do not have time to load the context for this today but ya lets chat about it next week.

I'm not entirely sure of the implications of these changes so i want to read the doc you linked and what additionalPrinterColumns actually does.

we may or may not have time to land this before the demo but I'm not against this in principal just want to understand it before we merge.

Issue #309, opened 2023-03-27
NovaMetadata service is connected to the service=nova-api pods. 


gibizer:
https://github.com/openstack-k8s-operators/nova-operator/blob/165d29fe64b1b749a600512e52403fd18daa6548/controllers/novametadata_controller.go#L437-L443

calls
https://github.com/openstack-k8s-operators/nova-operator/blob/165d29fe64b1b749a600512e52403fd18daa6548/controllers/novaapi_controller.go#L603-L607

that makes nova-metadata service connected to nova-api pods instead of the nova-metadata pods.

```
$ oc describe  pod nova-metadata | grep service=
                  service=nova-metadata
```

```
$ oc describe service nova-metadata | grep service=
                          service=nova-api
Selector:                 service=nova-api
```

This causes connection error when ovn-metadata agent tries to connect to the nova-metadata from the external compute

PR #310, opened 2023-03-27
Fix service labels for NovaMetadata

gibizer:
The nova-metadata service wrongly used the service=nova-api selector to route requests. This is fixed now.

Closes: #309


mrKisaoLamb (approved):
lgtm

PR #314, opened 2023-03-28
Init AEE with NetworkAttachments

fao89:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/nova.go#L56-L58


gibizer:
This is one of the [TODO](https://github.com/openstack-k8s-operators/nova-operator/pull/268#pullrequestreview-1344058191) left when merged the external compute. @fao89 thanks for picking it up. I would like to see envtest coverage for this change, but that first needs coverage for the base cases. @SeanMooney hopefully can help with that.

PR #316, opened 2023-03-28
Use service defaults for force_config_drive

karelyatin:
OVN/Nova metadata is now working fine so
let's use service defaults i.e metadata
instead of config_drive.

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -23,8 +23,6 @@ initial_cpu_allocation_ratio=4.0
 initial_ram_allocation_ratio=1.0
 initial_disk_allocation_ratio=0.9
 {{end}}
-{{/*using a config drive will void issues with ovn and metadata*/}}
-force_config_drive=True
 {{if .transport_url}}transport_url={{.transport_url}}{{end}}
 
 {{if eq .service_name "nova-api"}}


SeanMooney:
no we intentiolla want to use config drive by default because the ovn metadata is unreliable.

there is a but in neutron currently where it sends the network vif plugged event before it has actually configured metadata in the ovn metadata agent.

This causes vms to sometimes fail to get ssh keys on boot and it causes random test failure in the gate.

using config drive does not disbale the ability to use the metadata API, it complementary and its what we want to configure by default in our product going forward.

karelyatin:
> no we intentiolla want to use config drive by default because the ovn metadata is unreliable.
> 
> there is a but in neutron currently where it sends the network vif plugged event before it has actually configured metadata in the ovn metadata agent.
> 
> This causes vms to sometimes fail to get ssh keys on boot and it causes random test failure in the gate.
> 
> using config drive does not disbale the ability to use the metadata API, it complementary and its what we want to configure by default in our product going forward.

Thanks @SeanMooney for the context, iiuc the gate issue you mentioned is seen upstream(opendev), can you share the bug link or discussions around in case you have it.

I got the complementary option, Just trying to understand if this was a known issue why it's not being targeted for previous OSP releases.

SeanMooney:
The metadata probably is an upstream neutron issue that has not been addressed on master.
i don't know if/when they plan to fix it or if they have a tacker but it has been an issue for years. I'm going to close this for now give we have a way to orgiride this via the CRs

PR #318, opened 2023-03-29
Use nova-blank in nova-api and nova-metadata

mrKisaoLamb:


SeanMooney (approved):
this looks good to me thanks
for cleaning this up

PR #320, opened 2023-03-30
Include roles using their FQCN

rebtoor:
Since edpm-ansible now is a "real" collection and it's installed in the ee runner image with ansible-galaxy, its roles should be called with their FQCN.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/86

### DIFF
--- a/playbooks/deploy-libvirt.yaml
+++ b/playbooks/deploy-libvirt.yaml
@@ -93,7 +93,7 @@
         }
     - name: deploy nova container
       ansible.builtin.include_role:
-        name: edpm_container_manage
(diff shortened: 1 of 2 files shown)


rebtoor:
@gibizer https://github.com/openstack-k8s-operators/edpm-ansible/pull/86 has been merged, can we merge also this patch? Thanks a lot!

bogdando (approved):

gibizer (approved):
Dependency merged so this can land too

gibizer:
/unhold
dependency merged

PR #326, opened 2023-04-03
Fix typos

gibizer:
This PR is motivated by the fix https://github.com/openstack-k8s-operators/lib-common/pull/231

Depends-on #280 (merged)

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -53,13 +53,13 @@ const (
 	NovaAPIReadyInitMessage = "NovaAPI not started"
 
 	// NovaAPIReadyErrorMessage
-	NovaAPIReadyErrorMessage = "NovaAPI error occured %s"
+	NovaAPIReadyErrorMessage = "NovaAPI error occurred %s"
 
 	// NovaConductorReadyInitMessage
 	NovaConductorReadyInitMessage = "NovaConductor not started"
 
 	// NovaConductorReadyErrorMessage
-	NovaConductorReadyErrorMessage = "NovaConductor error occured %s"
+	NovaConductorReadyErrorMessage = "NovaConductor error occurred %s"
 
 	// NovaAllCellsDBReadyInitMessage
 	NovaAllCellsDBReadyInitMessage = "DB creation not started"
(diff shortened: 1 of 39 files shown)


mrKisaoLamb (commented):
Nice! Have you considered adding some tooling for this kind of check, like the 'misspell' linter from golangci-lint? It can check all committed files and report the most common misspellings. It also has an ignore option if we want to avoid correction of some words. I ran it locally and got quite similar results.

SeanMooney:
im warey of addign tooling for spelling as it will disproportionately affect me.
i will often have spelling issues in comment ectra and i don't want that to block me committing a change locally as somthime that misspelling will be nova or pre-commit or some other term like vdpa. i don't really want to have to maintain a dictionary of those in three or otherwise have to work around the tooling.

we could try it but to me typos in code comments are not a reason to reject a patch.

if it does not affect the functioning of the code it is not a bug or code quality issue.
that is what i want to use linter for.

we could try it but if its makes it significantly harder for me to work on this repo i would like to be able to remove it in the future.

SeanMooney (approved):

mrKisaoLamb:
> im warey of addign tooling for spelling as it will disproportionately affect me. i will often have spelling issues in comment ectra and i don't want that to block me committing a change locally as somthime that misspelling will be nova or pre-commit or some other term like vdpa. i don't really want to have to maintain a dictionary of those in three or otherwise have to work around the tooling.
> 
> we could try it but to me typos in code comments are not a reason to reject a patch.
> 
> if it does not affect the functioning of the code it is not a bug or code quality issue. that is what i want to use linter for.
> 
> we could try it but if its makes it significantly harder for me to work on this repo i would like to be able to remove it in the future.

Sure, I understand what you mean. I have the same issue (it's even more noticeable in Polish). I saw that the tool has an autocommit option, so in theory it can correct errors automatically, but I haven't tested it yet. If this check is annoying, I'll drop the question.

gibizer:
I have a spell check plugin in vscode locally but my reason of this change wasn't to fix typos in code / code comments, but to fix typos in external facing messages, like our condition status messages. That could be seen by the end user and having typos there is a marketing thing, I guess. :) But then I get carried away and fixed the code typos as well for consistency. My experience with automatic spell checking is that it needs a project / codebase specific allowlist to not being a blocker, and keeping that allow list sane is annoying.

Bottom line: If having an automated spellcheck with a manually maintained allowlist is not a burden then lets go for it, but I afraid it will be.

PR #330, opened 2023-04-06
Fix the zap config in envtest

gibizer:
The previous config forced verbose mode even if it is not requested via the command line.

Also this patch disables verbose mode in the make test target by default to limit the test output of a successfull run. It can be re-enabled with make test GINKGO_ARGS='-v'.


mrKisaoLamb (approved):
lgtm

SeanMooney (approved):

PR #335, opened 2023-04-14
Webhooks for default images

abays:
Default container images are now handled via env vars and internal webhooks. This defaulting logic will in turn later be integrated with OpenStack Operator via webhooks there that use the new defaulting mechanism (while at the same time disabling the Nova service operator internal webhooks in that context).


api/v1beta1/nova_types.go
```diff
@@ -85,17 +85,17 @@ type NovaSpec struct {
 	Debug Debug `json:"debug,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default={replicas:1}
+	// +kubebuilder:default={replicas:1,containerImage:""}
```
abays, line 88: This is needed to satisfy the new `Required` constraint of the sub-components.  Otherwise CRD installation/creation fails.

api/v1beta1/nova_webhook.go
```diff
@@ -0,0 +1,129 @@
...
+package v1beta1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// NovaDefaults -
```
SeanMooney, line 32: this seam incomplete to me.

when we discussed this before we had 3 levels of precidnece

1st the CR field
2nd the env vars
3rd a default in code that would be used when nothing is set in the env

this seams to be missing form this webhook implemenation.

main.go
```diff
@@ -130,7 +131,40 @@ func main() {
...
 	}
 
+	// Acquire environmental defaults and initialize Nova defaults with them
+	novaDefaults := novav1beta1.NovaDefaults{
+		APIContainerImageURL:       os.Getenv("NOVA_API_IMAGE_URL_DEFAULT"),
+		ConductorContainerImageURL: os.Getenv("NOVA_CONDUCTOR_IMAGE_URL_DEFAULT"),
+		MetadataContainerImageURL:  os.Getenv("NOVA_METADATA_IMAGE_URL_DEFAULT"),
+		NoVNCContainerImageURL:     os.Getenv("NOVA_NOVNC_IMAGE_URL_DEFAULT"),
+		SchedulerContainerImageURL: os.Getenv("NOVA_SCHEDULER_IMAGE_URL_DEFAULT"),
+	}
```
SeanMooney, line 141: so what missing to resolve my previous comment
is we need to add ad decault for all the os.Getenv() calls to set a default in code if there is no env var defined and we don't have it set in the CR.

Makefile
```diff
@@ -136,7 +136,18 @@ PROC_CMD = --procs ${PROCS}
 
 .PHONY: test
 test: manifests generate fmt vet envtest ginkgo ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) -v debug --bin-dir $(LOCALBIN) use $(ENVTEST_K8S_VERSION) -p path)" OPERATOR_TEMPLATES="$(PWD)/templates" OPERATOR_PLAYBOOKS="$(PWD)/playbooks" $(GINKGO) --trace --cover --coverpkg=../../pkg/nova,../../pkg/novaapi,../../pkg/novaconductor,../../controllers,../../api/v1beta1,../../pkg/novascheduler,../../pkg/novametadata --coverprofile cover.out --covermode=atomic --randomize-all ${PROC_CMD} $(GINKGO_ARGS) ./test/...
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) -v debug --bin-dir $(LOCALBIN) use $(ENVTEST_K8S_VERSION) -p path)" \
+	OPERATOR_TEMPLATES="$(PWD)/templates" \
+	OPERATOR_PLAYBOOKS="$(PWD)/playbooks" \
+	NOVA_API_IMAGE_URL_DEFAULT=quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo \
```
SeanMooney, line 142: if we have defaulted this in code properly then this should not be required.

SeanMooney (changes requested):

api/v1beta1/nova_webhook.go
abays: That is not the pattern we have used elsewhere, but it is something we can consider.

test/functional/suite_test.go
mrKisaoLamb, line 1: We either do not have an image in config/samples/nova_v1beta1_novaexternalcompute.yaml (although we do have a containerimage in other samples), or there may be an issue with the defaulting in the test helper function (then we also should remove containerImage from other samples)

api/v1beta1/nova_webhook.go
```diff
@@ -0,0 +1,129 @@
+/*
+Copyright 2022.
```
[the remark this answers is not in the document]
abays: Scaffolding injected this.  Will fix.

api/v1beta1/webhook_suite_test.go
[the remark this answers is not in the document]
abays: This file is created and adding by the `operator-sdk` scaffolding command for webhooks.  We can look into placing it elsewhere, but this is the recommended location for whatever reason.

main.go
abays: That does seem like an appropriate and reasonable thing to do.  Perhaps fall back to predefined constant if the associated environment variable is missing or is the zero-value empty string.

api/v1beta1/nova_webhook.go
```diff
@@ -0,0 +1,129 @@
+/*
+Copyright 2023.
```
SeanMooney, line 2: as an fyi I generally just avoid this entirely and only add the license header. That has been my upstream policy for years and what i would prefer to do here too
especially since we are not ascribing the copyright to anyone in this case.

so i would prefer to remove these rather then update them but this is not sufficient enough to warrant a respin on its own just a nit.

config/samples/nova_v1beta1_nova_collapsed_cell.yaml
```diff
@@ -11,9 +11,11 @@ spec:
         # conductor in cell1 will act both as the super conductor and as the
         # cell1 conductor
         replicas: 0
+        containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
```
SeanMooney, line 14: we should not need to specify this in the samples
this should be defaulted by the defaulting webhook

SeanMooney: that is sepseicaly true when setting replicas: 0

i would be more ok with this if it was "" instead of a URL
but it should not be required to set contaienrImage in the samples.
we put effort into making sure they were not required in the past so this would be a regression if its currently required with this patch.

main.go
SeanMooney: yep so i think instead of s.Getenv

we can do something like this

func defaultEnv(key, default string) string {
    if value, ok := os.LookupEnv(key); ok {
        return value
    }
    return default
}


novaDefaults := novav1beta1.NovaDefaults{
		APIContainerImageURL:       os.defaultEnv("NOVA_API_IMAGE_URL_DEFAULT", "quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo"),
		ConductorContainerImageURL: os.defaultEnv("NOVA_CONDUCTOR_IMAGE_URL_DEFAULT","quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo"),
		MetadataContainerImageURL:  os.defaultEnv("NOVA_METADATA_IMAGE_URL_DEFAULT","quay.io/tripleozedcentos9/openstack-nova-api:current-tripleo"),
		NoVNCContainerImageURL:     os.defaultEnv("NOVA_NOVNC_IMAGE_URL_DEFAULT","quay.io/tripleozedcentos9/openstack-nova-novnc:current-tripleo"),
		SchedulerContainerImageURL: os.defaultEnv("NOVA_SCHEDULER_IMAGE_URL_DEFAULT","quay.io/tripleozedcentos9/openstack-nova-scheduler:current-tripleo"),
	}

gibizer (commented):
About the testing changes. 
I think we can enable the webhook to run during the existing envtests so 
1. we can minimize the test change needed
2. we can cover the webhook logic during the testing. 

I did some trials here https://github.com/gibizer/simple-operator-example/commit/16a188a210efee50ebb7cc54785c73b145bff8b8#diff-0b47813b56876b912e922bfe885d4529c58407ff08d6b7c422abed6aff774d65 about webhooks in envtest.

api/v1beta1/nova_types.go
abays: Yes, I think we can switch it to optional and remove these.  I'll push another change for those and to include fall-back defaults for the various env vars as @SeanMooney suggested, and then the `containerImages` should hopefully always get a default value.  In that sense the `required` validation won't be needed.

gibizer:
Here is my stab on integrating the webhook with the existing envtest https://github.com/gibizer/nova-operator/commit/d1074224f9e746b4749e5e08ecc1938de5b16ba1

config/samples/nova_v1beta1_nova_collapsed_cell.yaml
abays: The problem currently is that the `envtest` logic doesn't use the webhooks but does use these samples.  So for my initial local `envtest` testing, I had to add these.  @gibizer is investigating a solution to this such that webhooks can be used in `envtest`.

abays:
> @abays: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/nova-operator-build-deploy 	[3685d00](https://github.com/openstack-k8s-operators/nova-operator/commit/3685d00dbc6de21411991060125e700cf52bf2d1) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/335/pull-ci-openstack-k8s-operators-nova-operator-master-nova-operator-build-deploy/1652957531942686720) 	false 	`/test nova-operator-build-deploy`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=nova-operator&pr=335). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:abays).

This CI failure seems to be caused by something setting the webhooks to be disabled (this snippet is from the `nova-controller-manager` `describe` log):

```
Environment:
      NOVA_API_IMAGE_URL_DEFAULT:         quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
      NOVA_CONDUCTOR_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified
      NOVA_METADATA_IMAGE_URL_DEFAULT:    quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
      NOVA_NOVNC_IMAGE_URL_DEFAULT:       quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified
      NOVA_SCHEDULER_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified
      NOVA_COMPUTE_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified
      NOVA_LIBVIRT_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified
      NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT:  quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
      ENABLE_WEBHOOKS:                    false    <------- *** webhooks disabled ***
      WATCH_NAMESPACE:                    openstack
      OPERATOR_CONDITION_NAME:            openstack-operator.v0.0.1
```

abays:
> > @abays: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > Test name 	Commit 	Details 	Required 	Rerun command
> > ci/prow/nova-operator-build-deploy 	[3685d00](https://github.com/openstack-k8s-operators/nova-operator/commit/3685d00dbc6de21411991060125e700cf52bf2d1) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/335/pull-ci-openstack-k8s-operators-nova-operator-master-nova-operator-build-deploy/1652957531942686720) 	false 	`/test nova-operator-build-deploy`
> > [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=nova-operator&pr=335). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:abays).
> 
> This CI failure seems to be caused by something setting the webhooks to be disabled (this snippet is from the `nova-controller-manager` `describe` log):
> 
> ```
> Environment:
>       NOVA_API_IMAGE_URL_DEFAULT:         quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
>       NOVA_CONDUCTOR_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified
>       NOVA_METADATA_IMAGE_URL_DEFAULT:    quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
>       NOVA_NOVNC_IMAGE_URL_DEFAULT:       quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified
>       NOVA_SCHEDULER_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified
>       NOVA_COMPUTE_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified
>       NOVA_LIBVIRT_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified
>       NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT:  quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
>       ENABLE_WEBHOOKS:                    false    <------- *** webhooks disabled ***
>       WATCH_NAMESPACE:                    openstack
>       OPERATOR_CONDITION_NAME:            openstack-operator.v0.0.1
> ```
> 
> As a result, the `Nova` CR does not have its various `containerImage` fields populated and they are instead the empty-string.

Actually, this is proper.  The webhooks should be turned off in the Nova controller-manager.  I think this is a chicken-and-egg problem, however.  While Nova controller-manager's webhooks should be disabled in the context of an OpenStack operator deployment (which this is), the OpenStack operator needs to have this version of Nova so that it can call the defaulting mechanism from within its own webhooks.  My guess is that previous initial webhook PRs for other operators avoided this problem by happening to have the `containerImage`s set in the `OpenStackControlPlane` sample YAML [1] used by CI.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane.yaml

api/v1beta1/nova_webhook.go
abays: BTW, I did end up adding this suggestion

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -66,19 +66,16 @@ type NovaExternalComputeSpec struct {
 	// but does not do any modification on the compute node itself.
 	Deploy bool `json:"deploy"`
 
-	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
abays: Ack, added

pkg/common/env.go
```diff
@@ -0,0 +1,12 @@
...
+import "os"
+
+// TODO: Move to lib-common?
+// GetEnvDefault - Get the value associated with key from environment variables, but use baseDefault as a value in the case of an empty string
+func GetEnvDefault(key string, baseDefault string) string {
+	if value, ok := os.LookupEnv(key); ok && value != "" {
+		return value
+	}
+	return baseDefault
+}
```
gibizer, line 12: I don't think we need to check `value != ""` if the ENV variable is set we get `ok==true`  if it is unset then we get `ok==false`.  If somebody explicitly set the ENV variable to "" then we are allowed to fail later if the explicitly set "" value is invalid.

pkg/common/env.go
```diff
@@ -0,0 +1,12 @@
+package common
+
+import "os"
+
+// TODO: Move to lib-common?
```
gibizer, line 5: Yepp, lets move it there. But it is OK to do that in a follow up

api/v1beta1/webhook_suite_test.go
```diff
@@ -0,0 +1,138 @@
+/*
```
gibizer, line 1: We can drop this whole file. The webhook is started by the test/functional/suite_test.go now.

test/functional/suite_test.go
```diff
@@ -184,12 +199,59 @@ var _ = BeforeSuite(func() {
...
+	novav1.SetupNovaCellDefaults(novaCellDefaults)
+
+	// Acquire environmental defaults and initialize NovaExternalCompute defaults with them
+	novaExternalComputeDefaults := novav1.NovaExternalComputeDefaults{
+		ComputeContainerImageURL:   common.GetEnvDefault("NOVA_COMPUTE_IMAGE_URL_DEFAULT", novav1.NovaComputeContainerImage),
+		LibvirtContainerImageURL:   common.GetEnvDefault("NOVA_LIBVIRT_IMAGE_URL_DEFAULT", novav1.NovaLibvirtContainerImage),
+		AnsibleEEContainerImageURL: common.GetEnvDefault("NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT", novav1.AnsibleEEContainerImage),
+	}
+
+	novav1.SetupNovaExternalComputeDefaults(novaExternalComputeDefaults)
```
gibizer, line 229: This is now the same as in main.go so to avoid divergence of this setup between the test run and the production run, I suggest to pull it out to a common place.

gibizer:
> > > @abays: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > > Test name 	Commit 	Details 	Required 	Rerun command
> > > ci/prow/nova-operator-build-deploy 	[3685d00](https://github.com/openstack-k8s-operators/nova-operator/commit/3685d00dbc6de21411991060125e700cf52bf2d1) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/335/pull-ci-openstack-k8s-operators-nova-operator-master-nova-operator-build-deploy/1652957531942686720) 	false 	`/test nova-operator-build-deploy`
> > > [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=nova-operator&pr=335). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:abays).
> > 
> > 
> > This CI failure seems to be caused by something setting the webhooks to be disabled (this snippet is from the `nova-controller-manager` `describe` log):
> > ```
> > Environment:
> >       NOVA_API_IMAGE_URL_DEFAULT:         quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
> >       NOVA_CONDUCTOR_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified
> >       NOVA_METADATA_IMAGE_URL_DEFAULT:    quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
> >       NOVA_NOVNC_IMAGE_URL_DEFAULT:       quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified
> >       NOVA_SCHEDULER_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified
> >       NOVA_COMPUTE_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified
> >       NOVA_LIBVIRT_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified
> >       NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT:  quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
> >       ENABLE_WEBHOOKS:                    false    <------- *** webhooks disabled ***
> >       WATCH_NAMESPACE:                    openstack
> >       OPERATOR_CONDITION_NAME:            openstack-operator.v0.0.1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As a result, the `Nova` CR does not have its various `containerImage` fields populated and they are instead the empty-string.
> 
> Actually, this is proper. The webhooks should be turned off in the Nova controller-manager. I think this is a chicken-and-egg problem, however. While Nova controller-manager's webhooks should be disabled in the context of an OpenStack operator deployment (which this is), the OpenStack operator needs to have this version of Nova so that it can call the defaulting mechanism from within its own webhooks. My guess is that previous initial webhook PRs for other operators avoided this problem by happening to have the `containerImage`s set in the `OpenStackControlPlane` sample YAML [1] used by CI.
> 
> [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane.yaml

I agree with your analysis. This is one of the example where testing nova-operator PR with the latest openstack-operator version does not really work as this nova-operator PR will only be compatible with openstack-operator after the webhook in openstack-operator is turned on. 

I think one way to solve this is to not remove the defaulting of the containerImage fields by kubebuilder in this PR to keep the compatibility with the openstack-operator. Then we can merge this PR with green CI. Then bump nova-operator version in openstack-operator and do the webhook integration in the same PR. Then a second nova-operator PR is needed to remove the kubebuilder defaulting from containerImages here.

An alternative solution could be to temporary merge a PR in openstack-operator where the nova part of the samples define the containerImage fields explicitly. And revert that after we did the webhook integration there.

config/samples/nova_v1beta1_nova_collapsed_cell.yaml
gibizer: I think we should be able to drop these (here and below) from the samples now.

config/samples/nova_v1beta1_novacell0.yaml
```diff
@@ -8,8 +8,12 @@ spec:
   cellDatabaseHostname: openstack
   cellMessageBusSecretName: rabbitmq-transport-url-nova-api-transport
   cellName: cell0
-  conductorServiceTemplate: {}
+  conductorServiceTemplate:
+    replicas: 1
+    containerImage: quay.io/tripleozedcentos9/openstack-nova-conductor:current-tripleo
```
gibizer, line 13: I think it should work even if we leave the whole `conductorServiceTemplate` field out from the sample. If it does not work today that means we have yet another defaulting issue around this field in the template struct. I'm fine to investigate this in a separate PR.

Makefile
abays: Updated to not need this

config/samples/nova_v1beta1_nova_collapsed_cell.yaml
abays: Ack, will try

main.go
```diff
@@ -130,7 +132,53 @@ func main() {
...
+		}
+		if err = (&novav1beta1.NovaCell{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "NovaCell")
+			os.Exit(1)
+		}
+		if err = (&novav1beta1.NovaExternalCompute{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "NovaExternalCompute")
+			os.Exit(1)
+		}
+	}
```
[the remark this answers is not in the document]
abays: Understood.  If there are use cases where all the individual types could be created independently, then we should add webhooks for them.  I just started with `Nova`, `NovaCell` and `NovaExternalCompute` based on my current (limited) understanding.  I will do those in a follow-up PR.

pkg/common/env.go
abays: Ack, will remove

test/functional/suite_test.go
abays: Good suggestion.  I will do that.

api/v1beta1/nova_types.go
```diff
@@ -164,3 +164,28 @@ func (c Nova) IsReady() bool {
 	readyCond := c.Status.Conditions.Get(condition.ReadyCondition)
 	return readyCond != nil && readyCond.Status == corev1.ConditionTrue
 }
+
+// NewNovaSpec returns a NovaSpec where the fields are defaulted according
+// to the CRD definition
+func NewNovaSpec(passwordSecretName string) NovaSpec {
```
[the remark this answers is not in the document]
abays: Ack, will remove

abays:
> > > > @abays: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > > > Test name 	Commit 	Details 	Required 	Rerun command
> > > > ci/prow/nova-operator-build-deploy 	[3685d00](https://github.com/openstack-k8s-operators/nova-operator/commit/3685d00dbc6de21411991060125e700cf52bf2d1) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/335/pull-ci-openstack-k8s-operators-nova-operator-master-nova-operator-build-deploy/1652957531942686720) 	false 	`/test nova-operator-build-deploy`
> > > > [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=nova-operator&pr=335). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:abays).
> > > 
> > > 
> > > This CI failure seems to be caused by something setting the webhooks to be disabled (this snippet is from the `nova-controller-manager` `describe` log):
> > > ```
> > > Environment:
> > >       NOVA_API_IMAGE_URL_DEFAULT:         quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
> > >       NOVA_CONDUCTOR_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified
> > >       NOVA_METADATA_IMAGE_URL_DEFAULT:    quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
> > >       NOVA_NOVNC_IMAGE_URL_DEFAULT:       quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified
> > >       NOVA_SCHEDULER_IMAGE_URL_DEFAULT:   quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified
> > >       NOVA_COMPUTE_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified
> > >       NOVA_LIBVIRT_IMAGE_URL_DEFAULT:     quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified
> > >       NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT:  quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
> > >       ENABLE_WEBHOOKS:                    false    <------- *** webhooks disabled ***
> > >       WATCH_NAMESPACE:                    openstack
> > >       OPERATOR_CONDITION_NAME:            openstack-operator.v0.0.1
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > As a result, the `Nova` CR does not have its various `containerImage` fields populated and they are instead the empty-string.
> > 
> > 
> > Actually, this is proper. The webhooks should be turned off in the Nova controller-manager. I think this is a chicken-and-egg problem, however. While Nova controller-manager's webhooks should be disabled in the context of an OpenStack operator deployment (which this is), the OpenStack operator needs to have this version of Nova so that it can call the defaulting mechanism from within its own webhooks. My guess is that previous initial webhook PRs for other operators avoided this problem by happening to have the `containerImage`s set in the `OpenStackControlPlane` sample YAML [1] used by CI.
> > [1] https://github.com/openstack-k8s-operators/openstack-operator/blob/master/config/samples/core_v1beta1_openstackcontrolplane.yaml
> 
> I agree with your analysis. This is one of the example where testing nova-operator PR with the latest openstack-operator version does not really work as this nova-operator PR will only be compatible with openstack-operator after the webhook in openstack-operator is turned on.
> 
> I think one way to solve this is to not remove the defaulting of the containerImage fields by kubebuilder in this PR to keep the compatibility with the openstack-operator. Then we can merge this PR with green CI. Then bump nova-operator version in openstack-operator and do the webhook integration in the same PR. Then a second nova-operator PR is needed to remove the kubebuilder defaulting from containerImages here.
> 
> An alternative solution could be to temporary merge a PR in openstack-operator where the nova part of the samples define the containerImage fields explicitly. And revert that after we did the webhook integration there.

Okay, let's try keeping the `kubebuilder` defaults for now and then removing them in a follow-up.

gibizer (commented):
This looks OK to me. We discussed with @abays that there will be follow ups:
* after the openstack-operator is using the nova-operator defaulting interface the kubebuilder defaults can be removed from nova-operator for containerImage fields
* separately, a followup can add webhooks for the sub CRs not handled today.

@SeanMooney do you have any open issues with this PR or can we land this now?

api/v1beta1/nova_webhook.go
SeanMooney: ack

SeanMooney (approved):

PR #339, opened 2023-04-18
Pointer types for certain API fields

abays:
It has come to our attention that defaulting webhooks can lead to strange behavior in regards to `kubebuilder` default annotations, due to the intermediate interaction of the defaulting webhooks:

1. The webhooks intercept the API request before it reaches the `kubebuilder` defaulting logic, which causes any field that is unset in the request and untouched by the webhook to be set to its zero-value
2. Thus the field is set to the zero-value in etcd by the webhook action, as if a user had explicitly done so in the CR itself
3. This results in the `kubebuilder` defaulting logic that follows to skip defaulting that field (because some other logic has already set an explicit value)

Therefore, if the `kubebuilder` default is specified as something other than the non-zero value for such a field described above, then that non-zero `kubebuilder` default will never be set and the default is always (incorrectly) the zero-value.  Using a pointer for the field type avoids this problem by allowing the field that the user did not explicitly provide to remain `nil` in and after the defaulting webhook logic (if the webhook does not in fact explicitly set it to something), and thus permits the `kubebuilder` defaulting logic to act on the `nil` field and set the specific default value.


abays:
This seems to work with `envtest` both as an isolated PR and when merged with #335.  The same is true for testing deployments via `install_yamls`.

api/v1beta1/common_types.go
```diff
@@ -33,7 +33,7 @@ type NovaServiceBase struct {
 	// +kubebuilder:validation:Maximum=32
 	// +kubebuilder:validation:Minimum=0
 	// Replicas of the service to run
-	Replicas int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas"`
```
gibizer, line 36: * Resources below could have similar problem if we ever remove the omitempty tag from it. As far as I see the RsourceRequirements struct is full of omitempty fields and no direct scalars are exposed so we are probably OK.
* CustomServiceConfig is OK as is. There is no semantic difference between not defined CustomServiceConfig field and when the field is defaulted to "". Both result in not having an extra config file provided to the service. So there it is OK to keep an optional string field.
* ContainerImage always defaulted by the webhook itself if not provided so we never expect an "" value of it here.

* The bool fields in the Debug struct  all defaulted to false and I'm OK to assume the same false default when the field is not provided and defaulted to the golang false default. So I think it is OK for now to keep those as is.

* I think PasswordSelector only works now as where that type is used it has a default value defined as e.g. `// +kubebuilder:default={service: NovaPassword}`. That is a repetition and we can probably get rid of that repetition if we switch the fields in PasswordSelector to *string.

* MetalLBConfig.SharedIP seems problematic as the default value there is defined as true. So I'm wondering how that gets defaulted now if a MetalLBConfig without an explicit SharedIP field is sent in the request.

* MetalLBConfig.Endpoint was discussed extensively before so I guess the Optional + Enum combination tested to work OK in this situation. 

* MetalLBConfig.SharedIPKey is optional too but it is defaulted to "" so we are probably OK with that.

api/v1beta1/novaapi_types.go
```diff
@@ -42,7 +42,7 @@ type NovaAPITemplate struct {
 	// +kubebuilder:validation:Maximum=32
 	// +kubebuilder:validation:Minimum=0
 	// Replicas of the service to run
-	Replicas int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas"`
```
gibizer, line 45: In the NovaAPISpec (I hate that github does not allow me to comment directly there) we have a bunch of problematic string fields where the default value is different from the golang default like ServiceUser. I think this is not a problem now for two reasons. 1) we don't have a direct webhook yet on NovaAPI 2) in a normal case NovaAPISpec is created from the NovaAPITemplate within the Nova CR and from the NovaSpec itself, so if the Nova Spec is defaulted properly then the field copy will bring that specific default to the NovaAPISpec as well.

So it remain to be see what happens when we add a webhook to the NovaAPI CR.

Also this comment is valid for the rest of the serviceSpec struct except NovaExternalCompute as that already has a webhook.

api/v1beta1/common_types.go
gibizer: I tried the MetalLBConfig.ShareIP defaulting in envtest with:
```diff
diff --git a/test/functional/nova_controller_test.go b/test/functional/nova_controller_test.go
index fb25b3b..f98dad5 100644
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -785,6 +785,9 @@ var _ = Describe("Nova controller", func() {
                        api := GetNovaAPI(novaAPIName)
                        Expect(api.Spec.NetworkAttachments).To(Equal(nova.Spec.APIServiceTemplate.NetworkAttachments))
                        Expect(api.Spec.ExternalEndpoints).To(Equal(nova.Spec.APIServiceTemplate.ExternalEndpoints))
+                       //Check that ShareIP is defaulted properly
+                       Expect(nova.Spec.APIServiceTemplate.ExternalEndpoints[0].SharedIP).To(Equal(true))
+                       Expect(api.Spec.ExternalEndpoints[0].SharedIP).To(Equal(true))
 
                        scheduler := GetNovaScheduler(novaSchedulerName)
                        Expect(scheduler.Spec.NetworkAttachments).To(Equal(nova.Spec.APIServiceTemplate.NetworkAttachments))
diff --git a/test/functional/novaapi_controller_test.go b/test/functional/novaapi_controller_test.go
index 9dd1689..8262d29 100644
--- a/test/functional/novaapi_controller_test.go
+++ b/test/functional/novaapi_controller_test.go
@@ -571,6 +571,11 @@ var _ = Describe("NovaAPI controller", func() {
                        keystoneEndpointName := types.NamespacedName{Namespace: namespace, Name: "nova"}
                        th.SimulateKeystoneEndpointReady(keystoneEndpointName)
 
+                       // Check that ShareIP is defaulted to true automatically
+                       // NOTE(gibi): this might break when webhook is added to NovaAPI CRD
+                       instance := GetNovaAPI(novaAPIName)
+                       Expect(instance.Spec.ExternalEndpoints[0].SharedIP).To(Equal(true))
+
                        // As the internal endpoint is configured in ExternalEndpoints it does not
                        // get a Route but a Service with MetalLB annotations instead
                        service := GetService(types.NamespacedName{Namespace: namespace, Name: "nova-internal"})
```

And the defaulting works there properly. Honestly I don't know why this is not broken the same way as Deploy or Replicas. But it is not borken.

gibizer: I've also confirmed that if I remove the repeated PasswordSelector defaulting then things start to break with:

```
  2023-04-19T14:03:43.220+0200	ERROR	Reconciler error	{"controller": "novascheduler", "controllerGroup": "nova.openstack.org", "controllerKind": "NovaScheduler", "NovaScheduler": {"name":"d4304a3b-d85c-4ef5-bf79-e9a5e2eb6082","namespace":"ae0875f7-aac1-4325-8886-c17a185ec01a"}, "namespace": "ae0875f7-aac1-4325-8886-c17a185ec01a", "name": "d4304a3b-d85c-4ef5-bf79-e9a5e2eb6082", "reconcileID": "885d2bfb-1ba2-4b0a-9597-690c94e28183", "error": "field '' not found in secret/test-secret"}
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
  	/home/gibi/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
  	/home/gibi/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
  	/home/gibi/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235
```
The message `"field '' not found in secret/test-secret"` shows that PasswordSelector is not defaulted any more.

gibizer: Interestingly switching the fields of the PasswordSelector to `*string`s does not help to trigger the proper defaulting...

gibizer: neither if I add omitempty either or both to the PasswordSelector fields and to the field where PasswordSelector is used as a field type. The only way I can make PasswordSelector defaulting work if I repeat at least one of the fields with its default value at the field where the PasswordSelector is used as a field type i.e.:
```
	// +kubebuilder:default={service: NovaPassword}
	// PasswordSelectors - Field names to identify the passwords from the
	// Secret
	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
```

I'm open to suggestions :/

abays: @gibizer When you say...

> I've also confirmed that if I remove the repeated PasswordSelector defaulting then things start to break

Do you mean if you remove the defaulting for it such as...

```
	// +kubebuilder:default={service: NovaPassword}     <----- imagine this is removed
	// PasswordSelectors - Field names to identify the passwords from the
	// Secret
	PasswordSelectors PasswordSelector `json:"passwordSelectors"`
```

...but leave this in place in the `PasswordSelector` struct definition itself?

```
	// +kubebuilder:default="NovaPassword"
	// Service - Selector to get the keystone service user password from the
	// Secret
	Service string `json:"service"`
```

...and in that case it does not default properly?

gibizer: Yes, that is what I did.

abays: I've banged my head against the wall thinking about this on and off for a week.  No solutions have become obvious to me.  I'm not really sure how to proceed here.  But I would ask, for the moment: are we talking about potential future problems here?  I just ask that because it seems like this works as-is given our current CRD definitions.  It seems to be that changes to these definitions such as removing the repeated `PasswordSelector` defaults introduces the problem.  Is the goal here to make sure that isn't possible, or is it perhaps reasonable to expect developers to define CRDs in such a way that they don't encounter these errors?

api/v1beta1/novaapi_types.go
dprince: @gibizer trying to follow your logic. Are you trying to make a case that we not include webhooks because there might be a future parameter that we would have to adjust to support them? Things seem be working in this PR though, and we have them in most other operators now too (Nova is one of the last to adopt these patches).

gibizer: @dprince No. My point is that we might see more defaulting issues when we add a webhook on the NovaAPI (and other subCR) level.

api/v1beta1/common_types.go
gibizer: @abays yes, it is not an immediate problem. We can keep the repeated defaults of the PasswordSelector. I just tried to understand why that is needed to find a way to remove the duplication. All these duplication are potential source of future bugs where we want to change a default but forget to update all the places the same default is mentioned.

Anyhow I have no better option than to keep the PasswordSelector as is.

api/v1beta1/novaapi_types.go
gibizer: > Things seem be working in this PR though, and we have them in most other operators now too (Nova is one of the last to adopt these patches).

I'm not sure what do you mean by this. The Replicas *int32 change is not landed everywhere as far as I see in https://github.com/search?q=org%3Aopenstack-k8s-operators+Replicas&type=code&p=3 

If you mean the webhook work then sure that is progressing nicely. However the current nova-operator webhook PR does not pass CI at the moment.

gibizer (approved):
Looks OK. 

* I still think we will see more defaulting complications when we add webhook to the sub CRs. 
* It feels dangerous to duplicate default values for PasswordSelectors but we was not able to come up with a solution that avoids them.

PR #344, opened 2023-04-24
Replace tripleo-repos by repo-setup

kajinamit:
... because TripleO is being deprecated.


gibizer (approved):
looks good

SeanMooney:
i guess there was no harm in merging this but just as an FYI i will be deleting this soon
it was intended to be deleted prior to the q1 demo so we don't need to keep this in sync anymore.

it is not currently used.

ill delete it later this week or early next when i finish my current in progress work

gibizer:
Yeah, I was aware of that this is just a local setup code for test env that will go away.

PR #345, opened 2023-04-26
[envtest] Use MessageBusSecret consistently

gibizer:
CellMessageBusSecretName and MessageBusSecret was used inconsistently in the conductor tests. It is fixed now by removing CellMessageBusSecretName


mrKisaoLamb (approved):
lgtm

PR #346, opened 2023-04-26
Metadata multi cell

mrKisaoLamb:
Add support for multicell deployment

Closes: [OSPRH-143](https://issues.redhat.com//browse/OSPRH-143)

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -482,7 +482,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return result, err
 	}
 
-	if instance.Spec.MetadataServiceTemplate.Replicas != 0 {
+	if *instance.Spec.MetadataServiceTemplate.Replicas != 0 {
 		result, err = r.ensureMetadata(
 			ctx, h, instance, cell0Template,
 			cellDBs[Cell0Name].Database, apiDB, apiMQSecretName, keystoneAuthURL,
@@ -492,7 +492,6 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		}
 	} else {
 		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
-		return ctrl.Result{}, nil
 	}
 
 	util.LogForObject(h, "Successfully reconciled", instance)
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -482,12 +482,17 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return result, err
 	}
 
-	result, err = r.ensureMetadata(
-		ctx, h, instance, cell0Template,
-		cellDBs[Cell0Name].Database, apiDB, apiMQSecretName, keystoneAuthURL,
-	)
-	if err != nil {
-		return result, err
+	if instance.Spec.MetadataServiceTemplate.Replicas != 0 {
+		result, err = r.ensureMetadata(
+			ctx, h, instance, cell0Template,
+			cellDBs[Cell0Name].Database, apiDB, apiMQSecretName, keystoneAuthURL,
+		)
+		if err != nil {
+			return result, err
+		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
+		return ctrl.Result{}, nil
 	}
 
 	util.LogForObject(h, "Successfully reconciled", instance)
--- a/controllers/novacell_controller.go
+++ b/controllers/novacell_controller.go
@@ -120,6 +120,9 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		if err != nil {
 			return result, err
 		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
+		return ctrl.Result{}, nil
 	}
 
 	util.LogForObject(h, "Successfully reconciled", instance)
@@ -150,6 +153,11 @@ func (r *NovaCellReconciler) initConditions(
 				novav1.NovaConductorReadyCondition,
 				condition.InitReason,
 				novav1.NovaConductorReadyInitMessage,
+			),
+			condition.UnknownCondition(
+				novav1.NovaMetadataReadyCondition,
+				condition.InitReason,
+				novav1.NovaMetadataReadyInitMessage,
 			),
 		)
 		instance.Status.Conditions.Init(&cl)
--- a/controllers/novacell_controller.go
+++ b/controllers/novacell_controller.go
@@ -115,6 +115,13 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		return result, err
 	}
 
+	if instance.Spec.MetadataServiceTemplate.Replicas != 0 {
+		result, err = r.ensureMetadata(ctx, h, instance)
+		if err != nil {
+			return result, err
+		}
+	}
+
 	util.LogForObject(h, "Successfully reconciled", instance)
 	return ctrl.Result{}, nil
 }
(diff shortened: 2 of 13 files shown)


controllers/nova_controller.go
```diff
@@ -437,12 +437,17 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		result, err = r.ensureMetadata(
+			ctx, h, instance, cell0Template,
+			cellDBs[Cell0Name].Database, apiDB, apiMQSecretName, keystoneAuthURL,
+		)
+		if err != nil {
+			return result, err
+		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
+		return ctrl.Result{}, nil
```
gibizer, line 450: you can drop this, so that we end up logging "Successfully reconciled" in both Replica==0 and !=0 cases.

controllers/novacell_controller.go
```diff
@@ -115,6 +115,16 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
 	}
 
+	if instance.Spec.MetadataServiceTemplate.Replicas != 0 {
+		result, err = r.ensureMetadata(ctx, h, instance)
+		if err != nil {
+			return result, err
+		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
+		return ctrl.Result{}, nil
```
gibizer, line 125: ditto

controllers/novacell_controller.go
```diff
@@ -201,10 +216,62 @@ func (r *NovaCellReconciler) ensureConductor(
 	return ctrl.Result{}, nil
 }
 
+func (r *NovaCellReconciler) ensureMetadata(
```
gibizer, line 219: how similar this to the ensureMetata in NovaReconciler? Can we pull out a common function?

controllers/novacell_controller.go
```diff
@@ -201,10 +216,62 @@ func (r *NovaCellReconciler) ensureConductor(
...
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1.NovaCell,
+) (ctrl.Result, error) {
+	metadataSpec := novav1.NewNovaMetadataSpec(instance.Spec)
```
gibizer, line 224: I guess the main difference between the two ensureMetadata is what is the input we use for creating the metadataSpec.

test/functional/nova_metadata_controller_test.go
```diff
@@ -311,6 +314,46 @@ var _ = Describe("NovaMetadata controller", func() {
...
+			metadata := CreateNovaMetadata(namespace, spec)
+			novaMetadataName = types.NamespacedName{Name: metadata.GetName(), Namespace: metadata.GetNamespace()}
+			DeferCleanup(DeleteInstance, metadata)
+			DeferCleanup(
+				k8sClient.Delete,
+				ctx,
+				CreateNovaMetadataSecret(namespace, SecretName),
+			)
+		})
+		It("generated config with correct local_metadata_per_cell", func() {
```
gibizer, line 330: Please add more check to the name of the exposed service name, as that also depends on the CellName now.

controllers/novacell_controller.go
```diff
@@ -115,6 +115,16 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		return result, err
 	}
 
+	if instance.Spec.MetadataServiceTemplate.Replicas != 0 {
```
gibizer, line 118: I remember we discussed that we don't need nova-metadata service deployed to cell0 as there are never running instances in cell0 as there are no computes in cell0. I think we should simply reject the cell0Template in the validation webhook if it has a metadataTemplate with replicas > 0. As the webhook PR is still open (but I think close to merge) this needs to be done in a follow up after the webhook PR lands.

gibizer (changes requested):
Overall looks good I have some minor comments inline.

Also I have an additional request to add a Nova sample file that deploys metadata. And a follow up request to do a validation so that nova-metadata is never requested to be deployed to cell0.

controllers/novacell_controller.go
mrKisaoLamb: Done

controllers/novacell_controller.go
mrKisaoLamb: Yes and because this difference it's hard to find good type to valid what we use there, using interface and cases make code overcomplicated IMO

controllers/nova_controller.go
```diff
@@ -481,12 +481,16 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if *instance.Spec.MetadataServiceTemplate.Replicas != 0 {
+		result, err = r.ensureMetadata(
+			ctx, h, instance, cell0Template,
+			cellDBs[Cell0Name].Database, apiDB, apiMQSecretName, keystoneAuthURL,
+		)
+		if err != nil {
+			return result, err
+		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
```
gibizer, line 494: This looks OK. We are only saving the instance in the deferred function at the end of Reconcile. So even though we add NovaMetadataReadyCondition at the start and removing it here, this is not visible to the outside word as we only save the conditions after this point.

controllers/novacell_controller.go
```diff
@@ -115,6 +115,15 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		return result, err
 	}
 
+	if *instance.Spec.MetadataServiceTemplate.Replicas != 0 && instance.Spec.CellName != Cell0Name {
```
gibizer, line 118: if we already rejecting MetadataServiceTemplate.Replicas != 0 for cell0 in the validation hook, then the extra check `instance.Spec.CellName != Cell0Name` here is not necessary.

controllers/novacell_controller.go
gibizer: OK. I guess we can maintain this 50 lines here and in the nova_controller. No worries.

test/functional/nova_multicell_test.go
```diff
@@ -641,6 +641,151 @@ var _ = Describe("Nova multicell", func() {
...
+			spec := GetDefaultNovaSpec()
+			cell0Template := GetDefaultNovaCellTemplate()
+			cell0Template["cellName"] = "cell0"
+			cell0Template["cellDatabaseInstance"] = novaNames.APIMariaDBDatabaseName.Name
+			cell0Template["cellDatabaseUser"] = "nova_cell0"
+			cell0Template["hasAPIAccess"] = true
+			// disable cell0 conductor
+			cell0Template["conductorServiceTemplate"] = map[string]interface{}{
+				"replicas": 0,
+			}
```
gibizer, line 744: You copied here the setup for collapsed cells from above. I think you don't need such complication here to test the metadata per cell setup.

controllers/novacell_controller.go
gibizer: //later
I see we haven't added this to the validation hook. So the check here is OK. But I suggest to do a follow up where we reject cell0 metadata from the validation webhook.

gibizer (commented):
My only currently actionable comment is about mixing collapsed cell setup and metadata per cell in the test case.

test/functional/nova_multicell_test.go
mrKisaoLamb: Done, 2a7d74b9df71524bcea48ab779f1d53d3c97de11

test/functional/nova_multicell_test.go
```diff
@@ -713,4 +713,87 @@ var _ = Describe("Nova multicell", func() {
...
+			DeferCleanup(th.DeleteDBService, th.CreateDBService(cell0.MariaDBDatabaseName.Namespace, cell0.MariaDBDatabaseName.Name, serviceSpec))
+			DeferCleanup(th.DeleteDBService, th.CreateDBService(cell1.MariaDBDatabaseName.Namespace, cell1.MariaDBDatabaseName.Name, serviceSpec))
+
+			spec := GetDefaultNovaSpec()
+			cell0Template := GetDefaultNovaCellTemplate()
+			cell0Template["cellDatabaseInstance"] = cell0.MariaDBDatabaseName.Name
+			cell0Template["cellDatabaseUser"] = "nova_cell0"
+
+			cell0Template["metadataServiceTemplate"] = map[string]interface{}{
+				"replicas": 1,
```
gibizer, line 741: OK, so this gets ignored. And later on this will be rejected by the validation webhook. Cool

gibizer (approved):
looks good

gibizer (approved):
looks good

PR #347, opened 2023-04-27
Dynamic serviceaccount/role/rolebinding creation

dprince:
This PR dyamically creates the above resources within the namespace of the requested Nova instance.

The above resources are no longer created via OLM.

This allows the operator to deploy openstack service into any namespace within the OpenShift cluster.


controllers/nova_controller.go
```diff
@@ -144,6 +154,27 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+			Verbs:         []string{"use"},
+		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"pods"},
+			Verbs:     []string{"create", "get", "list", "watch", "update", "patch", "delete"},
+		},
+	}
+	rbacResult, err := common_rbac.ReconcileRbac(ctx, h, instance, rbacRules)
+	if err != nil {
```
SeanMooney, line 172: nit: this can just
if err != nil || rbacResult != ctrl.Result{} {
return rbacResult, err
}

SeanMooney (approved):

PR #348, opened 2023-04-27
[envtest]Remove duplicated NovaScheduler instance

gibizer:
The NovaScheduler controller test created a NovaScheduler instance
before each test. However there are test cases later in the file that
also creates its own NovaScheduler instance with specific configuration.
So in those tests there are two NovaScheduler instances being
reconciled. This makes troubleshooting the test complex. So this patch
separates the tests with the global NovaScheduler to a deparate Describe
section. So that the rest of the tests can  create its own NovaScheduler
with the needed configuration.

### DIFF
--- a/test/functional/nova_scheduler_test.go
+++ b/test/functional/nova_scheduler_test.go
@@ -289,6 +289,15 @@ var _ = Describe("NovaScheduler controller", func() {
 			)
 		})
 	})
+})
+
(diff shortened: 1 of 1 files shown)


bogdando (commented):
lgtm

bogdando:
I'm working towards extending its scope in the names unifcation PR as well

mrKisaoLamb (approved):
lgtm

PR #352, opened 2023-05-02
Fix missing template keys

gibizer:
After the fix[1] in lib-common template utils missing keys in templates became hard failures. So this PR bumps the lib-common dependency along with replacing bare key reading in nova.conf template with calling index.

[1]https://github.com/openstack-k8s-operators/lib-common/pull/245


fmount:
+1 for me, I'm testing the same approach in glance and it works properly.

SeanMooney (approved):
ok this is resonably 

if this was python we are basically sawping form

if x.y:

to 

if hasttrib(x,y):

i.e. using a function to test if the value exist.

I'm fine with this approch

PR #353, opened 2023-05-03
Lib common bump test helper split

gibizer:
Adapting to https://github.com/openstack-k8s-operators/lib-common/pull/252

PR #354, opened 2023-05-04
Lib common test helper usage

gibizer:
Removed a couple of duplicated helper code as they are now available in lib-common

Depends-on: #353


mrKisaoLamb (approved):
lgtm

PR #355, opened 2023-05-04
normalize DEFAULT_IMG

yazug:
make bundle was leaving modified kustomizeation.yaml Add DEFAULT_IMG to the makefile to match other projects Update config/manager/kustomization.yaml with current generating values


gibizer (approved):
looks good to me

PR #359, opened 2023-05-05
[envtest]Remove unknown cellName field from template

gibizer:
The Nova.Spec.CellTemplate[].CellName is not a valid field. It was probably mixed up with NovaCell.Spec.CellName which exists. The k8s client ignores undefined fields so this only caused a warning in the test logs. Still it is better not to have it so now it is removed.


SeanMooney (approved):
yep simple fix thanks gibi

PR #362, opened 2023-05-08
[envtest]Force binaries to run on ipv4 localhost

gibizer:
If the localhost is resolved to ::1 instead of 127.0.0.1 then checking the webhook availability fails at the start of suite as it tries to parse the address as ipv4 and fails on the colons with:
```
  dial tcp: address ::1:40931: too many colons in address
```
So this PR forces envtest to always use 127.0.0.1


bogdando:
lgtm, thank you!

gibizer:
/test functional
Feels like an infra issue
```
  * could not run steps: execution cancelled
  * could not run steps: step functional failed: test "functional" failed: could not watch pod: context canceled 
[36mINFO[0m[2023-05-08T16:59:15Z] Reporting job state 'failed' with reason 'executing_graph:interrupted' 
```

bogdando (approved):

PR #364, opened 2023-05-10
Add basic nova scale kuttl tests

jamepark4:
Adds initial multi-cell KUTTL tests for the nova operator.

Closes: [OSPRH-96](https://issues.redhat.com//browse/OSPRH-96)


test/kuttl/tests/multi_cell/common/rabbitmq_cell1/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- rabbitmq_cell1.yaml
+namespace: openstack
+patches:
+- patch: |-
+    - op: replace
+      path: /metadata/name
+      value: rabbitmq-cell1
```
gibizer, line 10: I'm not sure about the goal of this kustomize. Is the goal to later allow creating more than one RabbitMQCluster from the same rabbitmq_cell1.yaml? How that will look like?

test/kuttl/tests/multi_cell/common/rabbitmq_cell1/rabbitmq_cell1.yaml
```diff
@@ -0,0 +1,15 @@
...
+  override:
+    statefulSet:
+      spec:
+        template:
+          spec:
+            securityContext: {}
+            containers: []
+            initContainers:
+            - name: setup-container
+              securityContext: {}
```
gibizer, line 15: kudos to the RabbitMQ devs that the almost empty sample creates a valid cluster.

kuttl-multicell-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+#  kubectl-kuttl test --config $NOVA_KUTTL_TEST_DIR/kuttl-multicell-test.yaml 
+#
+#
+# ASSUMPTIONS:
+#
+# 1. Latest version of kuttl is installed at /usr/local/bin/kubectl-kuttl
+#    - wget https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_x86_64
+#    - mv kubectl-kuttl_0.11.1_linux_x86_64 /usr/local/bin/kubectl-kuttl
+#    - chmod 755 /usr/local/bin/kubectl-kuttl
+# 2. An OCP 4.10+ CRC cluster with dependent operators (details below using devsetup and install_yamls)
```
SeanMooney, line 13: nit: OCP 4.12

we do no support 4.10 anymore

kuttl-multicell-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+#    2g. make keystone
+#    2h. make placement
+#    2i. make rabbitmq
+#    2j. make infra
+#    2k. make ansibleee
+#    2l. make nova
+#    2m. make mariadb_deploy
+#    2n. make keystone_deploy
+#    2o. make placement_deploy
+#    2p. make rabbitmq_deploy
```
SeanMooney, line 29: +1
thanks for documenting this
it will make automating this in a ci job much simpler

SeanMooney (commented):
overall this looks reasonable at a first glance
since we don't have ci I would like to try this out before merging but ill try and do that today if i can.

test/kuttl/tests/multi_cell/common/nova_v1beta1_nova-multi-cell.yaml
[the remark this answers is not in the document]
gibizer: This sample can be as small as:
```
kind: Nova
metadata:
  name: nova
spec:
  secret: osp-secret
```

The defaulting will still create cell0 and cell1 with all the necessary services. You might need to adjust the assert due some defaults are different than what is passed in currently here.

test/kuttl/tests/multi_cell/common/mariadb_cell1/mariadb_v1beta1_mariadb_cell1.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: mariadb.openstack.org/v1beta1
+kind: MariaDB
+metadata:
+  name: mariadb-cell1
```
gibizer, line 4: If you go with a small sample as:
```
kind: Nova
metadata:
  name: nova
spec:
  secret: osp-secret
```
then this mariadb needs to be name openstack-cell1 as the defeaulting in nova-operator uses that.

test/kuttl/tests/multi_cell/multi_cell_scale/02-assert.yaml
```diff
@@ -0,0 +1,23 @@
...
+    cell0:
+      conductorServiceTemplate:
+        replicas: 3
+    cell1:
+      conductorServiceTemplate:
+        replicas: 3
+  metadataServiceTemplate:
+    replicas: 3
+  schedulerServiceTemplate:
+    replicas: 3
```
gibizer, line 20: This part of the assert is not really checking any logic. This only checks that vaues from 02-scale-nova.yaml is persisted.

test/kuttl/tests/multi_cell/multi_cell_scale/02-assert.yaml
```diff
@@ -0,0 +1,23 @@
...
+status:
+  apiServiceReadyCount: 3
+  schedulerServiceReadyCount: 3
```
gibizer, line 23: These are good asserts.

I think it would make sense to assert:
* There is NovaAPI CR with NovaAPI.spec.replicas: 3
  This proves that Nova passed the replica setting to NovaAPI
* There is NovaScheduler CR with replicas 3
* There is NovaMetadata CR with replicas 3
* There are two NovaConductor CR (cell0 and cell1)  with replicas 3 each
* The corresponding StatefulSets and Pods are exits and running.

gibizer (commented):
Overall looks good. I was able to run it locally and the test passes. I have some suggestions inline.

kuttl-multicell-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+# 3. The environment variable NOVA_KUTTL_TEST_DIR should be set /path/to/nova-operator/test/kuttl/tests
```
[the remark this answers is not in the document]
jamepark4: I updated the config file to now have the pointer to the directory in the hopes of allowing granularity based on the types of deployments that might exist in the future.  So for multi-cell specific tests you'd call kuttl-multicell-test.yaml and maybe call a different kuttl config file in nova-operator for another suite of tests.  Let me know if you'd prefer something different.

test/kuttl/tests/multi_cell/common/nova_v1beta1_nova-multi-cell.yaml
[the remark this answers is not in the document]
jamepark4: I just ended up removing this file completely and now I symlink to config/samples/nova_v1beta1_nova.yaml

test/kuttl/tests/multi_cell/common/rabbitmq_cell1/kustomization.yaml
jamepark4: I was following the deployment approach of install_yamls, but since it is really not needed here I'll remove it.

test/kuttl/tests/multi_cell/multi_cell_scale/02-assert.yaml
jamepark4: Ack I took your advice based on the below comment and updated this and subsequent asserts to look for additional information.

PR #365, opened 2023-05-10
always generate the service_user config when deploying nova.

SeanMooney:
As part of adressing CVE-2023-2088 nova was modifed to required
the service_user to be configured. This change adapts the
config template to enable the service_user section.
see: https://bugs.launchpad.net/nova/+bug/2004555 for details.


gibizer (approved):
Looks good. Thanks

PR #366, opened 2023-05-12
nova-operator-base job

SeanMooney:
This change adds 2 new jobs, the first is a content-provider job
to build the operators and host them in a job local registry and the second
is a nova-operator-base job that will deploy crc using the content
from the content provider job.

Related: [OSPRH-202](https://issues.redhat.com//browse/OSPRH-202)


SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
ok with the rename to main we no longer need to force the job branch which is good
but its still haveign the same problem.

it using the v2 API endpoint and then trying to parse it as v1.

https://review.rdoproject.org/zuul/build/b4b422bad68f4d51b39f821a8061206a/log/ci-framework/logs/ci_make_011_openstack-operator_-_call_bundle.log#311-315

for v2 instead of 

++ jq -r '.tags[].name'
we would expect 
++ jq -r '.tags[] 
if https://github.com/openstack-k8s-operators/openstack-operator/blob/061c5434a0f20cadf1681fd1809aea0db1373bd3/hack/pin-custom-bundle-dockerfile.sh#L42 was correctly taken.

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
check-rdo

SeanMooney:
im expectign the content provider job to pass again now that the base job has been fixed but the CRC job will still fail because crc is not configured to use an insecure regeity yet.

so the images that are coming form quay.io will be pulled properly but the nova-operator image will not be pulled porperly.

SeanMooney:
ok this failed in the way i was expecting more or less 
the content provider job works and the content is made available to the other job

https://review.rdoproject.org/zuul/build/0c5a8036c4f2454db5046e64f8194f19/log/zuul-info/inventory.yaml#9-25

the job failed to activate the OpenStack namespace since i skipped that step in install yamls

https://review.rdoproject.org/zuul/build/0c5a8036c4f2454db5046e64f8194f19/log/job-output.txt#815

SeanMooney:
check-rdo

ci/nova-operator-compute-kit/topology/OpenStackControlPlane.yaml
```diff
@@ -0,0 +1,107 @@
...
+      apiServiceTemplate:
+        externalEndpoints:
+        - endpoint: internal
+          ipAddressPool: internalapi
+          loadBalancerIPs:
+          - 172.17.0.80
+      secret: osp-secret
+      metadataServiceTemplate:
+        externalEndpoints:
+         - endpoint: internal
```
bogdando, line 61: nit: indentation

ci/nova-operator-compute-kit/topology/OpenStackControlPlane.yaml
```diff
@@ -0,0 +1,107 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  glance:
+    template:
+      databaseInstance: openstack
+      storageClass: ""
```
bogdando, line 9: would it work, if we remove this?
looks confusing a bit

SeanMooney:
check-rdo

ci/nova-operator-compute-kit/playbooks/tempest.yaml
```diff
@@ -0,0 +1,24 @@
...
+  tasks:
+    - name: Use OpenStack Namespace
+      ansible.builtin.command: oc project openstack
+      changed_when: false
+
+    - name: Deploy OpenStack
+      ansible.builtin.command: "oc apply -k {{ nova_operator_basedir }}/ci/nova-operator-compute-kit/topology/tempest"
+
+
+    - name: run tempest tests
```
SeanMooney, line 14: i should add a wait here for the pod to be ready.

.zuul.yaml
```diff
@@ -1,6 +1,57 @@
...
+    provides:
+      - "nova-operator-content"
+
+- job:
+    name: nova-operator-base
+    description: |
+      This is the base job for all nova-operator tests that will deploy
+      openstack. the base job does not deploy openstack but it does use
+      the nova-operator-content-provider job to provide a registry with
+      built operators and installs them on a crc envionment. child jobs
```
gibizer, line 14: nit: Child

.zuul.yaml
```diff
@@ -1,6 +1,57 @@
...
+      should take this deploy crc and use the installed operators to
```
gibizer, line 15: deployed

.zuul.yaml
```diff
@@ -1,6 +1,57 @@
...
+      - zuul: github.com/openstack-k8s-operators/ci-framework
+    pre-run:
+      - ci/nova-operator-base/playbooks/pre.yaml
+    post-run:
+      - ci/nova-operator-base/playbooks/collect-logs.yaml
+    vars:
+      crc_parameters: "--memory 18000 --disk-size 120 --cpus 6"
+      network_isolation: true
+      crc_attach_default_interface: true
+      cifmw_edpm_prepare_skip_crc_storage_creation: true
```
gibizer, line 34: Why do we need this?

ci/nova-operator-base/playbooks/pre.yaml
```diff
@@ -0,0 +1,33 @@
...
+  tasks:
+    - name: Add crc creds
+      ansible.builtin.include_role:
+        name: rhol_crc
+        tasks_from: add_crc_creds
+
+    - name: Disable the openshift-marketplace
+      ansible.builtin.shell: |
+        oc patch operatorhubs/cluster --type merge --patch '{"spec":{"sources":[{"disabled": true,"name": "redhat-marketplace"}]}}'
+      changed_when: true
```
gibizer, line 20: This might be moved to cifmw-base-crc-openstack parent job

ci/nova-operator-base/playbooks/pre.yaml
```diff
@@ -0,0 +1,33 @@
...
+      changed_when: false
+
+    - name: Make sure all Openstack operators are deployed
+      ansible.builtin.shell: |
+        set -o pipefail && oc get csv -l operators.coreos.com/openstack-operator.openstack --no-headers=true | grep -i "succeeded"
+      register: operator_status
+      until: operator_status.rc == 0
+      changed_when: false
+      retries: 30
+      delay: 30
```
gibizer, line 33: hm, as this is in pre.yaml so I assume the operators where deployed by the base job, but then the base job should check this as well.

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
```diff
@@ -0,0 +1,34 @@
...
+# We can override the openstack operator image by defining openstack_img in the zuul job
+
+- hosts: controller
+  name: Deploy Openstack Operators
+  vars:
+    nova_operator_basedir: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/nova-operator"
+  tasks:
+    - name: Use OpenStack Namespace
+      ansible.builtin.command: oc project openstack
+      changed_when: false
```
gibizer, line 13: I'm not sure we can remove this but this is already set in pre.yaml so maybe

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
```diff
@@ -0,0 +1,34 @@
...
+      delay: 30
+
+    - name: Make sure Openstack CR is ready
+      ansible.builtin.shell: |
+        set -o pipefail && oc get OpenstackControlPlane/openstack  -o jsonpath='{.status.conditions[?(@.type == "Ready")].status}'  | grep -i "True"
+      register: openstack_status
+      until: openstack_status.rc == 0
+      changed_when: false
+      retries: 30
+      delay: 30
```
gibizer, line 34: If OpenStackControlPlane is Ready then Nova is Ready too. But sure it does not hurt to check both.

ci/nova-operator-compute-kit/playbooks/tempest.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Use OpenStack Namespace
+      ansible.builtin.command: oc project openstack
+      changed_when: false
+
+    - name: Deploy OpenStack
+      ansible.builtin.command: "oc apply -k {{ nova_operator_basedir }}/ci/nova-operator-compute-kit/topology/tempest"
+
+    - name: wait for tempest-worker pod to start
+      ansible.builtin.shell: |
+        set -o pipefail && oc get pod/tempest-worker -o jsonpath='{.status.conditions[?(@.type == "Ready")].status}'  | grep -i "True"
```
gibizer, line 16: we could use oc's  built in wait command. E.g.
```
oc wait pod tempest-worker --for condition=Ready --timeout=120s
```

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Use OpenStack Namespace
+      ansible.builtin.command: oc project openstack
+      changed_when: false
+
+    - name: Deploy OpenStack
+      ansible.builtin.command: "oc apply -k {{ nova_operator_basedir }}/ci/nova-operator-compute-kit/topology/ci"
+
+    - name: Make sure Nova CR is ready
+      ansible.builtin.shell: |
+        set -o pipefail && oc get Nova/nova  -o jsonpath='{.status.conditions[?(@.type == "Ready")].status}'  | grep -i "True"
```
gibizer, line 20: we could use oc's built in wait command. E.g.

oc wait Nova nova --for condition=Ready --timeout=120s

ci/nova-operator-base/playbooks/pre.yaml
```diff
@@ -0,0 +1,33 @@
...
+        oc patch operatorhubs/cluster --type merge --patch '{"spec":{"sources":[{"disabled": true,"name": "redhat-marketplace"}]}}'
+      changed_when: true
+
+    - name: Use OpenStack Namespace
+      ansible.builtin.command: oc project openstack
+      changed_when: false
+
+    - name: Make sure all Openstack operators are deployed
+      ansible.builtin.shell: |
+        set -o pipefail && oc get csv -l operators.coreos.com/openstack-operator.openstack --no-headers=true | grep -i "succeeded"
```
gibizer, line 28: we can probably use oc wait for this

ci/nova-operator-compute-kit/topology/base/OpenStackControlPlane.yaml
gibizer, line 1: As an idea, could we reuse the samples from the respective operator repository? So that if the operator CRD changes we can automatically get an updated sample and don't have to do a manual update in nova-operator.

.zuul.yaml
gibizer: // later // I see a different storageClass used in the samples than locally so I guess that is the reason

ci/nova-operator-base/playbooks/collect-logs.yaml
```diff
@@ -0,0 +1,72 @@
...
+  name: Collect logs on the controller
+  gather_facts: false
+  tasks:
+    - name: Create log dir
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/zuul-output/logs/controller"
+        state: directory
+        mode: 0755
+
+    - name: Collect pod logs
```
gibizer, line 22: This is not just pod logs. But also this is really generic log collector for all the podified jobs so I would move it to the base job or to a common place so jobs in other repos does not need to duplicate this.

.zuul.yaml
SeanMooney: thechnially we could use crc to deploy with the local volume plugin but
i would prefer to use the host-path provisioned we have in the ci framework jobs.

neather are something you would use in production but the host-path provisioned is at least a dynamic driver so its close to a real storage solution

ci/nova-operator-base/playbooks/collect-logs.yaml
SeanMooney: we could to that eventually yes

for now I would prefer to keep it here
but i can move it to the ci framework later.

I'm still trying to decide what is useful to capture and what is not

eventually, i would prefer to extract this into a role i think  but for now i kept it relatively simple.

this is a modified version of

 https://github.com/openstack-k8s-operators/install_yamls/blob/main/ci/playbooks/collect_logs.yaml#L22-L57

ci/nova-operator-base/playbooks/pre.yaml
SeanMooney: this might not be requried anymore i can double check.

i added this before i changed to cifmw-base-crc-openstack but yes i would expect this to be moved eventully.

ci/nova-operator-base/playbooks/pre.yaml
SeanMooney: it should but it didnt at least when i started working on this.
you are right long  term we should not need to do this here.
currently I'm just trying to be safe and check our preconditios early
to avoid failures later.

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
SeanMooney: so i think it can be remvoed yes
because when you do oc project OpenStack it updates the default context in your kubeconfig

but i didn't double check that so i put it in incase it only did that via an evn var as this will be a separate ansible playbook and connection.

but im pretty sure this is no longer required.

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
SeanMooney: so when i googled for how to wait for things this option never came up

so ya i can try that instead of the version i came up with instead :)

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
SeanMooney: you woudl think so but no.

i think we have a bug in the OpenStack operator

i have them in this order because the OpenstackControlPlane went ready before it had created all the CRs. so i think its not inviting the status properly.

i had to put in the nova check to ensure it actually waited for everything to be ready.

ci/nova-operator-compute-kit/playbooks/tempest.yaml
SeanMooney: ack

ci/nova-operator-compute-kit/topology/base/OpenStackControlPlane.yaml
SeanMooney: we could but im intentionally not doing that

we can actually just use a URL to the git repo as the source.

the reason I'm not doing that is to avoid changes to the other repos samples altering the behaviour of the job and potentially breaking it.

there are pros and cons to both approaches

SeanMooney (commented):
thanks for the review
we have some basic yaml ccking in pre-commit
https://github.com/openstack-k8s-operators/nova-operator/blob/main/.pre-commit-config.yaml#LL42C1-L43C17

but we can look at ansible and yaml lint after yes

gibizer:
I see keytone container failures in https://review.rdoproject.org/zuul/build/04570ff27e7c48449f1095b31fb04378/log/controller/pods/keystone-6d79c59ff9-vcftz-describe.txt 
```
  Warning  Unhealthy       71s (x2 over 4m45s)  kubelet            Readiness probe failed: Get "http://10.217.0.113:5000/v3": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
  Warning  Unhealthy       66s (x5 over 12m)    kubelet            Liveness probe failed: Get "http://10.217.0.113:5000/v3": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

.zuul.yaml
gibizer: ack

ci/nova-operator-compute-kit/playbooks/deploy-openstack.yaml
gibizer: Nice. Then it is a bug in openstack-operator

ci/nova-operator-compute-kit/topology/base/OpenStackControlPlane.yaml
gibizer: OK. I see the pro/cons. Let's see how frequently this breaks

ci/nova-operator-base/playbooks/collect-logs.yaml
```diff
@@ -0,0 +1,65 @@
...
+      tags:
+        - skip_ansible_lint
+
+- hosts: all
+  name: Copy files from controller on node
+  gather_facts: false
+  tasks:
+    - name: Copy files from controller on node
+      ansible.builtin.include_role:
+        name: fetch-output
```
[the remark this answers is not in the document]
SeanMooney: ack

gibizer:
Actually the tempest result looks good. The only failed thing is `tempest.api.compute.admin.test_networks.NetworksTest.test_list_all_networks[id-df3d1046-6fa5-4b2c-ad0c-cfa46a351cb9]` but there the test case is faulty. I've proposed a fix for the test case here https://review.opendev.org/c/openstack/tempest/+/883892

SeanMooney:
i have a typo in this patch anyway that i need to address so we will see if tempest in prow passes after i fix that that

the 
ci/prow/nova-operator-build-deploy-tempest  and 
ci/prow/nova-operator-build-deploy

jobs will get replaced by the new jobs soon anyway.

gibizer (approved):
The content provider job is passing so I think this is a good start.

PR #369, opened 2023-05-15
Remove manually setting internal endpoint on KeystoneAPI

gibizer:
The CreateKeystoneAPI helper already sets both the internal and public endpoints so the test cases do not need to set it explicitly.


mrKisaoLamb (approved):
nice! lgtm

PR #370, opened 2023-05-15
Use HashOfInputHashes from lib-common

gibizer:
So that we can remove our own implementation.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -266,19 +266,6 @@ func ensureConfigMap(
 	return hash, ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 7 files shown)


mrKisaoLamb (commented):
lgtm! However, we may need to hold off on the merge until all repositories have switched to the main branch

gibizer:
The tempest job fails for unrelated reasons. Being fixed here https://review.opendev.org/c/openstack/openstack-tempest-skiplist/+/883237

mrKisaoLamb (approved):

PR #373, opened 2023-05-16
Controller manager label finalization

abays:
We now have a final form for operator controller-manager pod labeling of the format `openstack.org/operator-name: <operator name>` and are no longer relying on the inconsistent labels created by various `operator-sdk` versions.  All operators that lack this standardized pattern are being updated.


gibizer:

Tempest was green before so I would like to see what happens here. The last tempest failed in a way that suggest env issues.

gibizer (approved):
Looks good.

PR #374, opened 2023-05-16
PoC: Use dataplane extraMounts

fao89:
Currently, we pass the inventory configmap name and the ssh secret name  to mount AEE, we could get the extraMounts from dataplane instead. It would allow store the logs when dataplane.extraMounts is set

https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/logs/ 

https://github.com/openstack-k8s-operators/install_yamls/pull/271

### DIFF
--- a/api/v1beta1/novaexternalcompute_types.go
+++ b/api/v1beta1/novaexternalcompute_types.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	"github.com/openstack-k8s-operators/lib-common/modules/storage"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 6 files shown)


SeanMooney:
im not entirely convicne this is the right approach

we could leverage 
https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/logs/
to get logs but i don't agree that we should use ExtraMounts []storage.VolMounts ` on the NovaExternalCompute CRD for this.

i want to reserve ExtraMounts for future use to mount additional volume to the nova-compute and or libvirt container not the AEE

controllers/novaexternalcompute_controller.go
```diff
@@ -680,54 +680,6 @@ func initAEE(
 	// pointer indirection via ansibleEE.Spec.ExtraMounts
 	ansibleEEMounts := storage.VolMounts{}
 
-	// mount ssh keys
```
SeanMooney, line 683: i think we shoudl continue to generate the mounts like this and not take them form the dataplane operator but we can append the mounts required for logging.

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -59,6 +60,11 @@ type NovaExternalComputeSpec struct {
 	// SSHKeySecretName is the name of the k8s Secret that contains the ssh keys to access the node
 	SSHKeySecretName string `json:"sshKeySecretName"`
 
+	// ExtraMounts containing files which can be mounted into an Ansible Execution Pod
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
+	ExtraMounts []storage.VolMounts `json:"extraMounts"`
```
SeanMooney, line 66: this is not what we want extra mounts on the nova external compute to be used for

we want to keep this for additional mounts to the libvirt/nova-compute container

SeanMooney (changes requested):

fao89:
https://github.com/openstack-k8s-operators/install_yamls/pull/271#issuecomment-1550417446

PR #375, opened 2023-05-17
Always enable NovaExternalCompute webhook

abays:
We have discovered a context in which the current webhook pattern used across our operators is not currently viable.  This happens when the DataPlane operator creates a `NovaExternalCompute` CR in the context of an overall deployment via OpenStack operator.  A problem then arises because...

1. The Nova operator is deployed via OpenStack operator and thus the Nova operator webhooks are disabled
2. The DataPlane operator creates a `NovaExternalCompute` CR
3. Because the `NovaExternalCompute` CR comes from the DataPlane operator and not the OpenStack operator, it avoids the OpenStack operator's defaulting webhooks
4. Without the OpenStack operator's defaulting webhooks firing, no default images can be set on the `NovaExternalCompute` CR
5. And also, since the Nova operator webhooks are disabled, no default images are set via that mechanism as well

The stop-gap solution proposed in this PR is to always have the `NovaExternalCompute` webhook enable for the Nova operator, regardless of whether the operator is deployed separately or as part of the OpenStack operator bundle.  We will then take time to think of a potential better solution.

### DIFF
--- a/main.go
+++ b/main.go
@@ -152,10 +152,6 @@ func main() {
 			setupLog.Error(err, "unable to create webhook", "webhook", "NovaConductor")
 			os.Exit(1)
 		}
-		if err = (&novav1beta1.NovaExternalCompute{}).SetupWebhookWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create webhook", "webhook", "NovaExternalCompute")
-			os.Exit(1)
-		}
 		if err = (&novav1beta1.NovaMetadata{}).SetupWebhookWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create webhook", "webhook", "NovaMetadata")
 			os.Exit(1)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -170,6 +166,14 @@ func main() {
 		}
 	}
 
+	// NOTE: HACK: NovaExternalCompute webhooks are always enabled due to a current
```
SeanMooney, line 169: we should be enabling all the webhooks for nova not just this one.

even if the defaulting happens in the OpenStack operator we should be still running the defaulting and validating webhook on the nova operator.
can you do this for all them instead.

SeanMooney (changes requested):

main.go
abays: This won't work as a solution anyhow due to the way OpenStack operator manipulates the service operator CSVs and deployments in its `csv-merger.go`.  A different solution has been proposed in https://github.com/openstack-k8s-operators/openstack-operator/pull/314.

abays:
Closing, as testing showed this approach to be incomplete.

PR #377, opened 2023-05-19
Add missing local webhooks

abays:
I missed these in the recent webhooks follow-up.

PR #381, opened 2023-05-22
Bump nova-api probe timeout

gibizer:
In tempest we saw[1] that a longer nova API request from tempest can block the pod probes resulting in a pod restart. So this patch bumps the timeout values of the probes to see if it mitigates the issue or not.

[1] https://github.com/openstack-k8s-operators/nova-operator/pull/377#issuecomment-1557188218

### DIFF
--- a/pkg/novaapi/deployment.go
+++ b/pkg/novaapi/deployment.go
@@ -47,13 +47,13 @@ func StatefulSet(
 	// After the first successful startupProbe, livenessProbe takes over
 	livenessProbe := &corev1.Probe{
 		// TODO might need tuning
-		TimeoutSeconds: 10,
-		PeriodSeconds:  10,
+		TimeoutSeconds: 30,
+		PeriodSeconds:  30,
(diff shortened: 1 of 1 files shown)


gibizer:
/retest all
One tempest passed

gibizer:

tempest failed but that is actually a tempest issue. I try to fix that in https://review.opendev.org/c/openstack/tempest/+/883892 but if the test_list_all_networks runs early when no network is created yet by other test cases then it will fail. So we should probably skiplist this test.

No nova-api container restart was observed during the run so I considers this tempest run a success from this PR perspective.

I also noticed that currently tempest is run with parallelism of 15. This is will not work later when we enable scenario tests.

gibizer:
tempest success 3/3

gibizer:
4/4 success

SeanMooney (approved):
thanks gibi, lets proceed with this for now and see if the ci stablises.

PR #384, opened 2023-05-26
Allow skip firewall setup when starting webhooks locally

gibizer:
The hacking/configure_local_webhooks.sh depends on `firewall-cmd` to open a port in the localhost's firewall. This is distro dependent. So this patch makes it possible to skip the firewall setup via an SKIP_FIREWALL ENV variable.

```
  SKIP_FIREWALL=true OPERATOR_TEMPLATES=./templates make run-with-webhook
```


bogdando (approved):

bogdando:
ugh, I didn't expect this to be merged before the CI passes :(

gibizer:
> ugh, I didn't expect this to be merged before the CI passes :(

The deploy jobs are non voting (not required) by the prow definition. Hence tide does not wait for them.

gibizer:
deploy job failed on:
```
Error from server (NotFound): deployments.apps "horizon-operator-controller-manager" not found
```
the tempest job seems to failed also on deploying the controllers. Neither seems to be related to this change

Issue #296, opened 2023-03-20
Failed AnsibleEE CRs should be deleted before retrying


kk7ds:
A failed ansible run leaves the job and ansibleee CRs in place for forensic reasons. However, even deleting the NovaExternalCompute will cause those to not be re-created on the next run, leading to confusion about why changes are not being taken into account.

Any time we're going to do an ansible run, we should delete/recreate/update a previous failed ansibleee CR so that new changes to any of the inputs are used for the subsequent job.

bogdando:
* NovaExternalCompute does no own the Ansibleee CR it creates and therefore when NovaExternalCompute is deleted the rest of the tree remains

bogdando:
this whole thing looks like is going to be addressed as the point 2 in https://github.com/openstack-k8s-operators/nova-operator/issues/319

SeanMooney:
yep that was the plan

SeanMooney:
this was not actully adressed by #387 although it is related.

bogdando:
Related https://github.com/openstack-k8s-operators/nova-operator/issues/319

gibizer:
This is targeting NovaExternalCompute but that is unused and will be deleted by https://github.com/openstack-k8s-operators/nova-operator/pull/476

Issue #319, opened 2023-03-30
Ansible compute followup TODOs


gibizer:
TODOs left for followup:
* do not requeu while waiting for the ansible jobs
* add controller ref to the ansible jobs we create to not leak them if nova is deleted while the job running
* pass NetworkAttachments to the ansiblee CR
* create a common ensureAEEDeploy function from the two similar one
* blank out nova-compute.conf
* add envtest coverage to the negative ansible execution case.

_Originally posted by @gibizer in https://github.com/openstack-k8s-operators/nova-operator/pull/268#pullrequestreview-1344058191_

bogdando:
* NovaExternalCompute does no own the Ansibleee CR it creates and therefore when NovaExternalCompute is deleted the rest of the tree remains

gibizer:
This issue is targeting NovaExternalCompute but that is unused now and will be deleted by https://github.com/openstack-k8s-operators/nova-operator/pull/476

PR #387, opened 2023-05-29
Do not left behind AnsibleEE job pods

bogdando:
Add nova external compute controller ref to the ansibleEE jobs to not leak them, if the external nova compute CR is deleted, while the AEE job running.

Related: https://github.com/openstack-k8s-operators/nova-operator/issues/319
Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/296

Partial: [OSPRH-148](https://issues.redhat.com//browse/OSPRH-148)

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -403,10 +403,17 @@ func CreateNovaExternalComputeSSHSecret(name types.NamespacedName) *corev1.Secre
 
 func GetAEE(name types.NamespacedName) *aee.OpenStackAnsibleEE {
 	instance := &aee.OpenStackAnsibleEE{}
+	novaExt := GetNovaExternalCompute(novaNames.ComputeName)
 
 	Eventually(func(g Gomega) {
 		g.Expect(k8sClient.Get(ctx, name, instance)).Should(Succeed())
 	}, timeout, interval).Should(Succeed())
+
+	Expect(instance.GetOwnerReferences()).To(HaveLen(1))
+	Expect(instance.GetOwnerReferences()[0]).To(HaveField("Name", novaExt.GetName()))
+	t := true
+	Expect(instance.GetOwnerReferences()[0]).To(HaveField("Controller", &t))
+
 	return instance
 }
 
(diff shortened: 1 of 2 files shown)


gibizer (dismissed):
Could you please assert the controller ref in the related envtest case?

test/functional/base_test.go
```diff
@@ -407,6 +407,12 @@ func GetAEE(name types.NamespacedName) *aee.OpenStackAnsibleEE {
 	Eventually(func(g Gomega) {
 		g.Expect(k8sClient.Get(ctx, name, instance)).Should(Succeed())
 	}, timeout, interval).Should(Succeed())
+
+	Expect(instance.GetOwnerReferences()).To(HaveLen(1))
+	Expect(instance.GetOwnerReferences()[0]).To(HaveField("Name", h.GetBeforeObject().GetName()))
```
bogdando, line 412: 'h' is nill, so it fails

bogdando: figured this out

SeanMooney (approved):

test/functional/base_test.go
```diff
@@ -403,10 +403,17 @@ func CreateNovaExternalComputeSSHSecret(name types.NamespacedName) *corev1.Secre
 
 func GetAEE(name types.NamespacedName) *aee.OpenStackAnsibleEE {
 	instance := &aee.OpenStackAnsibleEE{}
+	novaExt := GetNovaExternalCompute(novaNames.ComputeName)
```
gibizer, line 406: This make the GetAAE a bad candidate to moving into lib-common as a generic helper. I think we need to separate calls one that GetAEE and the other that asserts nova specific things on it.

Refactored it in https://github.com/openstack-k8s-operators/nova-operator/pull/412

SeanMooney: ack i agree although i did not think we were going to factor this out into lib common

ill add the refactor to my review list

gibizer: I'd like to move both GetAEE and the pulled out AssertControllerRef to lib-common as both useful for dataplane-operator testing as well.

PR #389, opened 2023-05-31
Use built-in function to evaluate ReadyCondition

kajinamit:
... instead of duplicating the logic.

### DIFF
--- a/api/v1beta1/nova_types.go
+++ b/api/v1beta1/nova_types.go
@@ -18,7 +18,6 @@ package v1beta1
 
 import (
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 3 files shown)


api/v1beta1/nova_types.go
```diff
@@ -165,9 +164,8 @@ func (s NovaStatus) GetConditions() condition.Conditions {
 }
 
 // IsReady returns true if Nova reconciled successfully
-func (c Nova) IsReady() bool {
-	readyCond := c.Status.Conditions.Get(condition.ReadyCondition)
-	return readyCond != nil && readyCond.Status == corev1.ConditionTrue
+func (instance Nova) IsReady() bool {
+	return instance.Status.Conditions.IsTrue(condition.ReadyCondition)
```
SeanMooney, line 168: what is the behaviour of IsTrue if the condition is nil
is nil treated as false

kajinamit: nil is treated as false. See https://github.com/openstack-k8s-operators/lib-common/blob/aca578f65ec99bebc30326c7ff1e5faca728656d/modules/common/condition/funcs.go#L138-L141

gibizer (approved):
Looks good to me.

PR #390, opened 2023-05-31
Watch Spec.Secret

gibizer:
If the content of the Secret pointed by the Nova*.Spec.Secret fields
changes then the given CR will be reconciled. If one of the relevant
passwords in that Secret is changed then the affected nova service
configs are regenerated and the CONFIG_HASH in the statefulsets are
updated to trigger service restart.

Closes: https://issues.redhat.com/browse/OSPRH-215

### DIFF
--- a/controllers/novaapi_controller.go
+++ b/controllers/novaapi_controller.go
@@ -24,8 +24,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	routev1 "github.com/openshift/api/route/v1"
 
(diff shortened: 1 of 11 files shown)


gibizer:
I did a quick measurement of how many times we call into the mapper function to map Secrets to reconcile functions. Starting the nova-operator in crc that has all the operators deployed but no openstack CRs created results in ~ 1000 calls for each Nova sub CRD Secret watcher (with 0 reconcile hit) so a total of ~ 4000 calls. 

Then run make openstack_deploy. That triggered an additional ~100 calls per CRD

The I run make edpm_compute_deploy. That triggered < 100 call per CRD. 

So a full deployment in an empty env generates 4500 - 5000 calls the the mapper. If there are no big Secret CRUD churn later then I only foresee issue at nova-operator startup as the ~90% of the calls happens that time.

controllers/novaapi_controller.go
```diff
@@ -622,5 +624,10 @@ func (r *NovaAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Owns(&routev1.Route{}).
 		Owns(&keystonev1.KeystoneEndpoint{}).
+		// NOTE(gibi): If watching for every secret is too much then
+		// we should consider switching to immutable Secret instead of watching
+		// mutable Secrets
```
gibizer, line 629: Let's move this comment to the definition of GetSecretMapperFor as that is at a common place. And amend it with some possible options:
* Use immutable secrets. But then we need to test what happens if the Secret is deleted and then recreated with the same name. If that does not trigger a reconcile then we still need a watch
* Move the watch to a central place (nova controller, or even openstack controller) and expose a "generation" field that the central component can bump to trigger a reconcile if the secret content changed.

gibizer:
Lets retry but It seem we started running more tempest test than we can

mrKisaoLamb (commented):
LGTM
I like this solution for now, even if it brings some risk. In my opinion, it's a "good enough" solution without too much overthinking. If we start experiencing issues as you mentioned, we can consider moving this pattern to a central place inside the operator and splitting it into smaller secrets. Then, we can use EnqueueRequestForOwner or even create an operator specifically designed to handle syncing smaller secrets in a more sophisticated manner.

bogdando:
I believe we should implement immutable secrets for that and drive the other openstack operators by Nova example.
There is a high risk that we'll face a thundering herd problem, if all operators start watching all secrets and increase the mapper function calls numbers in geomethrical progression

gibizer:
> I believe we should implement immutable secrets for that and drive the other Openstack operators by Nova example. There is a high risk otherwise that we'll face a thundering herd problem, if all operators start watching all secrets and increase the mapper function calls numbers in geometric progression

I mentioned this couple of times. I do want to try out immutable Secrets. But as a follow up of this work. I want to try it out as part of the smaller secrets created for nova sub CRs. That PR is still on my TODO list for Q2. I want to try immutable top of that as I foresee pros and cons with that approach that I want to show via a clear example. E.g. as soon as we create immutable secrets we need to re-create them, and therefore needs to know if the existing Secret is different from the Secret we need. This brings in extra complexity to track Secrets.

mrKisaoLamb (approved):
lgtm

PR #391, opened 2023-06-01
Remove DatabaseUser configurability

gibizer:
The current mariadb-operator ignores the provided database user name and always creates a user based on the name of the DB schema. So this PR removes the defunct database user configurability from our CRDs.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -648,7 +648,7 @@ func (r *NovaReconciler) ensureAPIDB(
 ) (*database.Database, nova.DatabaseStatus, error) {
 	apiDB := database.NewDatabaseWithNamespace(
 		nova.NovaAPIDatabaseName,
-		instance.Spec.APIDatabaseUser,
+		"unused",
 		instance.Spec.Secret,
 		map[string]string{
 			"dbName": instance.Spec.APIDatabaseInstance,
@@ -676,7 +676,7 @@ func (r *NovaReconciler) ensureCellDB(
 ) (*database.Database, nova.DatabaseStatus, error) {
 	cellDB := database.NewDatabaseWithNamespace(
 		"nova_"+cellName,
-		cellTemplate.CellDatabaseUser,
+		"unused",
 		instance.Spec.Secret,
 		map[string]string{
 			"dbName": cellTemplate.CellDatabaseInstance,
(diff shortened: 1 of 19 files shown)


gibizer:
There is a slack discussion about the future of the database user configurability. Let's wait for that to finish before we merge this.

controllers/nova_controller.go
```diff
@@ -1266,15 +1258,16 @@ func (r *NovaReconciler) ensureCellMapped(
 		"keystone_internal_url":  cell.Spec.KeystoneAuthURL,
 		"nova_keystone_user":     cell.Spec.ServiceUser,
 		"nova_keystone_password": string(ospSecret.Data[instance.Spec.PasswordSelectors.Service]),
-		// cell.Spec.APIDatabaseUser is empty for cells without APIDB access
-		"api_db_name":     instance.Spec.APIDatabaseUser, // fixme
-		"api_db_user":     instance.Spec.APIDatabaseUser,
+		"api_db_name":            nova.NovaAPIDatabaseName,
+		// mariadb-operator use the DB schema name as the user name
+		"api_db_user":     nova.NovaAPIDatabaseName,
```
bogdando, line 1263: but don't we remove api db user from all specs?

controllers/nova_controller.go
```diff
@@ -648,7 +648,7 @@ func (r *NovaReconciler) ensureAPIDB(
 ) (*database.Database, nova.DatabaseStatus, error) {
 	apiDB := database.NewDatabaseWithNamespace(
 		nova.NovaAPIDatabaseName,
-		instance.Spec.APIDatabaseUser,
+		"unused",
```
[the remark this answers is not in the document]
gibizer: > cannot use nil as string value in argument to NewDatabaseWithNamespace

controllers/nova_controller.go
gibizer: Yes, we removed them from our CRDs. But we still need them in the db connection url in our service configs. Do you suggest to change the config template and remove the *_db_user template parameter from it as we know that it is always the same as the value of the  *_db_name template parameter?

bogdando: yes, that would make sense, just leaving a note although

mrkisaolamb:
I think we can close this especially that https://github.com/openstack-k8s-operators/nova-operator/pull/681 is ongoing

gibizer:
good point, thanks

PR #396, opened 2023-06-02
Remote FIXME from envtest

gibizer:
The map_cell0 call moved from conductor dbsync job to a cell mapping job run from the nova controller. This happened while the envtest name handling refactor was in flight. So the FIXME is now resolved.


mrKisaoLamb (approved):
lgtm

PR #402, opened 2023-06-06
Validate lenght of cell name

mrKisaoLamb:
We need to limit the lenght of the name of the nova-cell as the operator generates names for the sub CRs from this name as a prefix. E.g. <nova-CR-name>-<cell name>-metadata-internal
K8s limits the name of a CR to 63 chars.

### DIFF
--- a/api/v1beta1/novacell_webhook.go
+++ b/api/v1beta1/novacell_webhook.go
@@ -97,12 +97,25 @@ func (r *NovaCell) ValidateMetadata() *field.Error{
 
 var _ webhook.Validator = &NovaCell{}
 
+// ValidateName validate cell name lenght
+func (r *NovaCell) ValidateName() *field.Error{
+	if len(r.Spec.CellName) > 35{
+		return field.Forbidden(field.NewPath("spec").Child("CellName"), "should be shorter than 35 characters")
+	}
+	return nil
+}
+
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *NovaCell) ValidateCreate() error {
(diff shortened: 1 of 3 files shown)


test/functional/suite_test.go
```diff
@@ -264,7 +264,7 @@ var _ = BeforeEach(func() {
 
 	novaName := types.NamespacedName{
 		Namespace: namespace,
-		Name:      uuid.New().String(),
+		Name:      uuid.New().String()[:25],
```
bogdando, line 267: why 25? shouldn't it be 35?

mrKisaoLamb: This namespace limit applies to the novaName and not the cellName. If we set this limit to 35, it will cause failures in tests that involve the cell because we add additional letters when creating the cellName.

SeanMooney (changes requested):

api/v1beta1/novacell_webhook.go
```diff
@@ -85,19 +88,49 @@ func (spec *NovaCellSpec) Default() {
 
 var _ webhook.Validator = &NovaCell{}
 
+// ValidateName validate cell name lenght
+func (r *NovaCell) ValidateName() *field.Error{
+	if len(r.Spec.CellName) > 35{
+		return field.Forbidden(field.NewPath("spec").Child("CellName"), "should be shorter than 35 signs")
```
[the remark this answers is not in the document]
mrKisaoLamb: thanks, done!

SeanMooney (approved):

SeanMooney (approved):
ah yes the ensure secret change has merged hence the rebase

still looks good to me

PR #408, opened 2023-06-16
update external dataplane samples

SeanMooney:
This change updtateds the exernal dataplane samples to work with
the recent change to use network isolation with crc and the
dataplane operator.


gibizer (approved):
Only a single tempest test failed. It is know to be failing.

PR #409, opened 2023-06-16
vnc fixups

SeanMooney:
- [noVNCProxy] Add basic reconciler and templates
- Add tests for NoVNCProxy
- [noVNCProxy] Add deployment in cell and ignore cell0
- [noVNCProxy]Add missing ServiceAccount and transport url
- [noVNCProxy]Change kolla config and nova.conf
- [noVNCProxy] Change order in reconcile loop
- Rename APIMessageBusSecretName for NoVNCProxy
- [noVNCProxy] Remove APIEndpoints from novnc controller
- use route to connect to vnc proxy.


templates/nova.conf
```diff
@@ -117,13 +117,18 @@ driver = noop
...
+novncproxy_host = "::0"
+novncproxy_port = 6080
+server_listen = "::0"
+{{ if eq .service_name "nova-compute"}}
+{{/* https://docs.openstack.org/oslo.config/latest/configuration/format.html#substitution    */}}
+# note we may want to use console_host instead of my_ip however it wont be resolved via
+# dns currently so we need to use my_ip for now.
+# https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.console_host
+server_proxyclient_address = "$my_ip"
+novncproxy_base_url = {{ .novncproxy_base_url }}/vnc_auto.html
```
SeanMooney, line 131: we should use vnc_lite.html not vnc_auto.html based on 

https://opendev.org/openstack/nova/src/branch/master/releasenotes/notes/deprecate-keymap-options-b41ad9f33a5923e1.yaml#L30-L34

its remvoed on master.

SeanMooney:
This was just to provide an example of what was needed to fix the config generation.

these changes have been encopretated into the main PR so closing this

PR #410, opened 2023-06-19
[NovaConductor]Use already read Secret

gibizer:
The #304 missed propagating the already read secret to the conductor config gen. This is fixed now.

### DIFF
--- a/controllers/novaconductor_controller.go
+++ b/controllers/novaconductor_controller.go
@@ -137,7 +137,7 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		required_secret_fields = append(required_secret_fields, instance.Spec.PasswordSelectors.APIDatabase)
 	}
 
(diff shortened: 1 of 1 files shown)


SeanMooney (approved):
ah yes that was indeed missed in the previous patch

controllers/novaconductor_controller.go
```diff
@@ -275,7 +276,11 @@ func (r *NovaConductorReconciler) ensureConfigMaps(
 }
 
 func (r *NovaConductorReconciler) generateConfigs(
-	ctx context.Context, h *helper.Helper, instance *novav1.NovaConductor, hashes *map[string]env.Setter,
+	ctx context.Context,
+	h *helper.Helper,
+	instance *novav1.NovaConductor,
+	hashes *map[string]env.Setter,
+	secret corev1.Secret,
```
SeanMooney, line 283: nit i really dont like flattening the args like this.

to mee ti make this much less readable

PR #411, opened 2023-06-19
Refactor validation webhook for cell

gibizer:
This is needed if we want to reject the API request at Nova CR creation. Also this will allow us to hook into the validation logic from openstack-operator as well see https://github.com/openstack-k8s-operators/openstack-operator/pull/355

This fixed the field descriptors to follow the yaml field names. This also fixed the Kind returned in the error message to match with the Kind being created.

### DIFF
--- a/api/v1beta1/nova_webhook.go
+++ b/api/v1beta1/nova_webhook.go
@@ -23,7 +23,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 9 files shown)


api/v1beta1/nova_webhook.go
```diff
@@ -104,19 +107,69 @@ func (spec *NovaSpec) Default() {
...
+
+	return errors
+}
+
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *Nova) ValidateUpdate(old runtime.Object) error {
 	novalog.Info("validate update", "name", r.Name)
 
-	// TODO(user): fill in your validation logic upon object update.
+	errors := r.Spec.ValidateCreate(field.NewPath("spec"))
```
abays, line 166: Should this be `r.Spec.ValidateUpdate` instead of `ValidateCreate`?

SeanMooney: perhaps although right now i think the validation we will need to do on update and create is the same so perhaps gibi is just reusing the same function.


there might be some filed we will not want to all updates too after the initall create but i dont think we currently have any that fall into that catagory today.

api/v1beta1/novacell_webhook.go
```diff
@@ -159,3 +142,31 @@ func (r *NovaCell) ValidateDelete() error {
 	// TODO(user): fill in your validation logic upon object deletion.
 	return nil
 }
+
+// ValidateCell0 validates cell0 Metadata template. This is expected to be called
+// by higher level validation webhooks
+func (r *NovaMetadataTemplate) ValidateCell0(basePath *field.Path) field.ErrorList {
```
SeanMooney, line 148: is this in the correct file.
should we move it to be located with the NovaMetadataTemplate defintion

test/functional/validation_webhook_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+		Expect(err).Should(HaveOccurred())
+		statusError, ok := err.(*errors.StatusError)
+		Expect(ok).To(BeTrue())
+		Expect(statusError.ErrStatus.Details.Kind).To(Equal("Nova"))
+		Expect(statusError.ErrStatus.Message).To(
+			ContainSubstring(
+				"invalid: spec.cellTemplates[cell0].metadataServiceTemplate.replicas: " +
+					"Invalid value: 1: should be 0 for cell0"),
+		)
```
SeanMooney, line 66: +1

this is quite a nice error to recive

test/functional/validation_webhook_test.go
```diff
@@ -0,0 +1,204 @@
...
+				Field:   "spec.cellTemplates[cell0].metadataServiceTemplate.replicas",
+			}),
+		)
+		Expect(statusError.ErrStatus.Details.Causes).To(
+			ContainElement(metav1.StatusCause{
+				Type: "FieldValueInvalid",
+				Message: "Invalid value: \"cell1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\": " +
+					"should be shorter than 36 characters",
+				Field: "spec.cellTemplates[cell1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx]",
+			}),
```
SeanMooney, line 201: ok nice so as a user you wont need to fix the cr just to get the next error and have to fix it again

you might fix it in a way that causes other error but atleast we are trying to provide a best effort validation of all the possible issues.

nice test coverage :)

api/v1beta1/nova_webhook.go
gibizer: @abays It should. I will change this to be symmetric. While we have not difference today between ValidateUpdate and ValidateCreate we can have that in the future.

api/v1beta1/novacell_webhook.go
gibizer: yeah, I will move it

gibizer: Done

api/v1beta1/nova_webhook.go
gibizer: Done

SeanMooney (approved):

PR #412, opened 2023-06-20
Decouple AssertControllerRef from GetAEE

gibizer:
This way both can be moved to lib-common and reused in other operators like dataplane.


gibizer:
recheck

SeanMooney (approved):

Issue #294, opened 2023-03-18
ExternalCompute controller re-runs the ansible deplyoment steps after each nova-operator restart


gibizer:
It seems the current reconciler re-runs the deployment steps after each nova-operator restart.
To reproduce:
1. create NovaExternalCompute and wait until it becomes ready after it run the ansible deploy steps
2. delete the nova-operator-controller-manager pod to trigger a restart of nova-operator

Expected behavior:
If nothing is changed in the NovaExternalCompute resource then the ansible code is not re-run

Actual behavior:
The ansible code is re-run each time.

While re-running ansible should not cause any trouble as we expect that the ansible code is idempotent it will still causes many 2N ansible execution in an environment with N compute for each nova-operator restart.

SeanMooney:
yep this is something i was aware of.
for the initial pr i kept this the same as all the other reconsile loop which also rerun on operator restart.

but i agree we should special case this in a follow up.

gibizer:
This is targeting NovaExternalCompute but that is unused now and will be deleted by https://github.com/openstack-k8s-operators/nova-operator/pull/476

PR #418, opened 2023-06-26
Revert "Do not left behind AnsibleEE job pods"

gibizer:
This reverts commit 48c9264e2c3393dd796bd987620a2cb0c69771e9.

That commit added controller reference from NovaExternalCompute to the AnsibleEE CR so that those CRs are cleaned up properly when NovaExternalCompute is deleted. However it has a side effect that NovaExternalCompute will be reconciled *after* the AnsibleEE is deleted. So in the happy path when the deploy succeeded and the nova-operator deleted the AnsibleEE there will be an extra reconcile on the NovaExternalCompute due to the controller reference. There is a know issue #294 in nova-operator that the successful run of AnsibleEE is not persisted properly. Therefore such extra reconcile now triggered a new AnsibleEE creation causing an infinite deploy loop.

We do need the controller reference for cleanup. But we cannot have it until #294 is fixed.

Note that there is a second revert in this PR that reverts the test refactor happened on top of 48c9264e2c3393dd796bd987620a2cb0c69771e9

### DIFF
--- a/controllers/novaexternalcompute_controller.go
+++ b/controllers/novaexternalcompute_controller.go
@@ -592,10 +592,6 @@ func (r *NovaExternalComputeReconciler) ensureAEEDeployLibvirt(
 		_, err = controllerutil.CreateOrPatch(ctx, h.GetClient(), ansibleEE, func() error {
 			initAEE(instance, ansibleEE, "deploy-libvirt.yaml")
 
-			err = controllerutil.SetControllerReference(instance, ansibleEE, h.GetScheme())
-			if err != nil {
-				return err
-			}
 			return nil
 		})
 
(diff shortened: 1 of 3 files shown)


gibizer:
I verified it locally that by reverting this the EDPM deploy loop is not happening any more.

SeanMooney (approved):
let me fix this when i work on the other cleanups
so ya lets revert this for now

mrKisaoLamb (approved):
lgtm

bogdando:
Related https://github.com/openstack-k8s-operators/nova-operator/issues/319

bogdando:
Also related https://github.com/openstack-k8s-operators/nova-operator/issues/296

Issue #422, opened 2023-06-26
The nova external compute reconsiler is useing the wrong secrete for the config generation


SeanMooney:
as of https://github.com/openstack-k8s-operators/nova-operator/commit/ab34bb06a8b2cb16fbb371cbfc8afd4b8bbe92ca

the edpm ssh key secrete is being passed to the config gen function instead of the instance CR secret

https://github.com/openstack-k8s-operators/nova-operator/commit/ab34bb06a8b2cb16fbb371cbfc8afd4b8bbe92ca#diff-195a08f97364d8db1117085a1fe611eb7a17542c70dd06f224e439bdc2ddd439R161-R194

as a result, the nova keystone user password is not set in the nova.conf in compute nodes.

PR #423, opened 2023-06-26
Use cell secret for compute config generation

SeanMooney:
- reporduce bug 422
- use cell secret for nova external compute config generation

Closes: #422


mrKisaoLamb (approved):
lgtm

PR #427, opened 2023-06-29
[envtest] Fix k8s conflict handling logic

gibizer:
We used
```
Eventualy(func(){
...
  g.Expect(err == nil || k8s_errors.IsConflict(err)).To(BeTrue())
  ...
})...
```
to say that we want to retry on conflict. But actually this will report success when conflict happens. So in the rare case when that conflict happens during test execution the test moves forward instead of retrying. This is fixed now.

### DIFF
--- a/test/functional/nova_metadata_controller_test.go
+++ b/test/functional/nova_metadata_controller_test.go
@@ -25,7 +25,6 @@ import (
 	. "github.com/openstack-k8s-operators/lib-common/modules/test/helpers"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


gibizer:
Example failure in the CI https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/419/pull-ci-openstack-k8s-operators-nova-operator-main-functional/1674410708822921216/build-log.txt

SeanMooney (approved):
nice find. the fixed code is also shorter which is nice.

PR #428, opened 2023-06-30
Align to the validation webhook pattern

gibizer:
The pattern is documented in https://github.com/openstack-k8s-operators/docs/pull/47

### DIFF
--- a/api/v1beta1/nova_webhook.go
+++ b/api/v1beta1/nova_webhook.go
@@ -23,6 +23,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


api/v1beta1/nova_webhook.go
```diff
@@ -162,8 +164,12 @@ func (r *NovaSpec) ValidateUpdate(old runtime.Object, basePath *field.Path) fiel
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *Nova) ValidateUpdate(old runtime.Object) error {
 	novalog.Info("validate update", "name", r.Name)
+	oldNova, ok := old.(*Nova)
```
SeanMooney, line 167: im not quite sure what htis is doing.

if i pretended this was C++ then i would say you are using the brackets to fix precedence so oldNova is the dereference result of following the point in old

i.e. the same as old->Nova where old.nova would be a pointer holding the address old->Nova would be the object pointed to by it

what does old.(*Nova) do in go.

it also partly looks like a cast but its on the wrong side is this the same as 

Nova *oldNova = (*Node) old
in c++

mrKisaoLamb: AFAIK, this is a type assertion. If old is of type Nova, it would look something like this: https://go.dev/tour/methods/15 . It's useful to avoid throwing a panic and instead return an error of your choice.

mrKisaoLamb (commented):
lgtm

api/v1beta1/nova_webhook.go
gibizer: yepp it is cast. It should never fail though

SeanMooney (approved):

PR #430, opened 2023-07-01
add inital molecule setup.

SeanMooney:
This patch uses molecule to deploy nova and libvirt
via the in tree palybooks on vagrant vms using the
openstack-ansibleee-runner images.
Related: [OSPRH-229](https://issues.redhat.com//browse/OSPRH-229)

### DIFF
--- a/molecule/default/INSTALL.rst
+++ b/molecule/default/INSTALL.rst
@@ -1,0 +1,23 @@
+*********************************
+Vagrant driver installation guide
+*********************************
+
+Requirements
+============
+
+* Vagrant
+* Virtualbox, Parallels, VMware Fusion, VMware Workstation or VMware Desktop
+
+Install
+=======
+
+Please refer to the `Virtual environment`_ documentation for installation best
+practices. If not using a virtual environment, please consider passing the
+widely recommended `'--user' flag`_ when invoking ``pip``.
+
+.. _Virtual environment: https://virtualenv.pypa.io/en/latest/
+.. _'--user' flag: https://packaging.python.org/tutorials/installing-packages/#installing-to-the-user-site
+
+.. code-block:: bash
+
+    $ pip install 'molecule-plugins[vagrant]'
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -1,0 +1,100 @@
+
+- name: Setup Test Environment
+  hosts: AEE
+  tasks:
+    - name: Copy playbooks to AEE vm
+      ansible.builtin.copy:
+        src: "{{ item.src }}"
+        dest: "{{ item.dest }}"
+        mode: "{{ item.mode | default('640') }}"
+      with_items:
+      - { "src": "../../playbooks/deploy-libvirt.yaml",
+          "dest": "~/deploy-libvirt.yaml"
+        }
+      - { "src": "../../playbooks/deploy-nova.yaml",
+          "dest": "~/deploy-nova.yaml"
+        }
+    - name: Copy molecule files
+      ansible.builtin.copy:
+        src: "{{ item.src }}"
+        dest: "{{ item.dest }}"
+        mode: "{{ item.mode | default('600') }}"
+      with_items:
+      - { "src": "{{ lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}/.vagrant/machines/compute-1/libvirt/private_key",
+          "dest": "~/private_key"
+        }
+    - name: install packages
+      become: true
+      ansible.builtin.package:
+        name:
+          - podman
+          - rsync
+        state: present
+    - name: secret config dir
+      become: true
+      ansible.builtin.file:
+        path: "{{ item.path }}"
+        state: directory
+        setype: "{{ item.setype | default('container_file_t') }}"
+        owner: "{{ item.owner | default(ansible_user) }}"
+        group: "{{ item.group | default(ansible_user) }}"
+        mode: "{{ item.mode | default('750') }}"
+        recurse: true
+      with_items:
+      - {"path": /var/lib/openstack/config/}
+    - name: Copy playbooks to AEE vm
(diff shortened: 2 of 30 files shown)


playbooks/deploy-libvirt.yaml
```diff
@@ -14,6 +14,7 @@
       with_items:
       - { "path": /var/lib/openstack/config/libvirt}
       - { "path": /var/lib/openstack/config/containers}
+      - { 'path': /var/lib/nova}
```
SeanMooney, line 17: note we actully had a race condtion between the nova and libvirt playbooks which this found.

libvirt need access to /var/lib/nova for the virtqemud container

playbooks/deploy-nova.yaml
```diff
@@ -31,6 +31,8 @@
       # these should be done in bootstrap
       - { 'path': /var/log/containers/stdouts, 'mode': '0750' }
       - { "path": /etc/ceph, "mode": "0750", "owner": "root", "group": "root"}
+      - { "path": /run/openvswitch, "mode": "0777"}
```
SeanMooney, line 34: and the nova compute container need access to /run/openvswitch if its configured to connect with the vsctl client instead of python.

in a normal env this will be created via the dataplane operator Bootstrap 
playbook so I'm wondering if i should emulate this in the molecules converge playbook instead?

molecule/default/converge.yml
```diff
@@ -0,0 +1,100 @@
...
+      ansible.builtin.file:
+        path: /etc/localtime
+        src: /usr/share/zoneinfo/UTC
+        state: link
+
+
+- name: Run Playbooks
+  hosts: AEE
+  tasks:
+    - name: disable selinux
```
SeanMooney, line 73: selinux is disabled otherwise the playbooks and the other bindmoutns will return a permission denied even as root because of selinux blocking access.

I'm not really sure of a better solution for now
so i just disable it.

molecule/default/converge.yml
```diff
@@ -0,0 +1,100 @@
...
+          - podman
+          - rsync
+          - python-pyyaml
+        state: present
+    - name: set /etc/localtime
+      become: true
+      ansible.builtin.file:
+        path: /etc/localtime
+        src: /usr/share/zoneinfo/UTC
+        state: link
```
SeanMooney, line 67: see note later but i could create /run/openvswitch or install ovs here
instead.

molecule/default/molecule.yml
```diff
@@ -0,0 +1,34 @@
...
+      - oepnshift
+  - name: compute-1
+    memory: 6000
+    cpus: 4
+    provider_options:
+      cpu_mode: 'host-passthrough'
+      nested: true
+      machine_type: 'q35'
+    groups:
+      - compute
```
SeanMooney, line 32: i could proably make both of these smaller by the way.
i need to ssh in and check the used ram but i would be surprised if i could not fit both in 4G 1G AEE and 3G compute.
perhaps even smaller.

molecule/default/converge.yml
```diff
@@ -0,0 +1,100 @@
...
+        -e ANSIBLE_SSH_ARGS="-o StrictHostKeyChecking=no" quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest \
+        ansible-playbook -i 'compute-1,' --key-file /runner/private_key -u vagrant  /runner/deploy-libvirt.yaml
+    - name: run deploy-nova.yaml with podman
+      ansible.builtin.command: |
+        podman run --net host --rm -v /var/lib/openstack/config/:/var/lib/openstack/config/ \
+        -v ~/deploy-nova.yaml:/runner/deploy-nova.yaml -v ~/private_key:/runner/private_key -e ANSIBLE_HOST_KEY_CHECKING=False \
+        -e ANSIBLE_REMOTE_USER=vagrant -e ANSIBLE_SSH_PRIVATE_KEY_FILE=/runner/private_key \
+        -e ANSIBLE_SSH_ARGS="-o StrictHostKeyChecking=no" quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest \
+        ansible-playbook -i 'compute-1,' --key-file /runner/private_key -u vagrant  /runner/deploy-nova.yaml
+    # - name: create ansible python venv
```
SeanMooney, line 93: this is the other way i could have run this.
its actually what i started with and i reverted to using podman because i could not get past a yaml module error.

turns out it was unrelated to this and i just needed to install pyyaml on the compute VM.

so I'm not sure if i want to just keep this for debugging or if i should remove this.

molecule/default/converge.yml
mrKisaoLamb: I don't have any better idea, I think similar solution is in tripleo-ansible molecule tests

molecule/default/converge.yml
mrKisaoLamb: If we want keep this probably some comment about it 'why' we keep it will be nice

mrKisaoLamb (commented):
lgtm but didn't run yet

molecule/default/converge.yml
SeanMooney: done

molecule/default/converge.yml
SeanMooney: i removed it

molecule/default/INSTALL.rst
```diff
@@ -0,0 +1,23 @@
+*********************************
```
SeanMooney, line 1: by the way this is the autogetnerted file form molecule i will come back in a follow up and update this with the actual instructions seperately

SeanMooney: it is something like this https://github.com/SeanMooney/ard/blob/master/bootstrap-repo.sh but i had previously run all or path of that so my laptop was not clean i need to do this in a clean VM to document the actual requirements.

gibizer (approved):

PR #432, opened 2023-07-04
update .gitignore to ignore debug binary

SeanMooney:


gibizer (approved):
The tempest failure is not relevant

PR #435, opened 2023-07-05
Remove endpoint url and keystone service id from status

kajinamit:
These items presented in CR status are not really used. Endpoints can be discovered using Keystone and we have no use case to require service id.

### DIFF
--- a/api/v1beta1/novaapi_types.go
+++ b/api/v1beta1/novaapi_types.go
@@ -153,14 +153,8 @@ type NovaAPIStatus struct {
 	// Map of hashes to track e.g. job status
 	Hash map[string]string `json:"hash,omitempty"`
 
-	// API endpoint
-	APIEndpoints map[string]string `json:"apiEndpoint,omitempty"`
-
 	// Conditions
 	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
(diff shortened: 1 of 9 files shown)


api/v1beta1/novaapi_types.go
```diff
@@ -153,15 +153,9 @@ type NovaAPIStatus struct {
...
 
-	// API endpoint
-	APIEndpoints map[string]string `json:"apiEndpoint,omitempty"`
-
 	// Conditions
 	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
 
-	// ServiceID - the ID of the registered service in keystone
-	ServiceID string `json:"serviceID,omitempty"`
-
```
SeanMooney, line 164: tentiitve +1 on this

let me check with gibi later today if we have any plans to use this our selves for anything but i agree it can likely be remvoed.

kajinamit: The only exception I've found so far is glance which uses this to manage limit resources in keystone.
In case you plan to implement it then we can leave it now or add it back when you actually work on it.

gibizer: I'm OK to remove it.

PR #436, opened 2023-07-05
Fix API and metadata worker counts

gibizer:
We try to set metadata_workers 0 for nova-api and osapi_compute_workers to 0 for nova-metadata service. However both option defines that the minimum value is 1. So currently both the nova-api and nova-metadata services reporting config validation error and only partially parse the config files.

Instead of setting the worker count to 0 we can use enabled_apis to define which api service running by which binary. So this patch changes the config template to do so.


mrKisaoLamb (approved):
lgtm

PR #437, opened 2023-07-05
Add test coverage for worker count

gibizer:
The fix was in 359004c3763cd474cf2de2e419ba3f50b82d9b2f but we missed the test coverage for it.


SeanMooney (approved):
looks good to me

PR #438, opened 2023-07-05
Add container volume mounts required for iscsi volumes

ASBishop:
Add volume mounts to the nova_compute container that are necessary for handling iscsi volumes attachments.

### DIFF
--- a/templates/novacompute/nova_compute.json
+++ b/templates/novacompute/nova_compute.json
@@ -19,6 +19,8 @@
         "/var/log/containers/nova:/var/log/containers/nova",
         "/var/lib/libvirt:/var/lib/libvirt",
         "/var/lib/nova/:/var/lib/nova/",
+        "/var/lib/iscsi:/var/lib/iscsi",
+        "/etc/iscsi:/etc/iscsi:ro",
         "/etc/ceph:/var/lib/openstack/config/ceph:ro"
     ]
 }


templates/novacompute/nova_compute.json
```diff
@@ -19,6 +19,8 @@
         "/var/log/containers/nova:/var/log/containers/nova",
         "/var/lib/libvirt:/var/lib/libvirt",
         "/var/lib/nova/:/var/lib/nova/",
+        "/var/lib/iscsi:/var/lib/iscsi",
```
SeanMooney, line 22: do we need :z here for selinux relableing

https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/nova/nova-compute-container-puppet.yaml#L1416C27-L1416C27


also do you need these multipath folders
https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/nova/nova-compute-container-puppet.yaml#L1424-L1425C69

ASBishop: I've always been fuzzy on the selinux labelling requirements. That said, there's no need to share the contents of /var/lib/iscsi between multiple containers. nova_compute needs access in order to properly manage its own connections, but no other containers need access.

Regarding multipath, I'll update the mounts after I update edpm_ansible to deploy multipathd (it's not deployed at all in the current code).

ASBishop: I should have mentioned that this patch got iscsi working for me. I have CRC and edpm compute node vms running on a machine in my office, and with this patch I can create and attach an iscsi volume on our NetApp SAN in TLV to an instance on the edpm compute node.

SeanMooney: cool in that case lets proceed with this as is and we can extend it later

thanks for the PR

SeanMooney (approved):

gibizer:
VM boot works for me with this patch too so it did not regress anything. Thanks @ASBishop

PR #440, opened 2023-07-06
Fix /var/lib/nova permissions in libvirt playbook

gibizer:
Both nova-compute and libvirt depends on the existence of /var/lib/nova so we added the creation of it to the libvirt playbook too (it was in the nova-compute playbook from the start), as in some tests the libvirt playbook reached this point before the nova-compute one. However the fix in bf18bbe5d5ac1bb836ae6ad1df99bcecd7b49540 was incomplete as it set the directory permission to 0750 if it was created by the libvirt playbook. This caused that the VM boot failed with permission error as the dir should have 0755 access.


gibizer:
Did an edpm deploy with this PR and the VM boot works.

SeanMooney (approved):
thanks gibi great find

PR #441, opened 2023-07-06
This change fixes binding mounting for virtlogd and qemud

SeanMooney:
This change fixes two issues with how the virtlogd and qemud
contaienrs were run.

First /var/log/libvirt/qemu is now
bindmoutned form virtlogd to /var/log/containers/qemu to expose
the qemu instances logs and persit them across reboots.

Second teh qemu.conf that is generated for the qemud contaienr
is now actully copied into place and used. As part of fixing this
i also fix a syntax error in the vnc_listen option as 0.0.0.0 was
invalid. This vnc_listen is not required as that is contoled
by nova and it works without this option being set so it is
simply removed.

i have also commented out the configuration of the migration port
range but left it in the file since we currently dont supprot
migration and we do not open any ports for it in the firewall.
We will adress this end to end in the future.

### DIFF
--- a/playbooks/deploy-libvirt.yaml
+++ b/playbooks/deploy-libvirt.yaml
@@ -48,6 +48,7 @@
       with_items:
       - { "path": /var/lib/libvirt}
       - { "path": /var/log/containers/libvirt, "mode": "0750" }
+      - { "path": /var/log/containers/qemu, "mode": "0750" }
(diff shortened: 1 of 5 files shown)


gibizer (approved):
The nova part of the deployment job succeeded cinder and neutron was progressing when job timed out. So I ignoring that job failure.

I deployed EPDM with this change and both issue is fixed now.

PR #443, opened 2023-07-07
Do not set transport_url for cell0 mapping

gibizer:
The cell0 mapping should never have transport_url set as there are no computes in cell0 that would use such message bus.

When we introduced a separate cell mapping job for each cell we regressed the cell mapping setup and cell0 is mapped with the API message bus transport_url. This causes that in a collapsed cell setup where only a single MQ is used for both API and cell1 the cell1 mapping job fails as the transport_url for cell1 is already used for cell0.

Closes: https://issues.redhat.com/browse/OSPRH-233

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -1324,7 +1324,15 @@ func (r *NovaReconciler) ensureCellMapped(
 		"openstack_region_name":  "regionOne", // fixme
 		"default_project_domain": "Default",   // fixme
 		"default_user_domain":    "Default",   // fixme
-		"transport_url":          string(mqSecret.Data["transport_url"]),
+	}
+
+	// NOTE(gibi): cell mapping for cell0 should not have transport_url
+	// configured. As the nova-manage command used to create the mapping
+	// uses the transport_url from the nova.conf provided to the job
+	// we need to make sure that transport_url is only configured for the job
+	// if it is mapping other than cell0.
+	if cell.Spec.CellName != novav1.Cell0Name {
+		templateParameters["transport_url"] = string(mqSecret.Data["transport_url"])
 	}
 
 	cms := []util.Template{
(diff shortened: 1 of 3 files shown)


gibizer:
Before the fix:
```
$ oc rsh nova-cell1-conductor-0
sh-5.1# nova-manage cell_v2 list_cells
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
|  Name |                 UUID                 |                               Transport URL                                |                 Database Connection                  | Disabled |
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
| cell0 | 00000000-0000-0000-0000-000000000000 | rabbit://default_user_8fMfRbdZ0_hJD5icv7K:****@rabbitmq.openstack.svc:5672 | mysql+pymysql://nova_cell0:****@openstack/nova_cell0 |  False   |
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
```
After the fix:
```
$ oc rsh nova-cell1-conductor-0
sh-5.1# nova-manage cell_v2 list_cells
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
|  Name |                 UUID                 |                               Transport URL                                |                 Database Connection                  | Disabled |
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
| cell0 | 00000000-0000-0000-0000-000000000000 |                                  rabbit:                                   | mysql+pymysql://nova_cell0:****@openstack/nova_cell0 |  False   |
| cell1 | 0b783897-d6e8-48b3-8cc7-e4973a8e0cc2 | rabbit://default_user_VlnE5zUOd1q4_De_A-0:****@rabbitmq.openstack.svc:5672 | mysql+pymysql://nova_cell1:****@openstack/nova_cell1 |  False   |
+-------+--------------------------------------+----------------------------------------------------------------------------+------------------------------------------------------+----------+
```

abays (commented):
+1 from me

SeanMooney (approved):
yep change looks correct and it includes envtest coverage

PR #445, opened 2023-07-10
Add DNSConfig property for NovaExternalCompute

rabi:
This would allow us to set the dnsserver details that ansibleee pods can use to resolve external computes.


SeanMooney:
/hold 
holding as i am not sure we should merge this before the dev preview so I'm holding it until we discuss that.

controllers/novaexternalcompute_controller.go
```diff
@@ -698,6 +698,7 @@ func initAEE(
 	playbook string,
 ) {
 	ansibleEE.Spec.Image = instance.Spec.AnsibleEEContainerImage
+	ansibleEE.Spec.DNSConfig = instance.Spec.DNSConfig
```
SeanMooney, line 701: the changes are pretty minimal but I think I would like
env-test coverage 

basically just assert that that if dns config is provided its propagated to the AEE CR.

rabi:
> /hold holding as i am not sure we should merge this before the dev preview so I'm holding it until we discuss that.

Not sure what you mean. This won't impact anything as long as you don't use dns service ie. provide ansibleHost IP for the compute nodes. But nova deploy won't work when we dns is used[1]. 

[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml#L10

rabi:
Also it has to be first wired-up in dataplane-operator.

SeanMooney:
> > /hold holding as i am not sure we should merge this before the dev preview so I'm holding it until we discuss that.
> 
> Not sure what you mean. This won't impact anything as long as you don't use dns service ie. provide ansibleHost IP for the compute nodes. But nova deploy won't work when we dns is used[1].
> 
> [1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml#L10

what i mean is we have imposed a soft feature freeze in the operator. and are being selective in what we merged between now and the dev preview release. we could include this but i want to check with gibi first. and yes today we require ansibleHost to be defied for all compute nodes. dns registration is currently optional since its not supported by the dataplane operator yet to my knolwsage so we cannot depend on it. also the host name field in the node may or may not align to the FQDN hostname used for the ansible connection. the current hostname filed is meant to be the hostname set on the node that will be used for OpenStack but the management network used for ansibel might not be the same as the console plane network so even with DNS in place we probably should still prefer to use ansibleHost although we could fall back to hostname if not defiend.

https://github.com/openstack-k8s-operators/nova-operator/blob/main/config/samples/external-compute/openstackdataplanenode.yaml#L8
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_1.yaml#L8
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml#L12

rabi:
> we probably should still prefer to use ansibleHost although we could fall back to hostname if not defiend.
That's what we do atm. so there is no impact.

api/v1beta1/novaexternalcompute_types.go
```diff
@@ -80,6 +81,10 @@ type NovaExternalComputeSpec struct {
 	// +kubebuilder:validation:Optional
 	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
 	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// DNSConfig allows to specify custom dnsservers and search domains
+	// +kubebuilder:validation:Optional
+	DNSConfig *corev1.PodDNSConfig `json:"dnsConfig,omitempty"`
```
gibizer, line 87: OK, we need omitempty due to the nested struct, but the field type is a pointer, so nil can be used as the unset value.

gibizer:
> 
> what i mean is we have imposed a soft feature freeze in the operator. and are being selective in what we merged between now and the dev preview release. we could include this but i want to check with gibi first. 

I will do a deployment with this patch included to be sure, but I agree that the change is fairly minimal and at the moment the new field is unused by the dataplane-operator. 

I more concerned about the direction change coming from https://issues.redhat.com/browse/OSP-25812 . If in the new architecture the nova-operator won't create AnsibleEE CR directly but only provide input to the dataplane-operator, then nova-operator probably does not need to know about the IP / FQDN of to be used to connect to the EDPM node to run ansible on it.

rabi:
> If in the new architecture the nova-operator won't create AnsibleEE CR directly but only provide input to the dataplane-operator

Right @gibbi, I did think about that angle before proposing the PR. However, currently dataplane can be deployed using ipam/dns but novaextracompute won't work without ansibleHost(IP) property. My thought was we can scrap it if and when everything goes away.

gibizer:
> I deployed compute services with this PR. But I see port binding failure when trying to boot a VM. Neutron says:
> 
> ```
> 2023-07-10 14:26:41.364 17 WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.mech_driver [req-93cf904c-c65b-4d9b-b2e5-e40cec321827 req-3b843128-0adf-4a2e-88ce-7d240f9a7141 919d4bf213884389bf58b48acb572db2 9da06589c4604c89ba14a0b4333cf84b - - default default] Refusing to bind port f27ae81b-8d9b-4608-9573-8bc8a14e8867 due to no OVN chassis for host: edpm-compute-0
> ```
> 
> I seems unrelated so I will do another deploy without this patch to see if it is an issue on main.

I was able to reproduce this issue without the current PR so it is unrelated.

rabi:
Looks like we are keen on blocking this in anticipation of architectural changes in Q3. If this does not make to tech preview then nova deploy can't use dns service like rest of the stuff.

gibizer:
> Looks like we are keen on blocking this in anticipation of architectural changes in Q3. If this does not make to tech preview then nova deploy can't use dns service like rest of the stuff.

I haven't merged this as I don't see what should work after this. Could you look at my comment at https://github.com/openstack-k8s-operators/nova-operator/pull/445#issuecomment-1629063670 and help me understand what is the expected behavior.

SeanMooney:
it will allow you to not set ansible_host to an IP

this change is configuring the DNS server used for DNS registration of the compute nodes which is not currently fully integrated in the dataplane operator.

it quires you to manually create the DNS entries i belvie 

but if you were to do that then it would allow you to simple set the hostname in the node cr without setting ansible_host or you could set ansible_host to an fqdn

this capability was never express as one that was intended to be supported in dev preview.
and without the integration in the dataplane operator to automatically registry nodes in dns its incomplete but it can be leverage manually if desired.

gibizer:
> and without the integration in the dataplane operator to automatically registry nodes in dns its incomplete but it can be leverage manually if desired.

Based on the printout in my comment https://github.com/openstack-k8s-operators/nova-operator/pull/445#issuecomment-1629063670 I believe there is compute DNS registration happening automatically. But still I don't see that I can use such DNS names from the control plane. What am I missing?

SeanMooney:
i guess https://github.com/openstack-k8s-operators/dataplane-operator/pull/231 
has merged however it did not update the datapalne samples to show how to use it

instead a followup https://github.com/openstack-k8s-operators/dataplane-operator/commit/cf93bf7fd3c895f58ff081193dcd362f54324c32

removed the etc/hosts config

with the proposed change 
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenode_edpm_compute_0.yaml#L8
would technically not be required any more

hostName: edpm-compute-0
or 
hostName: edpm-compute-0.my.fqdn

would be sufficient to allow ansibel to connect to the node

there is already a fallback to using hostName if ansible-host is not set.

rabi:
> > Looks like we are keen on blocking this in anticipation of architectural changes in Q3. If this does not make to tech preview then nova deploy can't use dns service like rest of the stuff.
> 
> I haven't merged this as I don't see what should work after this. Could you look at my comment at [#445 (comment)](https://github.com/openstack-k8s-operators/nova-operator/pull/445#issuecomment-1629063670) and help me understand what is the expected behavior.

@gibizer That should work unless computes can't reach dns service for some reason. I'll deploy and send paste the output of the same check later.  After this change and it's wired in dataplane-operator, nova ansbleee pods can reach compute nodes using hostname (with dns lookup) rather than ips. See the PR https://github.com/openstack-k8s-operators/install_yamls/pull/408, which is blocked by this[1].

[1] https://logserver.rdoproject.org/08/408/6f75d3bd581c2e405848e745d1178b25f7b11d2e/github-check/install-yamls-crc-podified-edpm-baremetal/f6440fd/ci-framework-data/logs/edpm/pods/openstack/nova-edpm-compute-0-deploy-libvirt-ssch7-logs.txt

rabi:
> and without the integration in the dataplane operator to automatically registry nodes in dns its incomplete but it can be leverage manually if desired.


I don't get all the comments, but here are the detils.....dataplane automatically registers nodes now and everything works other than the nova bits because of the reasons mentioned in the earlier comment. You can check the baremetal edpm job for details. For pre-provisioned nodes, ctlplane network would be not be managed by dataplane-operator and hence the you don't see dns entries for ctlplane network.  For pre-provisioned nodes, ansible_host would be mandatory.  As I mentioned everything other than nova bits would work with dns if this is not added in tech preview. The choice is yours. DNS and IPAM were always part of tech-preview unless I missed some decisions. IMO, we include things completed and working in Tech Preview.

SeanMooney:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml

dose not contain the nova template in the role so you never request nova to be deployed.

so that is expected.

if you want nova to be deployed you need to actually tell the dataplane operator to do that
by adding nova: {} to the   nodeTemplate.

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanerole_edpm_compute.yaml#L13

https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane_baremetal_with_ipam.yaml has nova disabled since by default EDPM nods are __NOT__ compute nodes by design.

rabi:
> dose not contain the nova template in the role so you never request nova to be deployed.

You think I didn't know that? :smile: https://github.com/openstack-k8s-operators/install_yamls/pull/408/commits/6f75d3bd581c2e405848e745d1178b25f7b11d2e meant to add that.

rabi:
> curl: (6) Could not resolve host: edpm-compute-0.internalapi.example.com

@gibizer  Looks like you've checked on a nova-conductor pod. They're not configured to use internal dns (as it's not required). Only ansibleee pods are configured with that.

```
[cloud-user@crc-rabi devsetup]$ oc rsh dataplane-deployment-run-os-edpm-compute-7w6q6
sh-5.2$ curl edpm-compute-0.internalapi.example.com
curl: (7) Failed to connect to edpm-compute-0.internalapi.example.com port 80 after 4 ms: Connection refused
sh-5.2$ cat /etc/resolv.conf 
search ctlplane.example.com
nameserver 192.168.122.80 ```

gibizer (approved):
I tested it does not break the ability to deploy an instance and we will rip this out anyway so land this for now.

rabi:
Rebased as precommit-check was failing without any merge conflicts. Would need the /lgtm again.

gibizer (approved):

PR #446, opened 2023-07-11
[EnvTest]Remove config deletion test cases

gibizer:
We had envtest case where the sub CR is deleted and asserted that the ConfigMap containing the generated service config is deleted. These cases only passed as they are checked the list of ConfigMaps in an non existing namespace. The test code wrongly passed the Name instead of the Namespace of the instance to the ListConfigMap call.

The ConfigMaps since replaced with Secrets but these testcases was missed in that commit.

The Secrets generated by the operator are all having a controller reference set. This way k8s automatically cleans up these resources after the CR parent is deleted. However this cleanup happens as part of a k8s controller which is not running in EnvTest therefore in EnvTest these resources are not deleted.

Therefore these false test cases now removed.

The resources we left behind in EnvTest are not causing any interaction between test cases as each test case runs in its own namespace.

### DIFF
--- a/test/functional/nova_metadata_controller_test.go
+++ b/test/functional/nova_metadata_controller_test.go
@@ -182,23 +182,6 @@ var _ = Describe("NovaMetadata controller", func() {
 				}, timeout, interval).Should(Succeed())
 
 			})
-
-			When("the NovaMetadata is deleted", func() {
-				It("deletes the generated ConfigMaps", func() {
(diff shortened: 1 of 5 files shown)


gibizer:
recheck retry limit

SeanMooney (approved):
Thanks for the detailed explanation in the PR/commit message i agree this can safely be removed.

PR #448, opened 2023-07-11
Update test helper references after reorg

gibizer:
Depends-On: openstack-k8s-operators/lib-common#301


gibizer:
As this is an envtest only change I have no problem landing it even before the devpreview.

The lib-common bump only pulls in two extra commits from main:
* https://github.com/openstack-k8s-operators/lib-common/commit/3945161b63e8a4802ea959138a6ee138489361fa
* https://github.com/openstack-k8s-operators/lib-common/commit/d1f65079749b04ee8e22817bae031904b32c495d
and both of those commits are envtest helpers only.

mrKisaoLamb (approved):
lgtm

PR #450, opened 2023-07-12
Copy rendered CRD files dir to /api/bases in make manifests

bshephar:
With the new api sub module, consumer operators just have to import the api submodule. For envtest it is required to install CRD for e.g. keystone-operator and mariadb-operator to be able to use it.

Moving the directory config/crd/bases which holds the generated files in to the api module that they are available with the module and link the dir in the original place does not work as kustomize autodetect fails on symlinked directories [1].

This updates the `generate` Makefile target to copy config/crd/bases to /api on successful generation run.

[1] kubernetes-sigs/kustomize#1886


bshephar:
I'm adding this because I need to be able to write some tests in Dataplane operator that allow it to get at least to the config map creation point to validate data. To do this, I need to import the CRDs for `NovaExternalCompute` in the envtest environment. 

This change is in-line with other operators to facilitate such efforts:
https://github.com/openstack-k8s-operators/keystone-operator/commit/5a39c3cee0d38e65d8c6fe572d5c0a30617ee78a

SeanMooney:
with the changes in the interaction between the service operator and the dataplane operator this might not be needed but i could be useful for the meta operator so I'm not against adding this i have not reviewed the PR yes just wanted to mention that at the start.

Makefile
```diff
@@ -105,7 +105,8 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 
 .PHONY: manifests
 manifests: gowork controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases && \
+ 	rm -f api/bases/* && cp -a config/crd/bases api/
```
SeanMooney, line 109: so this line is the real delta it just removes the old CRD and copies the newly updated ones.

given this does not seem to be breaking the pre-commit job I guess that means as long as the file content does not change it's okay to do this unconditionally.

SeanMooney (commented):
this looks reasonable to me but im not sure if we really want to include it in the dev preview.
that said I'm not entirely against it as it should not be able to affect the execution of the operator or impact customers. @gibizer  are you comfortable merging this or shall we hold of until after dev preview to merge this

SeanMooney (changes requested):
the functionl job failure is directly caused by this pr by the way so that will need to be fixed before this can proceed
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_nova-operator/450/pull-ci-openstack-k8s-operators-nova-operator-main-functional/1678991226314952704

SeanMooney:
/test functional

https://github.com/openshift/release/pull/41169 apprently is ment to fix the issue so perhaps it was just bad timing

kajinamit:
yeah we noticed that permission problem when adding the same to a few other operators and I submitted that CI fix to resolve it.

PR #451, opened 2023-07-13
Remove unused MessagingBusPassword

kajinamit:
These keys in the global secret are not used.


kajinamit:
I grepped these in current code but could not find any reference, so I believe these are unnecessary.

I've pushed https://github.com/openstack-k8s-operators/install_yamls/pull/402 to remove these from the secret generated by install_yamls as well.

gibizer (approved):

PR #452, opened 2023-07-13
Replace ansible execution with compute config generation

gibizer:
Based on the decisions in https://issues.redhat.com/browse/OSP-25812 the sevice-operators (like nova-operator) should not run or trigger ansible execution on the EDPM nodes. Instead each NovaCell should generate a Secret that contains configuration snippets that another operator (dataplane) can use to configure a nova-compute service connected to the given cell. 

Tasks:

- [x] Move compute config generation from `NovaExternalCompute` controller to `NovaCell` controller
- [x] Rename the compute config Secret to `<Nova name>-<NovaCell name>-compute-config` (e.g. `nova-cell1-compute-config`). Do not generate such config to `NovaCell/cell0` as computes cannot be connect to cell0.
- [x] Add a new `InputReady` codition to the `NovaCell.Status.Conditions` to be able to report if the input secrets are missing
- [x] Add a new `ComputeServiceConfig` condition to the `NovaCell.Status.Conditions` and 
    - [x] use it to indicate that the content of the config Secret is reconciled according to the `NovaCell.Spec`. 
    - [x] Also use it to indicate if for some reason (e.g. missing input) the config Secret is not exists / not up to date. 
    - [x] This condition should be part of the Ready condition but the negative state of this condition should not block the creation of the podified cell services.
- [x] Make sure that the compute service Secret is only generated **after** the `NovaNoVNCProxy` CR is ready in the cell as the config needs to contain the route to the service.
- [x] Ensure that if the cell configuration is changed the the content of the Secret is updated
- [x] Store the `hash` of the config Secret in the NovaCell.Status. This can be compared to the hash in the Status of the DataPlaneService (TBD) indicating what version of the configuration is deployed to decide if the latest config is already deployed or not.

Note that the cleanup of the NovaExternalCompute controller and the related pieces (code, test code, yaml definitions and samples, ansible, molecule, templates)  will be handled in a separate PR #476  to decouple this PR from the dataplane-operator work. E.g. we can support both the old and the new deployment procedure and clean up the old code only after dataplane-operator is switched to new deployment procedure.

Questions:
- ~~? Add `NovaCell.Status.ComputeConfig` string field that contains the name of the Secret containing the compute config~~ This is not needed any more as we store the hash of the config Secret in NovaCell.Status.Hash map where the key is the name of the config Secret.
- ~~? Add `NovaCell.Spec.CustomComputeServiceConfig` string field that provides a way for the end user to pass through an oslo.config snippet that will be added to the generated compute config Secret top of the operator generated default config? This would be a cell level config override that applies to each compute in the cell.~~ (agreed that we don't need it)

Closes: https://issues.redhat.com/browse/OSPRH-246

### DIFF
--- a/controllers/novacell_controller.go
+++ b/controllers/novacell_controller.go
@@ -20,14 +20,19 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
@@ -110,12 +115,48 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		}
 	}()
 
+	// For the compute config generation we need to read the input secrets
+	_, result, secret, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		[]string{
+			ServicePasswordSelector,
+		},
+		h.GetClient(),
+		&instance.Status.Conditions,
+		r.RequeueTimeout,
+	)
+	if err != nil {
+		return result, err
+	}
+
+	_, result, messageBusSecret, err := ensureSecret(
+		ctx,
+		types.NamespacedName{
+			Namespace: instance.Namespace,
+			Name:      instance.Spec.CellMessageBusSecretName,
+		},
+		[]string{
+			"transport_url",
+		},
+		h.GetClient(),
+		&instance.Status.Conditions,
+		r.RequeueTimeout,
+	)
+	if err != nil {
+		return result, err
+	}
+
+	// all our input checks out so report InputReady
+	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
+
 	result, err = r.ensureConductor(ctx, h, instance)
 	if err != nil {
 		return result, err
 	}
 
-	if *instance.Spec.MetadataServiceTemplate.Replicas != 0 && instance.Spec.CellName != novav1.Cell0Name {
+	isCell0 := (instance.Spec.CellName == novav1.Cell0Name)
+	if *instance.Spec.MetadataServiceTemplate.Replicas != 0 && !isCell0 {
 		result, err = r.ensureMetadata(ctx, h, instance)
 		if err != nil {
 			return result, err
@@ -124,13 +165,46 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		instance.Status.Conditions.Remove(novav1.NovaMetadataReadyCondition)
 	}
 
-	if *instance.Spec.NoVNCProxyServiceTemplate.Replicas != 0 && instance.Spec.CellName != novav1.Cell0Name {
+	cellHasVNCService := (*instance.Spec.NoVNCProxyServiceTemplate.Replicas != 0 && !isCell0)
+	if cellHasVNCService {
 		result, err = r.ensureNoVNCProxy(ctx, h, instance)
 		if err != nil {
 			return result, err
 		}
 	} else {
 		instance.Status.Conditions.Remove(novav1.NovaNoVNCProxyReadyCondition)
+	}
+
+	// We need to wait for the NovaNoVNCProxy to become Ready before we can try
+	// to generate the compute config secret as that needs the route of the
+	// proxy to be included.
+	// However NovaNoVNCProxy is never deployed in cell0, and optional in other
+	// cells too.
+	if cellHasVNCService && !instance.Status.Conditions.IsTrue(novav1.NovaNoVNCProxyReadyCondition) {
+		util.LogForObject(
+			h,
+			"Waiting for the NovaNoVNCProxyService to become Ready before "+
+				"generating the compute config", instance,
+		)
+		return ctrl.Result{}, nil
+	}
+
+	var vncHost *string
+	if cellHasVNCService {
+		vncHost, err = r.getVNCHost(ctx, h, instance)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if !isCell0 {
+		result, err = r.ensureComputeConfig(ctx, h, instance, secret, messageBusSecret, vncHost)
+		if (err != nil || result != ctrl.Result{}) {
+			return result, err
+		}
+
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaComputeServiceConfigReady)
 	}
 
 	util.LogForObject(h, "Successfully reconciled", instance)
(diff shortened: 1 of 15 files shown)


gibizer:
```
$ oc get -n openstack secret | grep compute-config
nova-cell1-compute-config                          Opaque                                2      3m5s

$ oc get secret nova-cell1-compute-config -o yaml | yq '.data."01-nova.conf"' -r | base64 --decode
[DEFAULT]
# concurrent live migrations are more likely to fail and are slower
# overall then serializing live migrations so set this to 1 explictly
state_path = /var/lib/nova
# enable log rotation in oslo config by default
max_logfile_count=5
max_logfile_size_mb=50
log_rotation_type=size

log_file = /var/log/containers/nova/nova-compute.log

debug=true

compute_driver=libvirt.LibvirtDriver

# ensure safe defaults for new hosts


force_config_drive=True

transport_url=rabbit://default_user_gzdK-bWHUd9p1A9PXPl:zTZD5yGZvj7KiMrxsYBckilY_y97jkQ6@rabbitmq-cell1.openstack.svc:5672

[..snip...]
[vnc]
enabled = True
novncproxy_host = "::0"
novncproxy_port = 6080
server_listen = "::0"


# note we may want to use console_host instead of my_ip however it wont be resolved via
# dns currently so we need to use my_ip for now.
# https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.console_host
server_proxyclient_address = "$my_ip"

novncproxy_base_url = http://nova-novncproxy-cell1-public-openstack.apps-crc.testing/vnc_lite.html

[...snip...]

[libvirt]
live_migration_permit_post_copy=true
live_migration_permit_auto_converge=true
live_migration_timeout_action=force_complete
cpu_mode=host-model
hw_machine_type=x86_64=q35
sysinfo_serial=unique
mem_stats_period_seconds=0
num_pcie_ports=24
images_type=qcow2
rx_queue_size=512
tx_queue_size=512
swtpm_enabled=True

[...snip...]

[service_user]
send_service_user_token = true
auth_url = http://keystone-internal.openstack.svc:5000
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
password = 12345678
cafile = 
```

gibizer:
/retest-required
telemetry-operator tagging has been fixed so the build should pass now

gibizer:
recheck

slagle (commented):
LGTM. I've tested this along with https://github.com/openstack-k8s-operators/dataplane-operator/pull/297 and it WFM :+1:

SeanMooney:
this is on my review list of Monday.
we are still trying to hold large feature until after the dev preview ships
but if we are confident that no new bug are going to need to be addressed we can likely proceed with this next week. the removal of the NovaExternalCompute is done in a follow up https://github.com/openstack-k8s-operators/nova-operator/pull/476 so we do not loses compatibility with the dev preview until that is merged.

controllers/novacell_controller.go
```diff
@@ -330,12 +419,112 @@ func (r *NovaCellReconciler) ensureMetadata(
...
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&novav1.NovaCell{}).
 		Owns(&novav1.NovaConductor{}).
 		Owns(&novav1.NovaMetadata{}).
 		Owns(&novav1.NovaNoVNCProxy{}).
+		// It generates and therefor owns the compute config secret
+		Owns(&corev1.Secret{}).
+		// and it needs to watch the input secrets
+		Watches(&source.Kind{Type: &corev1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(r.GetSecretMapperFor(&novav1.NovaCellList{}))).
```
SeanMooney, line 528: im not really that familary with this but i assume this is how we watch only the secretes relevent to this cell

templates/nova.conf
```diff
@@ -131,9 +131,11 @@ server_listen = {{ if (index . "novncproxy_service_host") }}{{ .novncproxy_servi
 # dns currently so we need to use my_ip for now.
 # https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.console_host
 server_proxyclient_address = "$my_ip"
+{{if (index . "novncproxy_base_url")}}
 novncproxy_base_url = {{ .novncproxy_base_url }}/vnc_lite.html
 {{ end }}
```
SeanMooney, line 136: ack this is to accound for the fact its optional

test/functional/nova_reconfiguration_test.go
```diff
@@ -430,6 +433,16 @@ var _ = Describe("Nova reconfiguration", func() {
...
+
+			// Expect that the compute config is updated with the new transport URL
+			Eventually(func(g Gomega) {
+				configDataMap := th.GetSecret(cell1.ComputeConfigSecretName)
+				g.Expect(configDataMap).ShouldNot(BeNil())
+				g.Expect(configDataMap.Data).Should(HaveKey("01-nova.conf"))
+				configData := string(configDataMap.Data["01-nova.conf"])
+				g.Expect(configData).Should(ContainSubstring("transport_url=rabbit://alternate-mq-for-cell1-secret/fake"))
+			}, timeout, interval).Should(Succeed())
+			Expect(GetNovaCell(cell1.CellName).Status.Hash[cell1.ComputeConfigSecretName.Name]).NotTo(Equal(oldComputeConfigHash))
```
SeanMooney, line 445: +1

the dataplane operator or a human would need to triger the compute to be updated for this to propagate but form a nova operator poit of vew this should be good.

test/functional/novacell_controller_test.go
```diff
@@ -242,10 +256,48 @@ var _ = Describe("NovaCell controller", func() {
 			// Expect(novaCell.Status.MetadataServiceReadyCount).To(Equal(int32(1)))
 		})
 
+		It("creates the compute config secret", func() {
```
SeanMooney, line 259: as a follow up it would be nice to also assert this with a kuttl test but im happy with the env-test coverage to move forward with this pr

test/functional/novacell_controller_test.go
```diff
@@ -256,6 +308,65 @@ var _ = Describe("NovaCell controller", func() {
...
+
+			computeConfigData := th.GetSecret(cell2.ComputeConfigSecretName)
+			Expect(computeConfigData).ShouldNot(BeNil())
+			Expect(computeConfigData.Data).Should(HaveKey("01-nova.conf"))
+			configData := string(computeConfigData.Data["01-nova.conf"])
+			Expect(configData).To(ContainSubstring("transport_url=rabbit://rabbitmq-secret/fake"))
+			Expect(configData).To(ContainSubstring("username = nova\npassword = service-password\n"))
+			// There is no VNCProxy created but we still get a compute config just
+			// without any vnc proxy url
+			Expect(configData).NotTo(ContainSubstring("novncproxy_base_url "))
```
SeanMooney, line 346: nice this also cover the optional proxy case when its disabeld cool.

templates/nova.conf
```diff
@@ -131,9 +131,11 @@ server_listen = {{ if (index . "novncproxy_service_host") }}{{ .novncproxy_servi
 # dns currently so we need to use my_ip for now.
 # https://docs.openstack.org/nova/latest/configuration/config.html#DEFAULT.console_host
 server_proxyclient_address = "$my_ip"
+{{if (index . "novncproxy_base_url")}}
```
SeanMooney, line 134: you could actully wrap all the vnc sectionin this if.
when we do not have a proxy we do not need to enable vnc on the compute.

SeanMooney (approved):
given james testing in https://github.com/openstack-k8s-operators/dataplane-operator/pull/297
and the fact i dont see any significant issues in this PR im happy to proceed with this and adress any follow ups seperatly to allow the datapalne work to continue.

controllers/novacell_controller.go
gibizer: This is basically a filter to watch only the secrets that's name is passed as Spec.Secret so we don't get reconciled in all the Secret changes in all the namespaces.

templates/nova.conf
gibizer: The we should not use novncproxy_base_url to indicate if the vnc section is needed or not

test/functional/novacell_controller_test.go
gibizer: I'm adding kuttl coverage for this in https://github.com/openstack-k8s-operators/nova-operator/pull/510

controllers/novacell_controller.go
```diff
@@ -133,6 +177,38 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+	// proxy to be included.
+	// However NovaNoVNCProxy is never deployed in cell0, and optional in other
+	// cells too.
+	if cellHasVNCService && !instance.Status.Conditions.IsTrue(novav1.NovaNoVNCProxyReadyCondition) {
+		util.LogForObject(
+			h,
+			"Waiting for the NovaNoVNCProxyService to become Ready before "+
+				"generating the compute config", instance,
+		)
+		return ctrl.Result{}, nil
```
bogdando, line 191: as we do not requeue it, would it reconciliate that "lost update" after the NovaNoVNCProxyReadyCondition is set to True?

gibizer: The novacell controller is the owner and controller of the NovaNoVNCProxy CR so it will reconcile if that CR changes.

PR #453, opened 2023-07-14
Fix *ReadyCount propagation

gibizer:
There was a bug in the NovaCell controller, it did not propagated the
ReadyCount from the cell service CRs to the NovaCell CR status.

Also the NovaMetadata.Status.ReadyCount wrongly propagated to the
Nova.Status.APIReadyCount. 

These are fixed now and the test coverage is extended to cover these fields.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -1260,7 +1260,7 @@ func (r *NovaReconciler) ensureMetadata(
 	if c != nil {
 		instance.Status.Conditions.Set(c)
 	}
-	instance.Status.APIServiceReadyCount = metadata.Status.ReadyCount
+	instance.Status.MetadataServiceReadyCount = metadata.Status.ReadyCount
 
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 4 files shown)


gibizer:
We discussed on the today's sync call and Sean suggested to top this field from the NovaCell and Nova CRs. I checked other operators and I see that ReadyCount is on Cinder Glance and Manila top level CR too. So I if we want to drop this from nova then we probably need to agree and drop it from the storage top level CRs too.

controllers/nova_controller.go
```diff
@@ -1260,7 +1260,7 @@ func (r *NovaReconciler) ensureMetadata(
 	if c != nil {
 		instance.Status.Conditions.Set(c)
 	}
-	instance.Status.APIServiceReadyCount = metadata.Status.ReadyCount
+	instance.Status.MetadataServiceReadyCount = metadata.Status.ReadyCount
```
SeanMooney, line 1263: yep that was definetly a bug

SeanMooney (approved):

Issue #454, opened 2023-07-17
VNC proxy request are randomly routed cross cell


gibizer:
Reproduction:
1. Deploy the default network isolation sample of OpenstackControlPlane
2. Deploy cell2 by adding a new section to cellTemplates in OpenStackControlPlane CR. Deploy a separate rabbitmq cluster for the new cell too.  (You can re-use the same DB instance as for cell1 or cell0 for this deployment)
3. Deploy at least one compute with OpenStackDataPlane. The default sample will deploy it to cell1.
4. Create a guest VM
5. grab the VNC url of the guest from `openstack console url show `
6. connect to the VNC console

Expected:
* The VNC connection is routed through the cell1 proxy
* The connection is established successfully

Actual:
* Sometimes the connection is routed to the cell2 proxy and therefore fails to validate the token
```
2023-07-17 13:06:24.027 1621 ERROR nova.console.websocketproxy nova.exception.InvalidToken: The token '***' is invalid or has expired
```

I think the reason of the failure is that both nova-novncproxy-cell1-public and nova-novncproxy-cell2-public serivce and endpoints are using the same set of labels and therefore the load balance balances requests equally between the cell1 and cell2 backend regardless of which service is used to do the request. If we look at the mariadb or the rabbitmq endpoints then we see different set of labels per backend - service pair to separate traffic.

gibizer:
```
$ oc get service nova-novncproxy-cell1-public -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-07-17T12:10:45Z"
  labels:
    public: "true"
    service: nova-novncproxy
...
$ oc get service nova-novncproxy-cell2-public -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2023-07-17T12:21:41Z"
  labels:
    public: "true"
    service: nova-novncproxy
...
$ oc get pods nova-cell1-novncproxy-0 -o yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2023-07-17T12:10:45Z"
  generateName: nova-cell1-novncproxy-
  labels:
    cell: cell1
    controller-revision-hash: nova-cell1-novncproxy-6ffbdb9c89
    service: nova-novncproxy
    statefulset.kubernetes.io/pod-name: nova-cell1-novncproxy-0
...
$ oc get pods nova-cell2-novncproxy-0 -o yaml
apiVersion: v1
kind: Pod
[...]

PR #455, opened 2023-07-17
Fix noVNC labels to be consistent for services and pods

mrKisaoLamb:
#454

### DIFF
--- a/controllers/novanovncproxy_controller.go
+++ b/controllers/novanovncproxy_controller.go
@@ -335,11 +335,9 @@ func (r *NovaNoVNCProxyReconciler) ensureDeployment(
 	inputHash string,
 	annotations map[string]string,
 ) (ctrl.Result, error) {
-	serviceLabels := map[string]string{
-		common.AppSelector: NovaNoVNCProxyLabelPrefix,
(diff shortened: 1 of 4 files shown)


gibizer (commented):
This works as expected. I tried to reproduce the vnc routing failure but could not with this fix. I did not verified the per cell metadata case but given the nature of the fix is the same I think we OK to land this. This is a bugfix so I think we should land this before the dev-preview.

controllers/novanovncproxy_controller.go
```diff
@@ -473,9 +471,10 @@ func (r *NovaNoVNCProxyReconciler) reconcileDelete(
 	return nil
 }
 
-func getNoVNCProxyServiceLabels() map[string]string {
+func getNoVNCProxyServiceLabels(cell string) map[string]string {
 	return map[string]string{
 		common.AppSelector: NovaNoVNCProxyLabelPrefix,
+		CellSelector:       cell,
```
SeanMooney, line 477: ok so here its uncondtionally requried as the novncproxy is only deployable at the cell level

SeanMooney (approved):

PR #456, opened 2023-07-17
update nova/libvirt bindmounts

SeanMooney:
This change adds :shared to ensure submounts propagate to the host.
This change add :z to enable selinux labeling to allow shared mounts
to be shared properly between multiple containers.


SeanMooney:

it failed because of the pull secret which is unrelated to this change

gibizer (approved):
Sean tested this locally and it does not break the existing functionality. So I'm OK to merge this in the hope that it fixes the telemetry issue.

Issue #457, opened 2023-07-17
NovaNoVNCProxy cannot be scaled to 0


gibizer:
This condition actually ignores the scaling request:
 
https://github.com/openstack-k8s-operators/nova-operator/blob/a4bd2d65b166e9196bb01c5cfe1fdccb9371f454/controllers/novacell_controller.go#L127-L134

gibizer:
similarly this affect nova-metadata as well:
https://github.com/openstack-k8s-operators/nova-operator/blob/a4bd2d65b166e9196bb01c5cfe1fdccb9371f454/controllers/novacell_controller.go#L118-L125

PR #458, opened 2023-07-18
Cell envtest improvements

gibizer:
This PR is envtest only and improve cell controller coverage in multiple steps.

PR #459, opened 2023-07-18
Propagate Replicas 0 to NoVNCProxy CR

gibizer:
The existing logic ignored all NovaNoVNCProxy CRs with Replicas=0. This
was added to not create NovaNoVNCProxy for cell0 or for cells where no
VNCProxy is requested at all. However this logic means that if VNCProxy
is deployed for a cell with replicas > 0 and then later we need to scale
it down to replica 0, then that later change is simply ignored as the
NoVNCProxy CR is not updated.

We are mixing two cases into replicas 0:
1) the service does not need to deployed at all
2) the service needs to be temporary scaled down to 0 and then scaled
   back up later

For the 2) we can clearly use the replicas field. But for 1) we should use a different field not to mix the two cases.

So the current PR introduces an Enabled field to the NoVNCProxyTemplate to
be used for case 1). If Enabled is set to false then the NovaNoVNCPRoxy
CR will not be created.

This patch also ensures that the Enabled field is defaulted to true
except for cell0 where it is defaulted to false. Also a validation
webhook ensures that it cannot be set to true to cell0.

There are two edge cases:
1. Enabled is set from False to True while a NovaNoVNCPRoxy CR exists. E.g. somebody created a NovaNoVNCProxy CR that is not owned by the NovaCell. In this case the NovaCell controller will not touch the NovaNoVNCProxy CR as it would not be able to fully update it due to immutable fields in the StatefulSet. Instead NovaCell goes to error state and ask the operator to delete the CR manually.
2. Enabled is set from True to False while NovaNoVNCProxy CR exists. If the NovaNoVNCProxy CR is owned by the NovaCell then it will be automatically deleted, but if it is not owned by the NovaCell then it will not be touched. Also in this case no error will be reported on the NovaCell.

Tasks: 
* [x] introduce an `Enabled` field to the `NovaNoVNCProxyTemplate`
* [x] implement delete logic for Enabled True -> False transition
* [x] ensure that we only delete it if we own it
* [x] ensure that Replicas is propagated if Enabled == True
* [x] ~ensure that if the resource already exists without our ownership then we overwrite it but also start owning it.~ It turned out that it is not possible as the label selector of the StatefulSet is immutable so the cell cannot update it when it starts owning the manually created NoVNCProxy. So instead we document that in such case the NoVNCProxy CR needs to be manually deleted first.

Depends-on: #458 (merged)
Closes: #457


api/v1beta1/nova_webhook.go
```diff
@@ -99,8 +100,16 @@ func (spec *NovaSpec) Default() {
...
 		}
 
+		if cellName == Cell0Name {
+			// NOTE(gibi): our cell0 NoVNCProxy templates needs different default
+			// values than the rest of the cells as cell0 should not have VNCProxy
+			if cellTemplate.NoVNCProxyServiceTemplate.Replicas == nil {
+				cellTemplate.NoVNCProxyServiceTemplate.Replicas = pointer.Int32(0)
+			}
+		}
+
```
gibizer, line 110: Until it is not built into the openstack-operator the current sample in openstack-operator will fail with: 

```
:"admission webhook \"vnova.kb.io\" denied the request: Nova.nova.openstack.org \"nova\" is invalid: spec.cellTemplates[cell0].noVNCProxyServiceTemplate.replicas: Invalid value: 1: should be 0 for cell0","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.
```

It is because the old defaulting built currently into openstack-operator will default the vncproxy replicas to 1 so when the request reaches the new webhook in the nova-operator it is not defaulted again, but the validation hook will reject it.

gibizer:
I'm reconsidering https://github.com/openstack-k8s-operators/nova-operator/pull/459 and https://github.com/openstack-k8s-operators/nova-operator/pull/460  in the light of the OpenStackControlPlane behavior where enabled:false is now deleting the sub CRs. I think we might conflate two separate use case in nova-operator:
1. As an end user I want to temporary disable a service by setting Replicas to 0 then I want to enable it later by restoring replicas to > 0
2. As an end user I don't want to deploy a service at all and therefore I don't want to create any kind of (even default) configuration for it. (E.g. one might not want to have metadata in cells at all)

I think we need to support both use cases. And therefore Replicas=0 cannot be used for both. I think we can start distinguishing the two cases by switching the <service>Template fields to pointers. Then nil would mean, "do not deploy the service", while Replicas=0 means deploy the service but (temporarily) set the StatefulSet Replicas to 0.

There are two complications though:
We need to handle when a <service>Template is set from non nil to nil and delete the service CR. This will happen based on the stable name we generate for the service CR (and we will check the owner reference to avoid deleting what we don't own).
We need to handle when a <service>Template is set to nil, a service CR with the stable name exists without our owner reference, and then the user set the <service>Template to non-nil. In this case the default behavior will be that the controller overwrites the pre-existing service CR and put an owner reference on it.

A more explicit alternative would be to have a new "enabled" flag for each optional service similarly how OpenStackControlPlane CRD has it.

gibizer:
> I'm reconsidering #459 and #460 in the light of the OpenStackControlPlane behavior where enabled:false is now deleting the sub CRs. I think we might conflate two separate use case in nova-operator:
> 
> 1. As an end user I want to temporary disable a service by setting Replicas to 0 then I want to enable it later by restoring replicas to > 0
> 2. As an end user I don't want to deploy a service at all and therefore I don't want to create any kind of (even default) configuration for it. (E.g. one might not want to have metadata in cells at all)
> 
> I think we need to support both use cases. And therefore Replicas=0 cannot be used for both. I think we can start distinguishing the two cases by switching the Template fields to pointers. Then nil would mean, "do not deploy the service", while Replicas=0 means deploy the service but (temporarily) set the StatefulSet Replicas to 0.
> 
> There are two complications though: We need to handle when a Template is set from non nil to nil and delete the service CR. This will happen based on the stable name we generate for the service CR (and we will check the owner reference to avoid deleting what we don't own). We need to handle when a Template is set to nil, a service CR with the stable name exists without our owner reference, and then the user set the Template to non-nil. In this case the default behavior will be that the controller overwrites the pre-existing service CR and put an owner reference on it.
> 
> A more explicit alternative would be to have a new "enabled" flag for each optional service similarly how OpenStackControlPlane CRD has it.

We are leaning fowards the more explicit "enabled" flag driven alternative.

gibizer:
it was just a rebase before I redesign it

gibizer:
> > I'm reconsidering #459 and #460 in the light of the OpenStackControlPlane behavior where enabled:false is now deleting the sub CRs. I think we might conflate two separate use case in nova-operator:
> > 
> > 1. As an end user I want to temporary disable a service by setting Replicas to 0 then I want to enable it later by restoring replicas to > 0
> > 2. As an end user I don't want to deploy a service at all and therefore I don't want to create any kind of (even default) configuration for it. (E.g. one might not want to have metadata in cells at all)
> > 
> > I think we need to support both use cases. And therefore Replicas=0 cannot be used for both. I think we can start distinguishing the two cases by switching the Template fields to pointers. Then nil would mean, "do not deploy the service", while Replicas=0 means deploy the service but (temporarily) set the StatefulSet Replicas to 0.
> > There are two complications though: We need to handle when a Template is set from non nil to nil and delete the service CR. This will happen based on the stable name we generate for the service CR (and we will check the owner reference to avoid deleting what we don't own). We need to handle when a Template is set to nil, a service CR with the stable name exists without our owner reference, and then the user set the Template to non-nil. In this case the default behavior will be that the controller overwrites the pre-existing service CR and put an owner reference on it.
> > A more explicit alternative would be to have a new "enabled" flag for each optional service similarly how OpenStackControlPlane CRD has it.
> 
> We are leaning fowards the more explicit "enabled" flag driven alternative.

This is implemented now.

test/functional/novacell_controller_test.go
```diff
@@ -361,9 +408,147 @@ var _ = Describe("NovaCell controller", func() {
...
+			Eventually(func(g Gomega) {
+				novaCell := GetNovaCell(cell2.CellName)
+				novaCell.Spec.ConductorServiceTemplate.Replicas = ptr.To[int32](3)
+
+				g.Expect(k8sClient.Update(ctx, novaCell)).To(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			// Just ensure that it is not automatically gets owned or deleted
+			// by the cell
+			Consistently(func(g Gomega) {
```
SeanMooney, line 496: i assume Consistently will run in a go routine adn continute to assert this until the current test case ends?
i havent used this but form context thise makes sense to me.

test/functional/novacell_controller_test.go
```diff
@@ -361,9 +408,147 @@ var _ = Describe("NovaCell controller", func() {
...
+			th.DeleteInstance(GetNovaNoVNCProxy(cell2.NoVNCProxyName))
+			// NOTE(gibi): This only needed in envtest, in a real k8s
+			// deployment the garbage collector would delete the StatefulSet
+			// when its parents, the NoVNCProxy, is deleted, but that garbage
+			// collector does not run in envtest. So we manually clean up here
+			th.DeleteInstance(th.GetStatefulSet(cell2.NoVNCProxyStatefulSetName))
+
+			// As the manually created NoVNCProxy is deleted the controller is
+			// unblocked to deploy its own NoVNCProxy CR
+			th.SimulateStatefulSetReplicaReady(cell2.NoVNCProxyStatefulSetName)
```
SeanMooney, line 535: im wonderign how this interact with teh previous Consitently line?

dose this not cause it to fail becuae the previous test is checking via uuid
instead of name and there for will not confilct when we create a new cr with the
same name but diffent uuid.

if so then cool we can leave thsi as is i just want to make sure this is not a race or potential instablity in the tests.

test/functional/validation_webhook_test.go
```diff
@@ -94,6 +94,72 @@ var _ = Describe("Nova validation", func() {
...
+		cell0Template["noVNCProxyServiceTemplate"] = map[string]interface{}{
+			"enabled": true,
+		}
+
+		spec["cellTemplates"] = map[string]interface{}{
+			"cell0": cell0Template,
+			// note that this is intentional to test that novncproxy is allowed
+			// in cell1 but not in cell0
+			"cell1": cell0Template,
+		}
```
SeanMooney, line 109: +1

SeanMooney (commented):
in generall i think this looks good. a few questions inlien if you can answer them ill apporve later today

api/v1beta1/novanovncproxy_types.go
```diff
@@ -28,6 +28,19 @@ import (
 // NovaNoVNCProxyTemplate defines the input parameters specified by the user to
 // create a NovaNoVNCProxy via higher level CRDs.
 type NovaNoVNCProxyTemplate struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
```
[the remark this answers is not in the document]
gibizer: I think I see your point. For cell0 we want to set the default to false and otherwise true, so one side of this defaulting already happens in the hook. I agree this should not have been split between kubebuilder and webhook. Let me fix this...

test/functional/base_test.go
```diff
@@ -250,16 +250,15 @@ func CreateNovaMessageBusSecret(namespace string, name string) *corev1.Secret {
...
-		"serviceAccount":            "nova",
+	spec := map[string]interface{}{
+		"cellName":                 cellName,
+		"secret":                   SecretName,
+		"cellDatabaseHostname":     "cell-database-hostname",
+		"cellMessageBusSecretName": MessageBusSecretName,
+		"keystoneAuthURL":          "keystone-auth-url",
+		"serviceAccount":           "nova",
 	}
+	return spec
```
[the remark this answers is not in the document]
gibizer: ohh this has to be a rebase artifact. I will switch it back...

test/functional/nova_multicell_test.go
```diff
@@ -98,6 +99,8 @@ var _ = Describe("Nova multicell", func() {
 			// assert that cell related CRs are created pointing to the API MQ
 			cell := GetNovaCell(cell0.CellName)
 			Expect(cell.Spec.CellMessageBusSecretName).To(Equal(fmt.Sprintf("%s-secret", cell0.TransportURLName.Name)))
+			// and NoVNCProxy is defaulted to be disabled in cell0
+			Expect(cell.Spec.NoVNCProxyServiceTemplate.Enabled).To(Equal(ptr.To(false)))
```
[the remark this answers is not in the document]
gibizer: `&false` is invalid in go as you cannot take the address of a const. `x:=false; &x` would work and I used that in the past. But there is the ptr module that does the same with `ptr.To[T any](T v)` https://github.com/kubernetes/utils/blob/3b25d923346b3814e0898684c97390092f31a61e/ptr/ptr.go#L50-L52 (Note that T is left out as it is inferred by the compiler)

test/functional/novacell_controller_test.go
gibizer: Yepp exactly. https://onsi.github.io/gomega/#consistently

I prefer to use Eventually with a condition that signals the end of an operation as that is faster than blindly checking untily `consistencyTimeout`. However in this case I have no signal to wait for from the operator as I want to check that actually nothing is done by the operator.

test/functional/novacell_controller_test.go
gibizer: Good point. So within this single test case there will be two NovaNoVNCProxy CR (and two StatefulSet) with the same name but different UID. All our helpers work based on names not UIDs. So here the helper does not know which StatefulSet is being updated the first or the second. 

Our DeleteInstance helper does not actually wait until the the CR disappears, it only checks that the Delete request was accepted.

So from the test perspective there is a race window between 
```
th.DeleteInstance(GetNovaNoVNCProxy(cell2.
th.DeleteInstance(th.GetStatefulSet(cell2.NoVNCProxyStatefulSetName))
```
and 
```
th.SimulateStatefulSetReplicaReady(cell2.NoVNCProxyStatefulSetName)
```
As the Simulate call might still see the first VNC CR and not the second one.

I will fix this up

gibizer (commented):
Good catches, I will couple of things to fix...

api/v1beta1/novacell_webhook.go
```diff
@@ -83,6 +84,18 @@ func (spec *NovaCellSpec) Default() {
...
+		// in cell0 disable VNC by default
+		if spec.NoVNCProxyServiceTemplate.Enabled == nil {
+			spec.NoVNCProxyServiceTemplate.Enabled = ptr.To(false)
+		}
+	} else {
+		// in other cells enable VNC  by default
+		if spec.NoVNCProxyServiceTemplate.Enabled == nil {
+			spec.NoVNCProxyServiceTemplate.Enabled = ptr.To(true)
+		}
+	}
```
SeanMooney, line 98: thanks that much nicer.

SeanMooney (commented):
ill take a look at this again tomorrow but i think it looks fine to proceed with.

gibizer:
/unhold 
fixed the review findings
(A note to myself: do not add text directly after /unhold as contrarry to /hold it will cause the command to be ignored)

controllers/novacell_controller.go
```diff
@@ -319,15 +325,59 @@ func (r *NovaCellReconciler) ensureConductor(
...
+
+	novncproxyName := getNoVNCProxyName(instance)
+	novncproxy := &novav1.NovaNoVNCProxy{}
+	err := r.Client.Get(ctx, novncproxyName, novncproxy)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	// If it is not created by us, we don't touch it
+	if !k8s_errors.IsNotFound(err) && !OwnedBy(novncproxy, instance) {
```
mrKisaoLamb, line 356: Just considering out loud: Would it be a good idea to perform this type of validation within the webhook to prevent entering an 'error' state?

mrKisaoLamb (commented):
lgtm

controllers/novacell_controller.go
gibizer: Do you mean rejecting a NovaNoVNCProxy CR creation if it is not having an owner reference set to a NovaCell CR instance? We could do that in the validation webhook but I think we don't want to prevent a standalone NovaNoVNCProxy CR creation by a human user (or the envtest test case).

mrKisaoLamb: Yes I was thinking about rejecting NovaNoVNCProxy CR with wrong ownership, but maybe when we set in NovaCell enable = true we can check in update webhook if already is novncproxy without NovaCell ownership and reject NovaCell update with NovaNoVNCProxy template with enable = true. In that case we will 'fail fast' and user will get error earlier. But I have doubts about whether this check within the competency of the webhook in our case.

gibizer: Ahh I see your point. So basically detecting the manually created NovaNoVNCProxy at NovaCell create or update. This would require that the webhook logic try to reads stuff from the kube-apiserver to see if NovaNoVNCProxy exists with the same name the NovaCell would like to use. 

The doc https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#best-practices-and-warnings does not implicitly forbid  doing such thing in a validating webhook. But:
* from that doc: `It is recommended that admission webhooks should evaluate as quickly as possible (typically in milliseconds)` Reading from the kube-apiserver would take significant time compared the to current processing of the webhook.
* in the context of the webhook code we don't have an api client set up with credentials to talk to the kube-apiserver
* even if the validation webhook would see that there is no NovaNoVNCProxy CR with the name in question we cannot assume in the Reconciler that such CR is not created in the meantime between the NovaCell validation webhook run and the NovaCell Reconcile call. So the logic in the Reconciler still need to be kept. 

At the moment I think it is too much additional complexity for a limited additional value.

mrKisaoLamb: Agree

gibizer:

job timed out even before finishing the tempest configuration
```
2023-08-29 11:30:12.296 127 DEBUG config_tempest.constants [-] Setting [image-feature-enabled] import_image = True set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.297 127 DEBUG config_tempest.constants [-] Setting [validation] image_ssh_user = cirros set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.447 127 INFO tempest.lib.common.rest_client [req-3a8b2f84-03dc-4f11-9256-549579f0e65a req-3a8b2f84-03dc-4f11-9256-549579f0e65a ] Request (main): 200 GET http://glance-public-openstack.apps.oko-11.oooci.ccitredhat.com/v2/images 0.149s[00m
2023-08-29 11:30:12.447 127 DEBUG config_tempest.constants [-] Setting [image] http_image = https://download.cirros-cloud.net/0.5.2/cirros-0.5.2-x86_64-disk.img set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.447 127 DEBUG config_tempest.constants [-] Setting [service_available] glance = True set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.448 127 DEBUG config_tempest.constants [-] Setting [service_available] barbican = False set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.448 127 DEBUG config_tempest.constants [-] Setting [service_available] zaqar = False set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
2023-08-29 11:30:12.448 127 DEBUG config_tempest.constants [-] Setting [service_available] gnocchi = False set /usr/lib/python3.9/site-packages/config_tempest/tempest_conf.py:105[00m
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 2h0m0s 
```

SeanMooney (approved):

PR #464, opened 2023-07-21
[envtest]Pull out NoVNCProxy Route Ingress simulation

gibizer:
Create an envtest helper that does the Route manipulation.

I intentionally dropped
```golang
			vncRoute.Spec.Host = "nova-novncproxy-cell1-public-openstack.apps-crc.testing"
```
from the helper. Changing the Spec of the Route from envtest triggers reconcile on the nova-operator that in turn changes things back. It should be enough to simulate the Ingress in the Status.

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


gibizer:
This was merged as 29a698b02b0db3edd42c88a990e951b7adbf34b4 part of https://github.com/openstack-k8s-operators/nova-operator/pull/452

PR #476, opened 2023-07-26
Remove NovaExternalCompute CRD

gibizer:
We are removing the support for deploying nova-compute on EDPM nodes
from nova-operator directly. The nova-compute deployment will be handled
by the dataplane-operator and nova-operator only generates config for it
implemented in https://github.com/openstack-k8s-operators/nova-operator/pull/452

**Note** that this is a heavily backward incompatible change. Any deployment that has NovaExternalCompute CRs cannot be really moved to this change without redeployment or extensive manual changes

Depends-on: #452


SeanMooney (changes requested):
ok so this is almost correct.
this should also delete teh playbook dir

https://github.com/openstack-k8s-operators/nova-operator/tree/main/playbooks

other then that this looks good to me

gibizer:
> ok so this is almost correct. this should also delete teh playbook dir
> 
> https://github.com/openstack-k8s-operators/nova-operator/tree/main/playbooks
> 
> other then that this looks good to me

Good point. Done.

SeanMooney (approved):

SeanMooney (approved):
good catch i forgot we included the playbooks in the docker image

PR #477, opened 2023-07-26
[ironic-compute] Add nova-compute-ironic controller

mrKisaoLamb:


api/v1beta1/conditions.go
```diff
@@ -47,6 +47,8 @@ const (
 	NovaMetadataReadyCondition condition.Type = "NovaMetadataReady"
 	// NovaNoVNCProxyReadyCondition indicates when the given NoVNCProxy instance is Ready
 	NovaNoVNCProxyReadyCondition condition.Type = "NovaNoVNCProxyReady"
+	// NovaComputeIronicReadyCondition indicates when the given NoVNCProxy instance is Ready
```
gibizer, line 50: copy paste: NoVNCProxy

api/v1beta1/conditions.go
```diff
@@ -47,6 +47,8 @@ const (
...
+	NovaComputeIronicReadyCondition condition.Type = "NovaComputeIronicReady"
```
gibizer, line 51: Wondering if we want a separate condition type per virt driver. But this depends on if we create a specific NovaComputeIronic instead of a generic NovaCompute with a virt driver field. (see my specific comment below)

api/v1beta1/novacomputeironic_types.go
```diff
@@ -0,0 +1,214 @@
...
+	// infinity
+	Debug Debug `json:"debug,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// NovaServiceBase specifies the generic fields of the service
+	NovaServiceBase `json:",inline"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
```
gibizer, line 125: I don't think we need to externally expose anything from the compute service

api/v1beta1/novacomputeironic_types.go
```diff
@@ -0,0 +1,214 @@
...
+	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
```
gibizer, line 70: I don't think we need to externally expose anything from the compute service

api/v1beta1/novacell_types.go
```diff
@@ -161,9 +167,19 @@ type NovaCellSpec struct {
...
+	NovaComputeIronicServiceTemplate NovaComputeIronicTemplate `json:"novaComputeIronicServiceTemplate"`
+
 	// +kubebuilder:validation:Required
 	// ServiceAccount - service account name used internally to provide Nova services the default SA name
 	ServiceAccount string `json:"serviceAccount"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="libvirt.LibvirtDriver"
+	// ComputeDriver defines which driver to use for controlling virtualization
+	ComputeDriver string `json:"computeDriver"`
```
gibizer, line 182: hm hm. I thought we wanted to add the ComputeDriver to the NovaCompute(Ironic)Template to select from Ironic and Fake drivers.

api/v1beta1/novacell_types.go
```diff
@@ -161,9 +167,19 @@ type NovaCellSpec struct {
 	// the cell.
 	NoVNCProxyServiceTemplate NovaNoVNCProxyTemplate `json:"noVNCProxyServiceTemplate"`
 
+	// +kubebuilder:validation:Required
+	// NovaComputeIronicServiceTemplate - defines the novvncproxy service dedicated for
+	// the cell.
+	NovaComputeIronicServiceTemplate NovaComputeIronicTemplate `json:"novaComputeIronicServiceTemplate"`
```
gibizer, line 173: I thought we wanted to allow multiple instances of the compute in the cell. @SeanMooney am I correct that for compute with Ironic virt driver we want to keep the replicas to 0-1 and allow adding more than one NovaCompute(Ironic)ServiceTemplate to the Cell instead? 
For fake driver we might want to support both increasing the replica count and therefore having more but identically configured fake computes, as well as adding more NovaCompute(Ironic)ServiceTemplate instances to support differently configured computes in a same cell.

api/v1beta1/novacell_types.go
```diff
@@ -161,9 +167,19 @@ type NovaCellSpec struct {
 	// the cell.
 	NoVNCProxyServiceTemplate NovaNoVNCProxyTemplate `json:"noVNCProxyServiceTemplate"`
 
+	// +kubebuilder:validation:Required
+	// NovaComputeIronicServiceTemplate - defines the novvncproxy service dedicated for
```
gibizer, line 171: copy paste: novvncpoxy

api/v1beta1/novacomputeironic_types.go
```diff
@@ -0,0 +1,214 @@
...
+	// +kubebuilder:validation:Optional
+	// ContainerImage - The service specific Container Image URL (will be set to environmental default if empty)
+	ContainerImage string `json:"containerImage"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
+	// +kubebuilder:validation:Minimum=0
+	// Replicas of the service to run
+	Replicas *int32 `json:"replicas"`
```
gibizer, line 41: If we go with the specific NovaComputeIronicTemplate then this should be maximum of 1. If we go with a generic NovaComputeTemplate then the current Replicas definition is OK, but we need a validation  webhook that reject replicas > 1 for NovaComputeTemplate with virt driver == ironic while allows it for virt driver == fake.

gibizer (changes requested):

gibizer:
I only looked at the CRD definitions so far but I think we need to discuss this a bit further.

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=ironic.IronicDriver;fake.FakeDriver
+	// +kubebuilder:default="ironic.IronicDriver"
```
SeanMooney, line 74: i would prefer to remove the default and make this required

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// +kubebuilder:validation:Required
+	// ServiceAccount - service account name used internally to provide Nova services the default SA name
+	ServiceAccount string `json:"serviceAccount"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=ironic.IronicDriver;fake.FakeDriver
+	// +kubebuilder:default="ironic.IronicDriver"
+	// ComputeDriver defines which driver to use for controlling virtualization
+	ComputeDriver string `json:"computeDriver"`
```
SeanMooney, line 137: same here

api/v1beta1/novacompute_webhook.go
```diff
@@ -0,0 +1,100 @@
...
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *NovaCompute) ValidateUpdate(old runtime.Object) error {
+	novacomputelog.Info("validate update", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object update.
+	return nil
+}
```
SeanMooney, line 92: we need to validat that replica is 1 if the compute driver is ironic

api/v1beta1/common_types.go
```diff
@@ -25,14 +25,15 @@ import (
...
-	AnsibleEEContainerImage     = "quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
+	NovaAPIContainerImage           = "quay.io/podified-antelope-centos9/openstack-nova-api:current-podified"
+	NovaConductorContainerImage     = "quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified"
+	NovaMetadataContainerImage      = "quay.io/podified-antelope-centos9/openstack-nova-api:current-podified"
+	NovaNoVNCContainerImage         = "quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified"
+	NovaSchedulerContainerImage     = "quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified"
+	NovaComputeContainerImage       = "quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified"
+	NovaLibvirtContainerImage       = "quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified"
+	AnsibleEEContainerImage         = "quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
+	NovaIronicComputeContainerImage = "quay.io/podified-antelope-centos9/openstack-nova-compute-ironic:current-podified"
```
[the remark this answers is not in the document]
SeanMooney: we do not need a seperate container for ironic.

looking at 
https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/os/nova-base/nova-compute-ironic/nova-compute-ironic.yaml

everything that is required is in 
https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/os/nova-base/nova-compute/nova-compute.yaml

api/v1beta1/novacomputeironic_types.go
```diff
@@ -0,0 +1,214 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NovaComputeIronicTemplate defines the input parameters specified by the user to
+// create a NovaComputeIronic via higher level CRDs.
+type NovaComputeIronicTemplate struct {
```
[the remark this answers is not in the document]
mrKisaoLamb: Added generic template

api/v1beta1/novacompute_types.go
mrKisaoLamb: done, 9e5fe712f3d0fb2ece9a389b80b83ffa1334827c

api/v1beta1/common_types.go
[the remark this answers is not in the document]
gibizer: The removal of the NovaExternalCompute is merged and I removed the images there. So I think we are good here after the necessary rebase.

api/v1beta1/novacell_types.go
```diff
@@ -73,11 +73,16 @@ type NovaCellTemplate struct {
 	// the cell.
 	NoVNCProxyServiceTemplate NovaNoVNCProxyTemplate `json:"noVNCProxyServiceTemplate"`
 
+	// +kubebuilder:validation:Optional
+	// List of nova computes with selected drivers
```
gibizer, line 77: I don't fully understand this field def. The documentation says it is a list but actually it is a map. Also it does not define what is the key of  the map. At least the doc needs to be improved so the user understands how to use this field.

nit: please follow the documentation pattern. The doc text of the field should start with the name of the field documented.

api/v1beta1/common_types.go
```diff
@@ -25,14 +25,15 @@ import (
...
-	NovaLibvirtContainerImage   = "quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified"
-	AnsibleEEContainerImage     = "quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
+	NovaAPIContainerImage           = "quay.io/podified-antelope-centos9/openstack-nova-api:current-podified"
+	NovaConductorContainerImage     = "quay.io/podified-antelope-centos9/openstack-nova-conductor:current-podified"
+	NovaMetadataContainerImage      = "quay.io/podified-antelope-centos9/openstack-nova-api:current-podified"
+	NovaNoVNCContainerImage         = "quay.io/podified-antelope-centos9/openstack-nova-novncproxy:current-podified"
+	NovaSchedulerContainerImage     = "quay.io/podified-antelope-centos9/openstack-nova-scheduler:current-podified"
+	NovaComputeContainerImage       = "quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified"
+	NovaLibvirtContainerImage       = "quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified"
+	AnsibleEEContainerImage         = "quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
```
gibizer, line 35: This is one of the merge conflict caused by the removal of NovaExternalCompute.

api/v1beta1/common_types.go
```diff
@@ -176,9 +178,10 @@ func SetupDefaults() {
 
 	// Acquire environmental defaults and initialize NovaCell defaults with them
 	novaCellDefaults := NovaCellDefaults{
-		ConductorContainerImageURL: util.GetEnvVar("NOVA_CONDUCTOR_IMAGE_URL_DEFAULT", NovaConductorContainerImage),
-		MetadataContainerImageURL:  util.GetEnvVar("NOVA_METADATA_IMAGE_URL_DEFAULT", NovaMetadataContainerImage),
-		NoVNCContainerImageURL:     util.GetEnvVar("NOVA_NOVNC_IMAGE_URL_DEFAULT", NovaNoVNCContainerImage),
+		ConductorContainerImageURL:         util.GetEnvVar("NOVA_CONDUCTOR_IMAGE_URL_DEFAULT", NovaConductorContainerImage),
+		MetadataContainerImageURL:          util.GetEnvVar("NOVA_METADATA_IMAGE_URL_DEFAULT", NovaMetadataContainerImage),
+		NoVNCContainerImageURL:             util.GetEnvVar("NOVA_NOVNC_IMAGE_URL_DEFAULT", NovaNoVNCContainerImage),
+		NovaIronicComputeContainerImageURL: util.GetEnvVar("NOVA_IRONIC_COMPUTE_IMAGE_URL_DEFAULT", NovaIronicComputeContainerImage),
```
gibizer, line 184: L167, L184, and L230 shows a duplication of logic. As the rest of the image URL field initialization is also duplicated I'm OK the follow that pattern here. But I think we should be able to deduplicate this in a separate PR. 
// later
I realized that the propagation of this value is also seems to be duplicated between the top level and the cell level.

api/v1beta1/novacell_webhook.go
```diff
@@ -100,6 +107,16 @@ func (r *NovaCellSpec) validate(basePath *field.Path) field.ErrorList {
 		)
 	}
 
+	if  len(r.NovaComputeTemplates) > 0{
+		for _, computeTemplate := range r.NovaComputeTemplates{
```
gibizer, line 111: I think we can simply this and drop the condition. If the templates map is empty then the iteration will be a noop anyhow. 
Do we need to handle if NovaComputeTemplate is nil?

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=ironic.IronicDriver;fake.FakeDriver
+	// ComputeDriver defines which driver to use for controlling virtualization
+	ComputeDriver string `json:"computeDriver"`
```
gibizer, line 75: I think in most of the production cases the user wants to use Ironic. So to help the user we can make this field optional with an Ironic default. @SeanMooney How do you feel about it?

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+func (s NovaComputeStatus) GetConditions() condition.Conditions {
+	return s.Conditions
+}
+
+// GetSecret returns the value of the NovaCompute.Spec.Secret
+func (n NovaCompute) GetSecret() string {
+	return n.Spec.Secret
+}
+
+// NewNovaComputeSpec constructs a NovaMetadataSpec
```
gibizer, line 194: nit: copy paste NovaMetadataSpec

api/v1beta1/novacompute_webhook.go
gibizer: It is a bit more complicated than that. We have the validation wired up on the Template (see the webhook for the cell). It is again depends on if we allow the user to create a standalone NovaCompute CR. If yes, then we need input validation on the Spec here, if no, then the validation on the template is enough. I'm not sure about the pattern we have here.

gibizer (commented):
I re-reviewed the API definition part. It looks better. I still have some comments but the overall direction is OK to me. I still have to look at the implementation and the test later

api/v1beta1/novacompute_webhook.go
mrKisaoLamb: done

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
```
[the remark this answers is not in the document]
mrKisaoLamb: done

api/v1beta1/novacell_webhook.go
mrKisaoLamb: done

api/v1beta1/novacompute_types.go
mrKisaoLamb: done

config/default/manager_default_images.yaml
```diff
@@ -27,3 +27,5 @@ spec:
           value: quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified
         - name: NOVA_ANSIBLE_EE_IMAGE_URL_DEFAULT
           value: quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest
+        - name: NOVA_IRONIC_COMPUTE_IMAGE_URL_DEFAULT
```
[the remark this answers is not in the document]
mrKisaoLamb: done

api/v1beta1/novacell_types.go
mrKisaoLamb: done, this is now really the map and it use keys in compute name

api/v1beta1/novacompute_types.go
```diff
@@ -0,0 +1,221 @@
...
+	ComputeDriver string `json:"computeDriver"`
+}
+
+// NovaComputeSpec defines the desired state of NovaCompute
+type NovaComputeSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +kubebuilder:validation:Required
+	// CellName is the name of the Nova Cell this novncproxy belongs to.
```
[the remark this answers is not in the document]
mrKisaoLamb: done

controllers/novacell_controller.go
```diff
@@ -261,6 +273,11 @@ func (r *NovaCellReconciler) initConditions(
 				condition.InitReason,
 				novav1.NovaComputeServiceConfigInitMessage,
 			),
+			condition.UnknownCondition(
+				novav1.NovaComputeReadyCondition,
```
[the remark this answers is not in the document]
mrKisaoLamb: change name to 'NovaAllCompusteReadyCondition' and status is map of status per compute node

controllers/novacell_controller.go
```diff
@@ -209,6 +209,18 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		instance.Status.Conditions.Remove(novav1.NovaComputeServiceConfigReady)
 	}
 
+	if len(instance.Spec.NovaComputeTemplates) == 0 {
+		instance.Status.Conditions.Remove(novav1.NovaComputeReadyCondition)
```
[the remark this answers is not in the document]
mrKisaoLamb: done, applied same schema here

controllers/novacell_controller.go
```diff
@@ -451,6 +468,58 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+			novav1.NovaComputeReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, err
+	}
+
+	if op != controllerutil.OperationResultNone {
+		util.LogForObject(h, fmt.Sprintf("NovaCompute %s.", string(op)), instance, "NovaCompute.Name", novacompute.Name)
+	}
+
+	instance.Status.NovaComputeReadyCount = novacompute.Status.ReadyCount
```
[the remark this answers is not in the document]
mrKisaoLamb: done removed this one

controllers/novacell_controller.go
```diff
@@ -451,6 +468,58 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+
+	c := novacompute.Status.Conditions.Mirror(novav1.NovaComputeReadyCondition)
+
+	if c != nil {
+		instance.Status.Conditions.Set(c)
```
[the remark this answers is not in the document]
mrKisaoLamb: done, moved

controllers/novacell_controller.go
```diff
@@ -206,6 +207,34 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		instance.Status.Conditions.Remove(novav1.NovaComputeServiceConfigReady)
 	}
 
+	for computeName, computeTemplate := range instance.Spec.NovaComputeTemplates {
```
gibizer, line 210: I think we can move this whole section before 
`if cellHasVNCService && !instance.Status.Conditions.IsTrue(novav1.NovaNoVNCProxyReadyCondition)`
so that the ironic compute deployment can progress in parallel with the NovaNoVNCProxy deployment to generate the compute config for the EDPM computes.

This will only be relevant for cells with mixed ironic and libvirt computes as for pure ironic cells NovaNoVNCProxy would not be deployed (if the human deployer is smart enough) as ironic does not support vnc proxy.

controllers/novacell_controller.go
```diff
@@ -206,6 +207,34 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+		if instance.Spec.CellName != novav1.Cell0Name {
+			result, err = r.ensureNovaCompute(ctx, h, instance, computeTemplate, computeName)
+			if err != nil {
+				return result, err
+			}
+			instance.Status.NovaComputesStatuses[computeName] = true
```
gibizer, line 216: I think at this point you only want to add the key to the status map, but with bool = false.
We have to model 3 states:
1. compute CR is created but not mapped (`nova-manage cell_v2 discover_hosts` isn't run): the key exists but with value false
2. compute CR is created and mapped: the key exists and the value is true
3. compute CR does not exist: the key does not exists

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
 	return ctrl.Result{}, nil
 }
 
+func getNovaComputeName(instance client.Object, computeName string) types.NamespacedName {
+	return types.NamespacedName{Namespace: instance.GetNamespace(), Name: instance.GetName() + "-" + computeName + "-compute"}
```
gibizer, line 607: We need to limit computeName in the API with a validation to avoid generating a too long CR name here.

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+		// Nothing to do as it does not exists
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	// If it is not created by us, we don't touch it
+	if !OwnedBy(compute, instance) {
+		util.LogForObject(
+			h, "NovaCompute is disabled, but there is a "+
```
gibizer, line 629: "NovaCompute isn't defined in the cell, but there is a "

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+		return nil
+	}
+
+	// OK this was created by us so we go and delete it
+	err = r.Client.Delete(ctx, compute)
+	if err != nil && k8s_errors.IsNotFound(err) {
+		return nil
+	}
+	util.LogForObject(
+		h, "NovaCompute is disabled, so deleted NovaCompute",
```
gibizer, line 641: "NovaCompute isn't defined in the cell...

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+	// One can think that this means the cell adopts the NovaCompute and
+	// starts managing it.
+	// However at least the label selector of the StatefulSet spec is
+	// immutable, so the controller cannot simply start adopting an existing
+	// NovaCompute. But instead the our CR will be in error state until the
+	// human deletes the manually created NovaCompute and then Nova will
+	// create its own NovaComputes.
+	// See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates
+
+	computeNovaName := getNovaComputeName(instance, computeName)
```
gibizer, line 664: supernit: I liked the `fullComputeName` local variable name in the delete function :)

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+		return ctrl.Result{}, err
+	}
+
+	// If it is not created by us, we don't touch it
+	if !k8s_errors.IsNotFound(err) && !OwnedBy(novacompute, instance) {
+		err := fmt.Errorf(
+			"cannot update NovaCompute/%s as the cell is not owning it", novacompute.Name)
+		util.LogErrorForObject(
+			h, err,
+			"NovaCompute is enabled, but there is a "+
```
gibizer, line 677: NovaCompute is defined in the cell,

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+				"NovaCompute CR not owned by us. We cannot update it. "+
+				"Please delete the NovaCompute.",
+			instance, "NovaCompute", computeNovaName)
+		instance.Status.Conditions.Set(condition.FalseCondition(
```
gibizer, line 681: ensureNovaCompute is called in a loop for each compute defined, so if two those calls fails then the second overrides the condition set by the first call. If you want to have a single condition for a list of computes then you have to aggregate the result of the ensureNovaCompute calls and set the aggregated result to the condition.

controllers/novacell_controller.go
```diff
@@ -566,6 +603,129 @@ func (r *NovaCellReconciler) ensureComputeConfig(
...
+		err := controllerutil.SetControllerReference(instance, novacompute, r.Scheme)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
```
gibizer, line 713: ditto this Set call will override the result of the previous compute in the list in the same cell

gibizer (commented):
We are moving to the right direction. I have couple comments on the CRD, validation, and condition handling. I haven't reviewed the test coverage yet but I like the fact that it is green.

api/v1beta1/novacell_types.go
```diff
@@ -160,6 +165,12 @@ type NovaCellSpec struct {
 	// the cell.
 	NoVNCProxyServiceTemplate NovaNoVNCProxyTemplate `json:"noVNCProxyServiceTemplate"`
 
+	// +kubebuilder:validation:Optional
+	// NovaComputeTemplates - map of nova computes template with selected drivers in format
+	// compute_name: compute_template. Key from map is arbitrary name for the compute.
+	// because of that there is a limit 20 signs
```
gibizer, line 171: nit: There is a 20 character limit on the compute name.

controllers/novacell_controller.go
```diff
@@ -206,6 +215,24 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+		}
+	}
+
+	if len(instance.Spec.NovaComputeTemplates) == 0 {
+		instance.Status.Conditions.Remove(novav1.NovaAllComputesReadyCondition)
+	} else {
+		instance.Status.Conditions.MarkTrue(
+			novav1.NovaAllComputesReadyCondition, condition.ServiceConfigReadyMessage,
+		)
+	}
```
gibizer, line 235: Sorry I wasn't clear before. I would move this also before the compute config generation to keep the NovaCompute logic in a single place

controllers/novacompute_controller.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NovaComputeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&novav1.NovaCompute{}).
+		Owns(&v1.StatefulSet{}).
+		Owns(&corev1.Service{}).
+		Owns(&corev1.Secret{}).
+		Owns(&routev1.Route{}).
```
gibizer, line 408: we don't create any routes here

controllers/novacompute_controller.go
```diff
@@ -0,0 +1,412 @@
...
+		CellSelector:       cell,
+	}
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NovaComputeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&novav1.NovaCompute{}).
+		Owns(&v1.StatefulSet{}).
+		Owns(&corev1.Service{}).
```
gibizer, line 406: we don't create any service here

pkg/novacompute/const.go
```diff
@@ -0,0 +1,27 @@
...
+package novacompute
+
+const (
+	// MergedServiceConfigPath - The location of the merged configuration file
+	// for the service
+	MergedServiceConfigPath = "/var/lib/openstack/config/nova-compute.json"
+	//APIServicePort - The port the nova-api service is exposed on
+	APIServicePort = 8774
+	// ServiceName - The name of the service exposed to k8s
+	ServiceName = "nova-compute"
```
gibizer, line 26: I don't see any usage of these constants

pkg/novacompute/deployment.go
```diff
@@ -0,0 +1,192 @@
...
+			Command: []string{
+				"/bin/true",
+			},
+		}
+		startupProbe.Exec = &corev1.ExecAction{
+			Command: []string{
+				"/bin/true",
+			},
+		}
+	}
```
gibizer, line 95: You can use a the check what we use for the conductor, i.e. checking if the process exists `"/usr/bin/pgrep", "-r", "DRST", "nova-conductor",`

pkg/novacompute/deployment.go
```diff
@@ -0,0 +1,192 @@
...
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: instance.Spec.ServiceAccount,
+					Volumes: []corev1.Volume{
+						nova.GetConfigVolume(nova.GetServiceConfigSecretName(instance.Name)),
+						nova.GetLogVolume(),
+					},
+					Containers: []corev1.Container{
+						// the first container in a pod is the default selected
+						// by oc log so define the log stream container first.
```
gibizer, line 135: We only need the sidecar container for logging from wsgi services. For directly started processes like this (and conductor, scheduler) we can let the process to log to its standard out and k8s will capture it without issue,

templates/nova.conf
```diff
@@ -13,7 +13,14 @@ log_file = {{ .log_file }}
 {{end}}
 debug=true
 {{if eq .service_name "nova-compute"}}
-compute_driver=libvirt.LibvirtDriver
+{{ if (index . "compute_driver") }}
```
gibizer, line 16: I would simplify this by adding this template parameter to the EDPM compute config generation codepath too and hardcode it to libvirt.LibvirtDriver there: https://github.com/openstack-k8s-operators/nova-operator/blob/532bc7f74c7f8893ad9ab53690bed442c843619b/controllers/novacell_controller.go#L573-L584

templates/nova.conf
```diff
@@ -117,9 +125,14 @@ driver = noop
 notify_on_state_change = vm_and_task_state
 {{ end }}
 
-{{ if eq .service_name "nova-compute" "nova-novncproxy"}}
+{{ if eq .service_name "nova-novncproxy" "nova-compute"}}
 [vnc]
+{{if eq .service_name "nova-compute"}}
+# TODO(ksambor) in tripleo we disable vnc console
```
gibizer, line 131: This feels strange. Do you mean vnc is disable for ironic computes or all computes in tripleo?

gibizer:
I still haven't reviewed the test coverage yet. 
Also during my review I noted the following missing pieces:
* having a struct for compute status
* AllComputeReady aggregate condition handling
* running discover hosts after computes are ready

templates/nova.conf
mrKisaoLamb: we disable only for ironic driver not all computes

gibizer: Ahh but this now disable for all computes as far as I understand.
I agree to disable it for ironic computes

api/v1beta1/novacompute_webhook.go
```diff
@@ -0,0 +1,159 @@
...
+		)
+	}
+
+	return errors
+}
+
+// ValidateReplicas validates replicas depend on compute driver
+func (r *NovaComputeTemplate) ValidateIronicDriverReplicas(basePath *field.Path) field.ErrorList {
+	var errors field.ErrorList
+	if *r.Replicas > 1 {
```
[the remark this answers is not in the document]
gibizer: thanks. It seems that when the webhook deserialize the request into NovaCompueTemplate struct the kubebuilder default, which is 1, applied.

gibizer:
recheck

controllers/novacompute_controller.go
```diff
@@ -0,0 +1,405 @@
...
+		"keystone_internal_url":  instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":     instance.Spec.ServiceUser,
+		"nova_keystone_password": string(secret.Data[ServicePasswordSelector]),
+		"openstack_cacert":       "",          // fixme
+		"openstack_region_name":  "regionOne", // fixme
+		"default_project_domain": "Default",   // fixme
+		"default_user_domain":    "Default",   // fixme
+		"transport_url":          string(secret.Data[TransportURLSelector]),
+		"log_file":               "/var/log/nova/nova-compute.log",
+		"nova_compute_image":     instance.Spec.ContainerImage,
```
gibizer, line 276: Why do we need this in the nova-compute config?

controllers/novacompute_controller.go
```diff
@@ -0,0 +1,405 @@
...
+	instance *novav1.NovaCompute,
+) error {
+	util.LogForObject(h, "Reconciling delete", instance)
+
+	// Successfully cleaned up everything. So as the final step let's remove the
+	// finalizer from ourselves to allow the deletion of NovaAPI CR itself
+	updated := controllerutil.RemoveFinalizer(instance, h.GetFinalizer())
+	if updated {
+		util.LogForObject(h, "Removed finalizer from ourselves", instance)
+	}
```
gibizer, line 383: As far as I see currently there is no finalizer added to NovaCompute. Either you add a finalizer at the beginning of Reconcile or remove the reconcileDelete codepath.

pkg/nova/common.go
```diff
@@ -89,6 +90,15 @@ const (
...
 	CellFailed CellDeploymentStatus = iota
+	// ComputeDiscovering indicates that NovaCompute reached the Ready status and it is
+	// being discover
+	ComputeDiscovering NovaComputeStatus = iota
+	// ComputeDiscoverFailed indicates that NovaCompute reached the Ready status but
+	// discover it failed
+	ComputeDiscoverFailed NovaComputeStatus = iota
+	// ComputeDiscoverReady indicates that NovaCompute reached the Ready status and
+	// it is discovered
+	ComputeDiscoverReady NovaComputeStatus = iota
```
gibizer, line 101: Please put this in a separate const() block to reset the iota value back to 0 as this is a separate enum from the cell status.

controllers/novacell_controller.go
```diff
@@ -160,6 +166,81 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+			"all computes ready", allComputesReady)
+	}
+
+	computeTemplatesHash, err := hashOfStringMap(computeTemplatesHashMap)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// When the nova compute is ready we need to discover service
+	if instance.Status.Hash[novav1.ComputeDiscoverHashKey] != computeTemplatesHash && allComputeDeployed {
```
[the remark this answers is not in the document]
gibizer: See https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/test/functional/job_test.go for details about how NewJob and DoJob can be used to run a Job only once if none of the input of the job changes.

controllers/novacell_controller.go
```diff
@@ -160,6 +166,81 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 
 	}
 
+	// manage nova-computes with selected driver based on created template
+	// we need to keep statuses for deleting and discover action and also create map with hashes
+	// to run discover job only when all computes are deployed and never discovered
+	computeTemplatesHashMap := make(map[string]string)
+	novaComputeImage := ""
+	for computeName, computeTemplate := range instance.Spec.NovaComputeTemplates {
```
gibizer, line 174: Please be aware that iterating a map in golang is done in an undefined order that can also change between each iteration. So if there is something that depends on the order of the iteration the you have to sort manually. I think at the moment it is not an issue as computes are independent. But if you follow my suggestion below to use the `readyCompute` list to make the Job's hash then you need to ensure that the list is sorted before hashed otherwise the different reconcile runs will generate the same list in different orders that will lead to differnt hashes.

gibizer:
About the CI failures:
```
[pod/nova-operator-controller-manager-6454b77f47-kmss2/manager] 2023-09-27T11:07:39.660Z	ERROR	controller-runtime.source	if kind is a CRD, it should be installed before calling Start	{"kind": "NovaCompute.nova.openstack.org", "error": "no matches for kind \"NovaCompute\" in version \"nova.openstack.org/v1beta1\""}
```
This is because the build in CI does not bump the nova-operator version in the openstack-operator and therefore the new NovaCompute CRD is never installed.

test/functional/nova_compute_ironic_controller_test.go
```diff
@@ -0,0 +1,530 @@
...
+		})
+	})
+})
+
+var _ = Describe("NovaCompute with ironic diver controller", func() {
+
+	When("with configure cellname", func() {
+		BeforeEach(func() {
+			spec := GetDefaultNovaComputeSpec(cell1)
+			spec["cellName"] = "some-cell-name"
```
gibizer, line 258: if you pass cell1 to GetDefaultNovaComputeSpec then it already sets the cellName field on the spec so you don't need to do it here again.

test/functional/nova_compute_ironic_controller_test.go
```diff
@@ -0,0 +1,530 @@
...
+		BeforeEach(func() {
+			DeferCleanup(
+				k8sClient.Delete, ctx, CreateCellInternalSecret(cell1))
+
+			spec := GetDefaultNovaComputeSpec(cell1)
+			spec["replicas"] = 0
+			novaCompute := CreateNovaCompute(novaNames.NovaComputeName, spec)
+			DeferCleanup(th.DeleteInstance, novaCompute)
+		})
+		It("and deployment is Ready", func() {
```
gibizer, line 505: OK this is another case we need to handle when doing the host discovery. If the compute is Ready because it was created with 0 replicas then we can run a discovery but it will not discovery anything. When the replica changes to >0 we need to discover again. But you are hashing the template so I guess ti will work.

test/functional/novacell_controller_test.go
```diff
@@ -149,6 +150,9 @@ var _ = Describe("NovaCell controller", func() {
 			spec["metadataServiceTemplate"] = map[string]interface{}{
 				"enabled": true,
 			}
+			spec["novaComputeTemplates"] = map[string]interface{}{
+				"ironic-compute": GetDefaultNovaComputeTemplate(),
```
gibizer, line 154: you have a constant for "ironic-compute" in base_test.go

test/functional/novacell_controller_test.go
```diff
@@ -328,20 +332,26 @@ var _ = Describe("NovaCell controller", func() {
...
-
+			cell := GetNovaCell(cell1.CellCRName)
 			th.ExpectCondition(
 				cell1.CellCRName,
 				ConditionGetterFunc(NovaCellConditionGetter),
 				condition.ReadyCondition,
 				corev1.ConditionTrue,
 			)
+
+			Expect(cell.Status.NovaComputesStatus).To(HaveKey("ironic-compute"))
```
gibizer, line 347: use the constant from base_test.go for the name

config/crd/bases/nova.openstack.org_novacomputeironics.yaml
gibizer, line 1: This seems to be a remnant of before the rename as it is called novacomputeironics

controllers/novacompute_controller.go
```diff
@@ -0,0 +1,405 @@
...
+		"service_name":           NovaComputeLabelPrefix,
+		"keystone_internal_url":  instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":     instance.Spec.ServiceUser,
+		"nova_keystone_password": string(secret.Data[ServicePasswordSelector]),
+		"openstack_cacert":       "",          // fixme
+		"openstack_region_name":  "regionOne", // fixme
+		"default_project_domain": "Default",   // fixme
+		"default_user_domain":    "Default",   // fixme
+		"transport_url":          string(secret.Data[TransportURLSelector]),
+		"log_file":               "/var/log/nova/nova-compute.log",
```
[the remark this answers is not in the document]
gibizer: I just realized that this means the pod does not log anything to the standard out. In a follow up PR please change this to allow the podified compute service to log to the standard out. You can take nova-scheduler or nova-conductor as an example

pkg/nova/common.go
```diff
@@ -69,6 +69,7 @@ const (
 )
 
 type CellDeploymentStatus int
+type NovaComputeStatus int
```
gibizer, line 72: nit (fix it in a separate PR): this is unused now

test/functional/novacell_controller_test.go
gibizer, line 1: in a different PR: you can assert here in this test suite that the instance.Status.NovaComputesStatus is filled properly.

gibizer (approved):
This works for me locally with the fake driver ( I can deploy multiple computes and boot fake VMs) and does not seem to break any existing functionality. @mrKisaoLamb tested it with ironic. So I think this OK to land now. There is a list of followup in previous comments that we will address separately and iteratively. 

Thanks @mrKisaoLamb for making this happen.

PR #485, opened 2023-08-07
Use inventory as a secret instead of configmap

fao89:
Changing it into secret, as the inventory may require secret values.
ref: https://github.com/openstack-k8s-operators/dataplane-operator/pull/330


gibizer:
Is this still relevant now that we merged https://github.com/openstack-k8s-operators/nova-operator/pull/452 ?

fao89:
no, it isn't, closing it

PR #486, opened 2023-08-08
Add jamepark4 to compute approvers

jamepark4:
Add jamepark4 to compute approvers.


SeanMooney (approved):
yep you have been added to the openshift org 
https://github.com/jamepark4 and to the compute team so approving

PR #491, opened 2023-08-11
Unpin deps after dev-preview

gibizer:
On the main branch we want to continue tracking the main branch of our operator dependencies.


gibizer:
recheck 
timeout during build but I don't actually see in the logs what timed out

SeanMooney (approved):
i might need to bump the time out on the content provider job if we continue to see timeouts but i have not seen many in the past so ill keep an eye on that. over all this makes sense to me

PR #496, opened 2023-08-22
switch to structured logging

pinikomarov:
This automatically adds additional fields like reconcile_id etc.. from the controller context.

before :

`2023-05-18T01:53:14+03:00 INFO  controllers.KeystoneAPI Reconciled Service init successfully`

after:

`2023-05-18T02:00:28+03:00       INFO    Controllers.KeystoneAPI Reconciled Service init successfully    {"controller": "keystoneapi", "controllerGroup": "keystone.openstack.org", "controllerKind":"KeystoneAPI", "KeystoneAPI": {"name":"keystone","namespace":"openstack"}, "namespace": "openstack", "name": "keystone", "reconcileID": "512a4d9b-f31d-4fa4-a4cc-cd6c13e4455d"}
`

*by using per reconcile function respective logger objects, the intention is to lay the ground for parallel reconciliation in future and avoid race conditions as ctx objects are reconcile run specific.

*example logs lines  are currently from keystone operator [#220_patch](https://github.com/openstack-k8s-operators/keystone-operator/pull/220) until this operator deployment patch testing  is is done

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/go-logr/logr"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
@@ -57,6 +58,11 @@ import (
 // NovaReconciler reconciles a Nova object
 type NovaReconciler struct {
 	ReconcilerBase
+}
+
+// getlog returns a logger object with a prefix of "conroller.name" and aditional controller context fields
+func GetLog(ctx context.Context, controller string) logr.Logger {
+	return log.FromContext(ctx).WithName("Controllers").WithName(controller)
 }
 
 //+kubebuilder:rbac:groups=nova.openstack.org,resources=nova,verbs=get;list;watch;create;update;patch;delete
@@ -87,7 +93,7 @@ type NovaReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/reconcile
 func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	l := log.FromContext(ctx)
+	l := GetLog(ctx, "nova")
 
 	// Fetch the NovaAPI instance that needs to be reconciled
 	instance := &novav1.Nova{}
@@ -110,7 +116,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		r.Client,
 		r.Kclient,
 		r.Scheme,
-		r.Log,
+		l,
 	)
 	if err != nil {
 		l.Error(err, "Failed to create lib-common Helper")
@@ -1256,6 +1262,8 @@ func (r *NovaReconciler) ensureCellMapped(
 	cellTemplate novav1.NovaCellTemplate,
 	apiDBHostname string,
 ) (nova.CellDeploymentStatus, error) {
+	l := GetLog(ctx, "nova")
+
 	ospSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
 	if err != nil {
 		return nova.CellMappingFailed, err
@@ -1383,7 +1391,7 @@ func (r *NovaReconciler) ensureCellMapped(
 	// information to the top level services so that each service can restart
 	// their Pods if a new cell is registered or an existing cell is updated.
 	instance.Status.RegisteredCells[cell.Name] = job.GetHash()
-	r.Log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.RegisteredCells[cell.Name]))
+	l.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.RegisteredCells[cell.Name]))
 
 	return nova.CellMappingReady, nil
 }
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -1391,7 +1391,7 @@ func (r *NovaReconciler) ensureCellMapped(
 	// information to the top level services so that each service can restart
 	// their Pods if a new cell is registered or an existing cell is updated.
 	instance.Status.RegisteredCells[cell.Name] = job.GetHash()
-	l.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.RegisteredCells[cell.Name]))
+	l.Info("Job hash added ", "job", jobDef.Name, "hash", instance.Status.RegisteredCells[cell.Name])
 
 	return nova.CellMappingReady, nil
 }
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -93,7 +93,7 @@ func GetLog(ctx context.Context, controller string) logr.Logger {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/reconcile
 func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	l := GetLog(ctx, "nova")
+	log := GetLog(ctx, "nova")
 
 	// Fetch the NovaAPI instance that needs to be reconciled
 	instance := &novav1.Nova{}
@@ -103,11 +103,11 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers. Return and don't requeue.
-			l.Info("Nova instance not found, probably deleted before reconciled. Nothing to do.")
+			log.Info("Nova instance not found, probably deleted before reconciled. Nothing to do.")
 			return ctrl.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		l.Error(err, "Failed to read the Nova instance.")
+		log.Error(err, "Failed to read the Nova instance.")
 		return ctrl.Result{}, err
 	}
 
@@ -116,10 +116,10 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		r.Client,
 		r.Kclient,
 		r.Scheme,
-		l,
+		log,
 	)
 	if err != nil {
-		l.Error(err, "Failed to create lib-common Helper")
+		log.Error(err, "Failed to create lib-common Helper")
 		return ctrl.Result{}, err
 	}
 	util.LogForObject(h, "Reconciling", instance)
@@ -1262,7 +1262,7 @@ func (r *NovaReconciler) ensureCellMapped(
 	cellTemplate novav1.NovaCellTemplate,
 	apiDBHostname string,
 ) (nova.CellDeploymentStatus, error) {
-	l := GetLog(ctx, "nova")
+	log := GetLog(ctx, "nova")
 
 	ospSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
 	if err != nil {
@@ -1391,7 +1391,7 @@ func (r *NovaReconciler) ensureCellMapped(
 	// information to the top level services so that each service can restart
 	// their Pods if a new cell is registered or an existing cell is updated.
 	instance.Status.RegisteredCells[cell.Name] = job.GetHash()
-	l.Info("Job hash added ", "job", jobDef.Name, "hash", instance.Status.RegisteredCells[cell.Name])
+	log.Info("Job hash added ", "job", jobDef.Name, "hash", instance.Status.RegisteredCells[cell.Name])
 
 	return nova.CellMappingReady, nil
 }
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -60,9 +60,9 @@ type NovaReconciler struct {
 	ReconcilerBase
 }
 
-// getlog returns a logger object with a prefix of "conroller.name" and aditional controller context fields
-func GetLog(ctx context.Context, controller string) logr.Logger {
-	return log.FromContext(ctx).WithName("Controllers").WithName(controller)
+// getlogger returns a logger object with a prefix of "conroller.name" and aditional controller context fields
+func (r *NovaReconciler) GetLogger(ctx context.Context) logr.Logger {
+	return log.FromContext(ctx).WithName("Controllers").WithName("Nova")
 }
 
 //+kubebuilder:rbac:groups=nova.openstack.org,resources=nova,verbs=get;list;watch;create;update;patch;delete
@@ -93,7 +93,7 @@ func GetLog(ctx context.Context, controller string) logr.Logger {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/reconcile
 func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	log := GetLog(ctx, "nova")
+	Log := r.GetLogger(ctx)
 
 	// Fetch the NovaAPI instance that needs to be reconciled
 	instance := &novav1.Nova{}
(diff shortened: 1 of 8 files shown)


controllers/novaapi_controller.go
```diff
@@ -216,7 +215,7 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 
 	// Only expose the service is the deployment succeeded
 	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
-		util.LogForObject(h, "Waiting for the Deployment to become Ready before exposing the sevice in Keystone", instance)
+		l("Waiting for the Deployment to become Ready before exposing the sevice in Keystone")
```
gibizer, line 218: I think you wanted to say l.Info(...

controllers/novaapi_controller.go
```diff
@@ -388,10 +386,12 @@ func (r *NovaAPIReconciler) ensureDeployment(
 	inputHash string,
 	annotations map[string]string,
 ) (ctrl.Result, error) {
+	l := GetLog(ctx, "novaapi")
+
 	ss := statefulset.NewStatefulSet(novaapi.StatefulSet(instance, inputHash, getAPIServiceLabels(), annotations), r.RequeueTimeout)
 	ctrlResult, err := ss.CreateOrPatch(ctx, h)
 	if err != nil && !k8s_errors.IsNotFound(err) {
-		util.LogErrorForObject(h, err, "Deployment failed", instance)
+		l.Error(err, "Deployment failed", "instance", instance)
```
gibizer, line 394: You don't need to pass the instance that is already part of the logger.

controllers/nova_controller.go
```diff
@@ -1383,7 +1391,7 @@ func (r *NovaReconciler) ensureCellMapped(
 	// information to the top level services so that each service can restart
 	// their Pods if a new cell is registered or an existing cell is updated.
 	instance.Status.RegisteredCells[cell.Name] = job.GetHash()
-	r.Log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.RegisteredCells[cell.Name]))
+	l.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.RegisteredCells[cell.Name]))
```
gibizer, line 1394: to make it structured it should look something like:
```golang
l.Info("Job hash added", "job", jobDef.Name, "hash", instance.Status.RegisteredCells[cell.Name])
```

controllers/novaapi_controller.go
```diff
@@ -400,7 +400,7 @@ func (r *NovaAPIReconciler) ensureDeployment(
 			err.Error()))
 		return ctrlResult, err
 	} else if (ctrlResult != ctrl.Result{} || k8s_errors.IsNotFound(err)) {
-		util.LogForObject(h, "Deployment in progress", instance)
+		l.Info("Deployment in progress", "instance", instance)
```
gibizer, line 403: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -439,10 +439,10 @@ func (r *NovaAPIReconciler) ensureDeployment(
 	}
 
 	if instance.Status.ReadyCount > 0 {
-		util.LogForObject(h, "Deployment is ready", instance)
+		l.Info("Deployment is ready", "instance", instance)
```
gibizer, line 442: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -439,10 +439,10 @@ func (r *NovaAPIReconciler) ensureDeployment(
...
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
 	} else {
-		util.LogForObject(h, "Deployment is not ready", instance, "Status", ss.GetStatefulSet().Status)
+		l.Info("Deployment is not ready", "instance", instance, "Status", ss.GetStatefulSet().Status)
```
gibizer, line 445: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -570,7 +572,7 @@ func (r *NovaAPIReconciler) ensureKeystoneEndpointDeletion(
 	if err = h.GetClient().Update(ctx, endpoint); err != nil && !k8s_errors.IsNotFound(err) {
 		return err
 	}
-	util.LogForObject(h, "Removed finalizer from nova KeystoneEndpoint", instance)
+	l.Info("Removed finalizer from nova KeystoneEndpoint", "instance", instance)
```
gibizer, line 575: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -580,7 +582,9 @@ func (r *NovaAPIReconciler) reconcileDelete(
 	h *helper.Helper,
 	instance *novav1.NovaAPI,
 ) error {
-	util.LogForObject(h, "Reconciling delete", instance)
+	l := GetLog(ctx, "novaapi")
+
+	l.Info("Reconciling delete", "instance", instance)
```
gibizer, line 587: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -591,10 +595,10 @@ func (r *NovaAPIReconciler) reconcileDelete(
 	// finalizer from ourselves to allow the deletion of NovaAPI CR itself
 	updated := controllerutil.RemoveFinalizer(instance, h.GetFinalizer())
 	if updated {
-		util.LogForObject(h, "Removed finalizer from ourselves", instance)
+		l.Info("Removed finalizer from ourselves", "instance", instance)
```
gibizer, line 598: You don't need to pass the instance that is already part of the logger.

controllers/novaapi_controller.go
```diff
@@ -591,10 +595,10 @@ func (r *NovaAPIReconciler) reconcileDelete(
...
 	}
 
-	util.LogForObject(h, "Reconciled delete successfully", instance)
+	l.Info("Reconciled delete successfully", "instance", instance)
```
gibizer, line 601: You don't need to pass the instance that is already part of the logger.

controllers/novacell_controller.go
```diff
@@ -82,13 +80,13 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
 		r.Scheme,
-		r.Log,
+		l,
 	)
 	if err != nil {
 		l.Error(err, "Failed to create lib-common Helper")
 		return ctrl.Result{}, err
 	}
-	util.LogForObject(h, "Reconciling", instance)
+	l.Info("Reconciling", "instance", instance)
```
gibizer, line 89: You don't need to pass the instance that is already part of the logger.

controllers/novacell_controller.go
```diff
@@ -183,11 +181,7 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 	// However NovaNoVNCProxy is never deployed in cell0, and optional in other
 	// cells too.
 	if cellHasVNCService && !instance.Status.Conditions.IsTrue(novav1.NovaNoVNCProxyReadyCondition) {
-		util.LogForObject(
-			h,
-			"Waiting for the NovaNoVNCProxyService to become Ready before "+
-				"generating the compute config", instance,
-		)
+		l.Info("Waiting for the NovaNoVNCProxyService to become Ready before generating the compute config", "instance", instance)
```
gibizer, line 184: You don't need to pass the instance that is already part of the logger.

controllers/novacell_controller.go
```diff
@@ -209,7 +203,7 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		instance.Status.Conditions.Remove(novav1.NovaComputeServiceConfigReady)
 	}
 
-	util.LogForObject(h, "Successfully reconciled", instance)
+	l.Info("Successfully reconciled", "instance", instance)
```
gibizer, line 206: You don't need to pass the instance that is already part of the logger.

controllers/novacell_controller.go
```diff
@@ -305,7 +301,7 @@ func (r *NovaCellReconciler) ensureConductor(
 	}
 
 	if op != controllerutil.OperationResultNone {
-		util.LogForObject(h, fmt.Sprintf("NovaConductor %s.", string(op)), instance, "NovaConductor.Name", conductor.Name)
+		l.Info("", "NovaConductor", string(op), "instance", instance, "NovaConductor.Name", conductor.Name)
```
gibizer, line 304: The empty log seems strange. I would do the following instead:
```golang
l.Info(fmt.Sprintf("NovaConductor %s.", string(op)))
```
Both the conductor name and the instance is part of the logger.

controllers/novacell_controller.go
```diff
@@ -358,7 +356,7 @@ func (r *NovaCellReconciler) ensureNoVNCProxy(
 	}
 
 	if op != controllerutil.OperationResultNone {
-		util.LogForObject(h, fmt.Sprintf("NovaNoVNCProxy %s.", string(op)), instance, "NovaNoVNCProxy.Name", novncproxy.Name)
+		l.Info("", "NovaNoVNCProxy", string(op), "instance", instance, "NovaNoVNCProxy.Name", novncproxy.Name)
```
gibizer, line 359:
```golang
l.Info(fmt.Sprintf("NovaNoVNCProxy %s.", string(op)))
```

controllers/novacell_controller.go
```diff
@@ -410,7 +410,7 @@ func (r *NovaCellReconciler) ensureMetadata(
 	}
 
 	if op != controllerutil.OperationResultNone {
-		util.LogForObject(h, fmt.Sprintf("NovaMetadata %s.", string(op)), instance, "NovaMetadata.Name", metadata.Name)
+		l.Info("", "NovaMetadata", string(op), "instance", instance, "NovaMetadata.Name", metadata.Name)
```
gibizer, line 413:
```golang
l.Info(fmt.Sprintf("NovaMetadata %s.", string(op)))
```

controllers/novaconductor_controller.go
```diff
@@ -90,13 +89,13 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		r.Scheme,
-		r.Log,
+		l,
 	)
 	if err != nil {
 		l.Error(err, "Failed to create lib-common Helper")
 		return ctrl.Result{}, err
 	}
-	util.LogForObject(h, "Reconciling", instance)
+	l.Info("Reconciling", "instance", instance)
```
gibizer, line 98: you can drop the instance it is already part of the logger

controllers/novaconductor_controller.go
```diff
@@ -186,7 +185,7 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return result, err
 	}
 
-	util.LogForObject(h, "Successfully reconciled", instance)
+	l.Info("Successfully reconciled", "instance", instance)
```
gibizer, line 188: you can drop the instance it is already part of the logger

controllers/novaconductor_controller.go
```diff
@@ -278,15 +277,16 @@ func (r *NovaConductorReconciler) generateConfigs(
...
 	messageBusSecret := &corev1.Secret{}
 	secretName := types.NamespacedName{
 		Namespace: instance.Namespace,
 		Name:      instance.Spec.CellMessageBusSecretName,
 	}
 	err := h.GetClient().Get(ctx, secretName, messageBusSecret)
 	if err != nil {
-		util.LogForObject(
-			h, "Failed reading Secret", instance,
+		l.Info("Failed reading Secret", "instance", instance,
```
gibizer, line 289: you can drop the instance it is already part of the logger

controllers/nova_controller.go
```diff
@@ -1256,6 +1262,8 @@ func (r *NovaReconciler) ensureCellMapped(
 	cellTemplate novav1.NovaCellTemplate,
 	apiDBHostname string,
 ) (nova.CellDeploymentStatus, error) {
+	l := GetLog(ctx, "nova")
```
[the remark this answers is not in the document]
SeanMooney: is this really needed.
why cant we use r.Log since that is already created via 
GetLog(ctx, "nova")

SeanMooney (changes requested):

controllers/nova_controller.go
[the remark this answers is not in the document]
pinikomarov: I believe we discussed it here : 
It's to do with parallel reconcile use case and making room for differeantianting each reconcile_id if that happens.
https://github.com/openstack-k8s-operators/keystone-operator/pull/220#issuecomment-1483117420

controllers/nova_controller.go
pinikomarov: ack

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novaapi_controller.go
pinikomarov: done

controllers/novacell_controller.go
pinikomarov: done

controllers/novacell_controller.go
pinikomarov: done

controllers/novacell_controller.go
pinikomarov: done

controllers/novacell_controller.go
pinikomarov: done

controllers/novacell_controller.go
pinikomarov: done

controllers/novaconductor_controller.go
pinikomarov: done

controllers/novaconductor_controller.go
pinikomarov: done

controllers/novaconductor_controller.go
pinikomarov: done

controllers/nova_controller.go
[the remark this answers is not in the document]
SeanMooney: ok but we dont plan to do that.

if were are not going to use that logger then done use l as the log var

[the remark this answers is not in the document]
pinikomarov: Since this is a wide set of patches , ad they should be the same across all operators, I'm for the path of least resistance here , @SeanMooney I believe @gibizer mentioned you brought the parallel usage issue up , so if we're not ever going to use that , then we can use the receiver to hold the logger object, the plus side is less lines of code, and the negative is if we do use parallelism then we have to revisit this set of patches again.
 @gibizer wdyt ?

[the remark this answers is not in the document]
gibizer: > why cant we use r.Log since that is already created via

r.Log is not request local but reconciler global. But the logger we get from the context has request specific information attached. Anything request specific we store in the reconciler will promote a dangerous pattern. Not just because it prevents parallelism (that we don't use today but might want to use in the future) but also because  data in the reconciler struct automatically passed between two subsequent request. 

> if were are not going to use that logger then done use l as the log var

I'm not sure I can follow.  Does your comment about using a longer name as the variable name for the logger?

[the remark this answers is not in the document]
SeanMooney: yes im asking for the variable to be changed

l := GetLog(ctx, "nova")

to

log := GetLog(ctx, "nova")

or 

logger := GetLog(ctx, "nova")

I'm generally not ok with single character variable names unless its a loop variable
and env then it generally better to use a real name.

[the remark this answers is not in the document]
gibizer: In general I agree that the specificity of the name of a variable should be related to the size of its context. In this specific case I would be OK to name this as "l" as convention. But I'm OK to rename it to "log" too.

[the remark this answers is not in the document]
pinikomarov: I think to use it like:

	info, error := GetLog(ctx, "novaapi").Info, GetLog(ctx, "novaapi").Error

	if err != ...

			info("NovaAPI instance not found, probably deleted before reconciled. Nothing to do.")
		}
		error(err, "Failed to read the NovaAPI instance.")



*and make lib-common's func NewHelper drop the log instantiation parameter, since it's being replaced anyway by these patches

wdyt ?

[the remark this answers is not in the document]
gibizer: I would not go to that direction. Let's name it `log` as a compromise

pinikomarov:
continued in : 
https://github.com/openstack-k8s-operators/nova-operator/pull/576

PR #497, opened 2023-08-22
Remove ensureConfigMap as it is dead code

gibizer:
We missed this removal when we removed NovaExternalCompute


SeanMooney (approved):
makes sense

PR #499, opened 2023-08-22
Fix missing condition update

gibizer:
In multiple places if createOrPatch on a sub CR fails the error is not propagated to the higher level CR (Nova or NovaCell) as the call `condition.FalseCondition` only generates a condition locally but without its result passed to `instance.Status.Conditions.Set` the instance Status is not update.

These places are hard to test in envtest as it would require to break the running kube-apiserver to generate an error at that point. So they slip through.

Now this is fixed based on grepping for
  `grep -P '^\t*condition.FalseCondition' -R`

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -855,13 +855,12 @@ func (r *NovaReconciler) ensureAPI(
 	})
 
 	if err != nil {
-		condition.FalseCondition(
+		instance.Status.Conditions.Set(condition.FalseCondition(
 			novav1.NovaAPIReadyCondition,
 			condition.ErrorReason,
 			condition.SeverityError,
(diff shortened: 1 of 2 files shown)


gibizer:
/retest
```
 * copying system image from manifest list: parsing image configuration: Error fetching blob: invalid status code from registry 503 (Service Unavailable)
 ```

SeanMooney (approved):
is this something you could add to operator lint?

gibizer:
> is this something you could add to operator lint?

We can add a specific rule to it yes as we can detect statically if the result of FalseCondition is not used

gibizer:
Noted in  https://github.com/openstack-k8s-operators/nova-operator/pull/499  but I don't know when will I have time to get to it.

Issue #143, opened 2022-11-21
Wait for the Secret CellMessageBusSecretName to exists and include its hash to the InputHash


gibizer:
Current TransportURL Secret handling does not ensure the Secret exists, has a transport_url key, and we don't include it in the InputHash of our Service CRD so we don't detect if the content of the Secret changes.

Issue #79, opened 2022-10-03
Restructure secret handling


gibizer:
1) Pass Username via the secret
2) We might also want to pass the keystone auth URL, Database URL, rabbit URLs via Secret
3) Generate a separate secret for each lower level CR.

PR #513, opened 2023-08-31
Merge MessageBusSecret into the internal secret

gibizer:
This is a follow up of the work that created a scoped down internal
secrets for our sub CRs. Instead of passing down the internal secret and
the message bus secret separately this patch moves the content
(transport_url) from the message bus secret to the internal secret and
only passes down the internal secret. So we have one less secret per sub
CR to handle.

Note that this is a backwards incompatible CRD change as we deleted the
MessageBusSecret fields from our sub CRDs.

Closes: https://issues.redhat.com/browse/OSPRH-223
Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/143 (makes it unnecessary)
Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/79

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -310,7 +310,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// Create TransportURLs to access the message buses of each cell. Cell0
 	// message bus is always the same as the top level API message bus so
 	// we create API MQ separately first
-	apiMQSecretName, apiMQStatus, apiMQError := r.ensureMQ(
+	apiTransportURL, apiMQStatus, apiMQError := r.ensureMQ(
 		ctx, h, instance, instance.Name+"-api-transport", instance.Spec.APIMessageBusInstance)
 	switch apiMQStatus {
 	case nova.MQFailed:
@@ -339,18 +339,18 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	var failedMQs []string
 	var creatingMQs []string
 	for _, cellName := range orderedCellNames {
-		var cellMQ string
+		var cellTransportURL string
 		var status nova.MessageBusStatus
 		var err error
 		cellTemplate := instance.Spec.CellTemplates[cellName]
 		// cell0 does not need its own cell message bus it uses the
 		// API message bus instead
 		if cellName == novav1.Cell0Name {
-			cellMQ = apiMQSecretName
+			cellTransportURL = apiTransportURL
 			status = apiMQStatus
 			err = apiMQError
 		} else {
-			cellMQ, status, err = r.ensureMQ(
+			cellTransportURL, status, err = r.ensureMQ(
 				ctx, h, instance, instance.Name+"-"+cellName+"-transport", cellTemplate.CellMessageBusInstance)
 		}
 		switch status {
@@ -362,7 +362,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		default:
 			return ctrl.Result{}, fmt.Errorf("Invalid MessageBusStatus from ensureMQ: %d for cell %s", status, cellName)
 		}
-		cellMQs[cellName] = &nova.MessageBus{SecretName: cellMQ, Status: status}
+		cellMQs[cellName] = &nova.MessageBus{TransportURL: cellTransportURL, Status: status}
 	}
 	if len(failedMQs) > 0 {
 		instance.Status.Conditions.Set(condition.FalseCondition(
@@ -429,7 +429,8 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 
 		cell, status, err := r.ensureCell(
 			ctx, h, instance, cellName, cellTemplate,
-			cellDB.Database, apiDB, cellMQ.SecretName, keystoneAuthURL, secret,
+			cellDB.Database, apiDB, cellMQ.TransportURL,
+			keystoneAuthURL, secret,
 		)
 		cells[cellName] = cell
 		switch status {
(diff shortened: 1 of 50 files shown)


gibizer:
/test functional
lets try again to see if it fails now or not

gibizer:
I cannot reproduce the functional failure locally or here in the CI if run with more logging (need serial mode for that). So lets see how frequent the issue in a normal CI run

gibizer:
nova-api got killed due to liveness probe during tempest

controllers/novaapi_controller.go
```diff
@@ -327,19 +328,6 @@ func (r *NovaAPIReconciler) generateConfigs(
...
-		Namespace: instance.Namespace,
-		Name:      instance.Spec.APIMessageBusSecretName,
-	}
-	err := h.GetClient().Get(ctx, secretName, apiMessageBusSecret)
-	if err != nil {
-		util.LogForObject(
-			h, "Failed reading Secret", instance,
-			"APIMessageBusSecretName", instance.Spec.APIMessageBusSecretName)
-		return err
-	}
```
SeanMooney, line 341: this is now doen in the nova contoler for us when creating the new secret so ya thats nice to remove or rather move eairlier.

controllers/novacell_controller.go
```diff
@@ -123,23 +123,7 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
-
-	_, result, messageBusSecret, err := ensureSecret(
-		ctx,
-		types.NamespacedName{
-			Namespace: instance.Namespace,
-			Name:      instance.Spec.CellMessageBusSecretName,
-		},
-		[]string{
-			"transport_url",
+			TransportURLSelector,
```
SeanMooney, line 126: ack this is just transport_url but pulled out into a constnat so we dont have to keep it in sync everywhere +1

controllers/novametadata_controller.go
```diff
@@ -288,20 +289,6 @@ func (r *NovaMetadataReconciler) generateConfigs(
...
-		Name:      instance.Spec.APIMessageBusSecretName,
-	}
-	err := h.GetClient().Get(ctx, secretName, apiMessageBusSecret)
-	if err != nil {
-		util.LogForObject(
-			h, "Failed reading Secret", instance,
-			"APIMessageBusSecretName", instance.Spec.APIMessageBusSecretName)
-		return err
-	}
-
```
SeanMooney, line 304: yep again this is not assserted when the internal secret is created so no need to check for it in the templating and its also validated in the 
the call to ensureSecret

test/functional/base_test.go
```diff
@@ -347,16 +319,18 @@ func NovaSchedulerNotExists(name types.NamespacedName) {
 }
 
 type CellNames struct {
```
SeanMooney, line 321: this is kind of not a greate name any more but its ok.

its more a set of names of resouce needed by the cell 

so CellDepNames woudl be better but this is a minor nit
and unrealated to your patches main content so jsut noting this partly to myself.

test/functional/nova_controller_test.go
```diff
@@ -224,48 +216,48 @@ var _ = Describe("Nova controller", func() {
...
-			Expect(internalCellSecret.Data).To(HaveLen(3))
+			Expect(internalCellSecret.Data).To(HaveLen(4))
 			Expect(internalCellSecret.Data).To(
 				HaveKeyWithValue(controllers.APIDatabasePasswordSelector, []byte("api-database-password")))
 			Expect(internalCellSecret.Data).To(
 				HaveKeyWithValue(controllers.CellDatabasePasswordSelector, []byte("cell0-database-password")))
 			Expect(internalCellSecret.Data).To(
 				HaveKeyWithValue(controllers.ServicePasswordSelector, []byte("service-password")))
+			Expect(internalCellSecret.Data).To(
+				HaveKeyWithValue("transport_url", []byte("rabbit://cell0/fake")))
```
SeanMooney, line 239: +1

test/functional/nova_controller_test.go
```diff
@@ -342,21 +334,22 @@ var _ = Describe("Nova controller", func() {
 				HaveKeyWithValue(controllers.ServicePasswordSelector, []byte("service-password")))
 			Expect(internalTopLevelSecret.Data).To(
 				HaveKeyWithValue(controllers.MetadataSecretSelector, []byte("metadata-secret")))
+			Expect(internalTopLevelSecret.Data).To(
+				HaveKeyWithValue("transport_url", []byte("rabbit://cell0/fake")))
```
SeanMooney, line 338: ah yep its a byte array, this anser my previosu question on the string converation but its what i was expecting anyway.
cool.

test/kuttl/test-suites/default/scale-tests/03-assert.yaml
```diff
@@ -31,7 +31,6 @@ metadata:
 spec:
   apiDatabaseHostname: openstack
   apiDatabaseUser: nova_api
-  apiMessageBusSecretName: rabbitmq-transport-url-nova-kuttl-api-transport
```
SeanMooney, line 34: this made me realise there is no way to do a negitive assert via the yaml method with kuttl
that shoudl not really be an issue jsut something i realised looking at this.

bogdando: a negative assert may be made with kubeassert plugin, for example a check to verify that there is no namespaces stuck in deletion:
```
kubectl assert not-exist-enhanced namespace --field-selector metadata.deletionTimestamp!='<none>',spec.finalizers[*]!='<none>'
```

SeanMooney: sure but your missing the point i was making
i explicitly was stating we cant do that using yaml asserts.

in this case we don't need to make an assert I was noting that this is a limitation of Kuttl when not using shell scripts.

controllers/nova_controller.go
```diff
@@ -1162,7 +1160,23 @@ func (r *NovaReconciler) ensureMQ(
...
+		return "", nova.MQFailed, err
+	}
+
+	url, ok := secret.Data[TransportURLSelector]
+	if !ok {
+		return "", nova.MQFailed, fmt.Errorf(
+			"the TransportURL secret %s does not have 'transport_url' field", transportURL.Status.SecretName)
+	}
+
+	return string(url), nova.MQCompleted, nil
```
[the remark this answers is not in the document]
gibizer: yes this is the []byte -> string conversion. (to be fair golang does this pretty nicely without string encoding complications)

test/kuttl/test-suites/default/scale-tests/03-assert.yaml
gibizer: We can do negative asserts with 03-errors.yaml https://kuttl.dev/docs/testing/asserts-errors.html#format

SeanMooney: oh your right we can cool another use where bash is not required :)

PR #515, opened 2023-09-01
[metadata]Generate metadata external secret for neutron

gibizer:
The neutron metadata agent running on EDPM compute nodes needs to connect to the nova metadata service running in k8s. To be able to do that the neutron metadata agent needs to know the secret shared between neutron and nova, and the nova metadata service hostname.

The nova metadata service can be deployed either centrally or per cell. So the above information depends on the nova metadata deployment mode.

To ease the neutron configuration on the EDPM node, nova-operator now exposes the above information in Secret objects per NovaMetadata CR. If a single nova metadata service is deployed on the top level then the name of the secret is `<Nova/name>-metadata-neutron-config` (i.e. nova-metadata-neutron-config in the default deployment). If the nova metadata service is deployed per cell then the name of the Secrets are `<Nova/name>-<cell name>-metadata-neutron-config` (i.e in cell1 in the default deployment nova-cell1-metadata-neutron-config).

These secretes contains a single oslo config snippet that can be added to the neutron conf.d directory to extend the existing neutron metadata configuration with the nova specific parameters. The single key in the secret is named 05-nova-metadata.conf

* [x] Test coverage
  * [x] Metadata shared secret changes reflected in the generated config
* [x] Documentation

### DIFF
--- a/controllers/novametadata_controller.go
+++ b/controllers/novametadata_controller.go
@@ -37,6 +37,7 @@ import (
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	nad "github.com/openstack-k8s-operators/lib-common/modules/common/networkattachment"
+	common_secret "github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/statefulset"
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
@@ -195,11 +196,26 @@ func (r *NovaMetadataReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return result, err
 	}
 
-	result, err = r.ensureServiceExposed(ctx, h, instance)
+	apiEndpoints, result, err := r.ensureServiceExposed(ctx, h, instance)
 	if (err != nil || result != ctrl.Result{}) {
 		// We can ignore RequeueAfter as we are watching the Service and Route resource
 		// but we have to return while waiting for the service to be exposed
 		return ctrl.Result{}, err
+	}
+
+	// We have to wait until our service is fully exposed so that we can
+	// generate the compute config containing the metadata host
+	if !instance.Status.Conditions.IsTrue(condition.ExposeServiceReadyCondition) {
+		util.LogForObject(
+			h, "Waiting for the service to be exposed before generating compute configuration", instance)
+		return ctrl.Result{}, nil
+	}
+
+	// TODO(gibi): fix lib-common endpoint.ExposeEndpoints return value to
+	// avoid the need for the cast
+	err = r.ensureNeutronConfig(ctx, h, instance, apiEndpoints[string(endpoint.EndpointInternal)], secret)
+	if err != nil {
+		return result, err
 	}
 
 	util.LogForObject(h, "Successfully reconciled", instance)
@@ -257,6 +273,11 @@ func (r *NovaMetadataReconciler) initConditions(
 				condition.NetworkAttachmentsReadyCondition,
 				condition.InitReason,
 				condition.NetworkAttachmentsReadyInitMessage,
+			),
+			condition.UnknownCondition(
+				novav1.NovaComputeServiceConfigReady,
+				condition.InitReason,
+				novav1.NovaComputeServiceConfigInitMessage,
 			),
 		)
 
@@ -416,7 +437,7 @@ func (r *NovaMetadataReconciler) ensureServiceExposed(
 	ctx context.Context,
 	h *helper.Helper,
 	instance *novav1.NovaMetadata,
-) (ctrl.Result, error) {
+) (map[string]string, ctrl.Result, error) {
 	var ports = map[endpoint.Endpoint]endpoint.Data{
 		endpoint.EndpointInternal: {Port: novametadata.APIServicePort},
 	}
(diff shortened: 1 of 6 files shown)


gibizer:
I tried this locally in crc and it generated the config secret:
```
$ oc get secret nova-metadata-compute-config -o yaml | yq '.data."05-nova-metadata.conf"' -r | base64 --decode
[DEFAULT]
nova_metadata_host = http://nova-metadata-internal.openstack.svc:8775
metadata_proxy_shared_secret = 1234567842
```

gibizer:
@SeanMooney how do you feel about teaching nova-operator how to (partially) configure the neutron-metadata-agent? See my thinking process in https://github.com/openstack-k8s-operators/neutron-operator/pull/203#discussion_r1312701074

mrKisaoLamb (commented):
This looks good and IMO ready to merge when we reach agreement with OVN team

controllers/novametadata_controller.go
```diff
@@ -508,6 +525,73 @@ func getMetadataServiceLabels(cell string) map[string]string {
...
+	)
+
+	return nil
+}
+
+func (r *NovaMetadataReconciler) generateComputeConfigs(
+	ctx context.Context, h *helper.Helper,
+	instance *novav1.NovaMetadata, endpoint string, secret corev1.Secret,
+) error {
+	configName := instance.GetName() + "-compute-config"
```
booxter, line 557: we may have to deploy metadata agent (and nova metadata?) on nodes that don't run libvirt (Networker nodes). Considering this, I wonder if the `compute` suffix is always fitting. Maybe `external-config`? (We are going to use this suffix for neutron secrets.)

SeanMooney: i do not really want to see edpm reference in this operator and i kind of object to it in any of our code so we could a "dataplane" suffix i guess.

a better suffix would be "cell-config" and "global-config" for when its deployed at the top level.

I'm border line -2 on this approach as it is for what its worth.

we could generated this configuration but I'm not entirely sold that we should.

really the neutorn operator should be generating this and confining ovn
the problem is neutron and ovn are not currently cell aware and that partly for valid reasons.

that leaves the nova operator generating config for use by a OpenstackDataplaneService which will configure nutron/ovn which is not ideal but it better then some of the other options.

gibizer: > we may have to deploy metadata agent (and nova metadata?) on nodes that don't run libvirt (Networker nodes). Considering this, I wonder if the `compute` suffix is always fitting. Maybe `external-config`? (We are going to use this suffix for neutron secrets.)

I don't think we want to deploy nova-metadata outside of k8s. Is there a use case where we have to do that?

I also assumed that neutron-metadata-agent needs to run on the same node where the VM runs hence the -compute suffix. But if this is not true then sure we can find a better name for this Secret.

> a better suffix would be "cell-config" and "global-config" for when its deployed at the top level.

I lean towards `nova-metadata-external-config` and `nova-cell1-metadata-external-config`

gibizer: We discussed this on the today's nova sync call and the compromise was `nova-metadata-neutron-config` and `nova-cell1-metadata-neutron-config`

gibizer: > I also assumed that neutron-metadata-agent needs to run on the same node where the VM runs [...] But if this is not true [...]

So if the neutron-metadata-agent does not need to be run on the same node as the VMs. And nova-metadata deployed per cell. Then how the end user will know which cell config to use for the neutron-metadata-agent? Will there be different network nodes for each cell? Or a single network node will serve multiple cells and for each cell it will run a separate neutron-metadata-agent?

gibizer:
/test functional

It seems like the runner VM run out of memory / cpu as kube-api stopped to respond

controllers/novametadata_controller.go
booxter: Not that this necessarily changes the calculations here, but for context:

> I also assumed that neutron-metadata-agent needs to run on the same node where the VM runs hence the -compute suffix
In case of SR-IOV ports, the metadata port is explicitly bound to a "different" chassis. But on the other hand...

> if the neutron-metadata-agent does not need to be run on the same node as the VMs
For regular ports, metadata service is local (it's backed by a `localport` OVN port type - meaning no traffic leaves the node.)

booxter: As to your questions about how metadata-agent will be deployed for different cells on "Networker" nodes, this a great question and I don't have an immediate answer. I think it's a good topic to consider in the ongoing reference architecture discussions.

gibizer: Thanks for the context. Keep me in the loop on the reference architecture discussion.

templates/neutron-metadata.conf
```diff
@@ -0,0 +1,3 @@
+[DEFAULT]
+nova_metadata_host = {{.nova_metadata_host}}
+metadata_proxy_shared_secret = {{.metadata_proxy_shared_secret}}
```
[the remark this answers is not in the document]
gibizer: @mrKisaoLamb 
> If yes, maybe we should put somewhere a note.
At least this part of is easy. I added a code comment to the code filling the template now.

SeanMooney (approved):
i dont have a better solution right now beyond generging the config in ansible form data passed in the secret

we can always change to that form where if we want to so for now we should keep this simple and proceed with this patch.

thanks for updating the design doc to capture the reasoning for this approach and how it will be consumed.

SeanMooney:
correct the better solution  is we are going to discuss adding a new API to nova to remove the need for the config options entirely but that is a non backporable change and therefor only relevant to future OpenStack releases.

doc/design.md
```diff
@@ -55,3 +55,51 @@ oslo.config will apply the different config snippets. The nova-operator will
...
+
+1. If a single nova-metadata service is deployed on the top level (i.e. when
+`Nova.Spec.MetadataServiceTemplate.Enabled` is true) then the name of the
+Secret is `<Nova/name>-metadata-compute-config` (i.e.
+`nova-metadata-compute-config`).
+
+2. If the nova-metadata service is deployed per cell
+(i.e. `Nova.Spec.CellTemplates[].MetadataServiceTemplate.Enabled` is true) then
+the name of the Secrets are `<Nova/name>-<cell name>-metadata-compute-config`
+(i.e in cell1 `nova-cell1-metadata-compute-config`)
```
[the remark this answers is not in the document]
gibizer: good point. I will push a follow up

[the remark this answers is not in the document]
gibizer: https://github.com/openstack-k8s-operators/nova-operator/pull/526

doc/design.md
```diff
@@ -55,3 +55,51 @@ oslo.config will apply the different config snippets. The nova-operator will
...
+necessary neutron-metadata-agent config snippet that defines how the
+nova-metadata service can be accessed by the neutron-metadata-agent.
+
+There are two possible nova-metadata deployment modes and this also means that
+there are two possible configuration schemes for the neutron-metadata-agent.
+
+1. If a single nova-metadata service is deployed on the top level (i.e. when
+`Nova.Spec.MetadataServiceTemplate.Enabled` is true) then the name of the
+Secret is `<Nova/name>-metadata-compute-config` (i.e.
+`nova-metadata-compute-config`).
```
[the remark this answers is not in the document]
gibizer: https://github.com/openstack-k8s-operators/nova-operator/pull/526

templates/neutron-metadata.conf
```diff
@@ -0,0 +1,3 @@
+[DEFAULT]
+nova_metadata_host = {{.nova_metadata_host}}
```
karelyatin, line 2: with this neutron metadata agent fails to start, reported https://github.com/openstack-k8s-operators/nova-operator/issues/532

PR #520, opened 2023-09-05

dprince:
This fixes a warning when building bundles with USE_IMAGE_DIGESTS=true by removing the related image constant for METADATA which is exactly the same as API. This does not change internal operator code which could support a separate METADATA image at some point in the future


gibizer:
> This fixes a warning when building bundles with USE_IMAGE_DIGESTS=true by removing the related image constant for METADATA which is exactly the same as API. This does not change internal operator code which could support a separate METADATA image at some point in the future

OK this only affects our downstream build where we use the defaults from the env to pin down the images to a certain hash. In that process there is no way now to pin different images for nova-api and nova-metadata. This is probably OK as I don't really see why we would need a different image at the end.

gibizer:
The tempest issue is known and unrelated (I'm trying to find a the reason separately why nova-api requests during tempest can hang)

gibizer (approved):

PR #518, opened 2023-09-04
Use heartbeat_in_pthread=false for nova-api 

gibizer:
With heartbeat_in_pthread=true we saw that sometimes nova-api hangs. If
the amqp logging was also enabled then the hang could be reproduced with
any request that needed rabbitmq connection.

It seems that the root cause is that the native thread running the
heartbeat tries to log but the oslo.log stack was initialized in the
main thread where locking was monkey patched by eventlet. The oslo.log
wants to take a lock but taking an eventlet lock from another native
thread is not supported by eventlet and the heartbeat thread hangs at:

```
/usr/lib/python3.9/site-packages/oslo_messaging/_drivers/impl_rabbit.py:782 in __init__
    `LOG.debug('[%(connection_id)s] Connected to AMQP server on '`

/usr/lib64/python3.9/logging/__init__.py:1800 in debug
    `self.log(DEBUG, msg, *args, **kwargs)`

/usr/lib64/python3.9/logging/__init__.py:1844 in log
    `self.logger.log(level, msg, *args, **kwargs)`

/usr/lib64/python3.9/logging/__init__.py:1512 in log
    `self._log(level, msg, args, **kwargs)`

/usr/lib64/python3.9/logging/__init__.py:1589 in _log
    `self.handle(record)`

/usr/lib64/python3.9/logging/__init__.py:1599 in handle
    `self.callHandlers(record)`

/usr/lib64/python3.9/logging/__init__.py:1661 in callHandlers
    `hdlr.handle(record)`

/usr/lib64/python3.9/logging/__init__.py:950 in handle
    `self.acquire()`

/usr/lib64/python3.9/logging/__init__.py:901 in acquire
    `self.lock.acquire()`

/usr/lib64/python3.9/threading.py:160 in acquire
    `rc = self._block.acquire(blocking, timeout)`

/usr/lib/python3.9/site-packages/eventlet/semaphore.py:115 in acquire
    `hubs.get_hub().switch()`

/usr/lib/python3.9/site-packages/eventlet/hubs/hub.py:313 in switch
    `return self.greenlet.switch()`
```

while the main thread prints:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/eventlet/hubs/hub.py", line 476, in fire_timers
    timer()
  File "/usr/lib/python3.9/site-packages/eventlet/hubs/timer.py", line 59, in __call__
    cb(*args, **kw)
[...]

PR #521, opened 2023-09-06
Run two nova-api and -metadata processes per pods

gibizer:
We want our services to be scaled by pod replicas but we still need two
processes per API pods to always have a room for a healthcheck API
query.

### DIFF
--- a/templates/novaapi/config/httpd.conf
+++ b/templates/novaapi/config/httpd.conf
@@ -38,7 +38,9 @@ LogLevel info
   WSGIProcessGroup nova-api
   WSGIApplicationGroup %{GLOBAL}
   WSGIPassAuthorization On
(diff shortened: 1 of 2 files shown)


gibizer:
3rd try

gibizer:
So far with this change I cannot reproduce the tempest server create request hang.

SeanMooney:
i mean this seams to be working fairly well do you want to try this a few more times or shall we proceeded.

gibizer:
> i mean this seams to be working fairly well do you want to try this a few more times or shall we proceeded.

Yeah this seems to be not affected by the hang. I have some interesting findings in https://github.com/openstack-k8s-operators/nova-operator/pull/518 It seems that nova-api hangs at the rabbitmq hearthbeat

SeanMooney (approved):
lets proceed with this and we can experiment with the pthread option too if we want too also via the other pr.

PR #533, opened 2023-09-18
[envtest]Use the common UpdateSecret helper

gibizer:
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/337 (merged)


SeanMooney (approved):
ack this just uses the new test helper instead of our in tree one

Issue #532, opened 2023-09-18
nova_metadata_host is exposed in wrong format


karelyatin:
https://github.com/openstack-k8s-operators/nova-operator/pull/515 exposes nova_metadata_host as below:-

$ cat 05-nova-metadata.conf 
[DEFAULT]
nova_metadata_host = http://nova-metadata-internal.openstack.svc:8775
metadata_proxy_shared_secret = 1234567842

And with this neutron-ovn-metadata-agent fails to start with error:-
2023-09-18 08:00:04.449 2 CRITICAL neutron [-] Unhandled error: oslo_config.cfg.ConfigFileValueError: Value for option nova_metadata_host from LocationInfo(location=<Locations.user: (4, True)>, detail='/etc/neutron.conf.d/05-nova-metadata.conf') is not valid: http://nova-metadata-internal.openstack.svc:8775 is not a valid host address
2023-09-18 08:00:04.449 2 ERROR neutron Traceback (most recent call last):
2023-09-18 08:00:04.449 2 ERROR neutron   File "/usr/lib/python3.9/site-packages/oslo_config/types.py", line 843, in __call__
2023-09-18 08:00:04.449 2 ERROR neutron     value = self.ip_address(value)
2023-09-18 08:00:04.449 2 ERROR neutron   File "/usr/lib/python3.9/site-packages/oslo_config/types.py", line 728, in __call__
2023-09-18 08:00:04.449 2 ERROR neutron     self.version_checker(value)
2023-09-18 08:00:04.449 2 ERROR neutron   File "/usr/lib/python3.9/site-packages/oslo_config/types.py", line 748, in _check_both_versions
2023-09-18 08:00:04.449 2 ERROR neutron     raise ValueError("%s is not IPv4 or IPv6 address" % address)
2023-09-18 08:00:04.449 2 ERROR neutron ValueError: http://nova-metadata-internal.openstack.svc:8775 is not IPv4 or IPv6 address
2023-09-18 08:00:04.449 2 ERROR neutron
2023-09-18 08:00:04.449 2 ERROR neutron During handling of the above exception, another exception occurred:
2023-09-18 08:00:04.449 2 ERROR neutron
2023-09-18 08:00:04.449 2 ERROR neutron Traceback (most recent call last):
2023-09-18 08:00:04.449 2 ERROR neutron   File "/usr/lib/python3.9/site-packages/oslo_config/types.py", line 846, in __call__
[...]

PR #534, opened 2023-09-18
Fix neutron config format

gibizer:
The nova_metadata_host was set to a full URL but that only support hostname. So the config generation changed to split the URL and fill nova_metadata_host, nova_metadata_port, and nova_metadata_protocol accordingly.

Closes: #532


mrKisaoLamb (approved):
lgtm

karelyatin (commented):
/lgtm

Issue #217, opened 2023-01-10
The [keystone_authtoken] www_authenticate_uri option should point public url


kajinamit:
Currently the ``[keystone_authtoken] www_authenticate_uri`` option points internal url, but this should point public url, because the url is returned to clients when auth request fails.

We probably have to add separate options for public/internal endpoint and use different endpoints accordingly.

gibizer:
I believe nova-api using the public keystone URL for everything:
```console
$ oc get -o yaml NovaAPI/nova-api | grep keystoneAuthURL
  keystoneAuthURL: http://keystone-public-openstack.apps-crc.testing
```

I think the template parameter name is misleading here:
https://github.com/openstack-k8s-operators/nova-operator/blob/cb238f87b4d203340c55f5b63cf862e4b15aadd1/templates/nova.conf#L196

But the actual value is coming from here:
https://github.com/openstack-k8s-operators/nova-operator/blob/cb238f87b4d203340c55f5b63cf862e4b15aadd1/controllers/nova_controller.go#L810-L814

So my questions is: Is it OK to use the public endpoint for everything? 
* If yes, then I will rename the template parameter
* If no, then I need to have a list of config params names where we need the public and another list of config param names where we need the internal keystone auth url? Then I will introduce a separate field on NovaAPI CR and also in the template.

SeanMooney:
my intenstion is for nova to use the internal endpoitns of all service it calls but our external links in the API repoce should really be to the endpoint you invoked.

so if you call to the internal endpoint it should be internal and external external.

gibizer:
> my intenstion is for nova to use the internal endpoitns of all service it calls but our external links in the API repoce should really be to the endpoint you invoked.
> 
> so if you call to the internal endpoint it should be internal and external external.

1. So I guess to force nova to call the internal endpoint of keystone when calling to other services we need to set the  `auth_url ` to the internal endpoint in each [<service>] section. So that mean I will introduce a new field to the NovaAPI CR for the internal keystone endpoint.

2. > so if you call to the internal endpoint it should be internal and external external.
Do you know how to configure nova to get this behavior?

SeanMooney:
i dont think we should have to change the CR we should just take one keystone URL and that should be the internal one although we can always use endpoint to lookup the internal one too programmatically regardless of what is passed by the user in the cr.

for 2 no i thought we looked at the URL used to make the query to determine how we render the refurls in the responce.

kajinamit:
[...]

PR #535, opened 2023-09-18
[NovaAPI]Fix www_authenticate_uri configuration

gibizer:
According to the keystone middleware doc we have to set www_authenticate_uri to the public keystone endpoint as an unauthenticated user is redirected to this URL.

Our service uses the internal endpoint for auth and so far it used the internal endpoint for www_authenticate_uri as well. But this will not work for external users.

A new field is added to the NovaAPI CR to allow configuring not just the internal endpoint but also the public endpoint.

Closes: #217


gibizer:
/test nova-operator-build-deploy
```
{  failed to wait for the created cluster claim to become ready: timed out waiting for the condition}
```

api/bases/nova.openstack.org_novaapis.yaml
```diff
@@ -155,6 +155,13 @@ spec:
                   type: object
                 type: array
               keystoneAuthURL:
+                description: KeystoneAuthURL configures the keystone API endpoint
+                  to be used by the service for authentication and authorization
+                type: string
+              keystonePublicAuthURL:
```
SeanMooney, line 161: dumb question but can we look this up form the keystone catalog?

I'm not nessisarly against adding this just wondering if it is needed

gibizer: We can. We can look for KeystoneEndpoint CRs (how the Nova controller does it to fill this value), or we can use the golang sdk and call keytone internal api to get the catalog and get the public URL.  I went with the former in the nova controller as we already looking for KeystoneEndpoints there.

gibizer:
We know that tempest will only be green again after https://github.com/openstack-k8s-operators/openstack-operator/pull/457 is landed.

SeanMooney (approved):

PR #537, opened 2023-09-20
Remove external-compute samples

booxter:
UPD:

    They are obsolete and not used anywhere. If we ever need something
    similar for CI purposes, we will have to effectively start from
    scratch anyway.

---

ORIGINAL DESCRIPTION:

Setting it is no longer needed since ovn role will configure sb connection strings in metadata agent and ovn-controller via ConfigMaps / Secrets prepared by neutron and ovn operators.

NOTE: I don't know if this sample is current at all. If this is not helpful, let me know and I will abandon the PR. Thanks.


gibizer:
> Setting it is no longer needed since ovn role will configure sb connection strings in metadata agent and ovn-controller via ConfigMaps / Secrets prepared by neutron and ovn operators.
> 
> NOTE: I don't know if this sample is current at all. If this is not helpful, let me know and I will abandon the PR. Thanks.

I think the whole external-compute directory can be dropped now as they are heavily outdated and is no test fails they are unused.

SeanMooney:
ya we can drop them
i woudl like to reintrodce a similar set of examples as part of ci jobs but we don't need the current ones

booxter:
Ack. I killed the whole subdirectory as you suggested.

booxter:
```
oc project cert-manager-operator
error: unable to default to a user name: the server is currently unable to handle the request (get users.user.openshift.io ~)

```

booxter:
recheck

SeanMooney (approved):

PR #539, opened 2023-09-22
[envtest]Use keystone helpers from keystone-operator

gibizer:
This was necessary to remove a dependency cycle from lib-common


gibizer:
/unhold
dependencies landed

gibizer:
you can test that it does not depend on the lib-common keystone helpers any more by adding the following replace and run the test locally. The lib-common hash there is the one that removes the keystone helpers from lib-common (https://github.com/openstack-k8s-operators/lib-common/pull/339)
```
replace github.com/openstack-k8s-operators/lib-common/modules/test => github.com/gibizer/lib-common/modules/test v0.0.0-20230925122035-657099e5679d
```

stuggi (approved):
/lgtm

PR #544, opened 2023-09-26
Enable multipath for volume attachments

ASBishop:
Configure nova-compute so that it always requests (via os-brick) volume attachments provide multiple paths whenever possible. Multiple paths are requested, but not required. An attachment request can succeed, even when only a single path can be supplied.


gibizer:
recheck
I feel like it was slow node

gibizer (approved):
All green

PR #549, opened 2023-09-29
Fix VNC config generation

gibizer:
863615c0c5b899f73f0f98049a14c9c2203d9e0e reworked the VNC config template but made a mistake that causes that the vncproxy service gets two [vnc] section generated and non vnc related services gets a [vnc]enabled=false generated.

This is fixed now.


mrKisaoLamb (approved):
lgtm

gibizer:
recheck

PR #559, opened 2023-10-04
[nova-compute-discovery] Add host discovery on nova lvl

mrKisaoLamb:


controllers/nova_controller.go
```diff
@@ -803,14 +811,136 @@ func (r *NovaReconciler) ensureCell(
 	status, err := r.ensureCellMapped(
 		ctx, h, instance,
 		cell, cellTemplate, apiDB.GetDatabaseHostname(), cellTransportURL)
+
+	if status == nova.CellMappingReady {
+		// As we not always need discover we can set already ready
+		status = nova.CellReady
+	}
```
gibizer, line 818: this is now a duplicate of L828 - L830 now. Do we need both?

controllers/nova_controller.go
```diff
@@ -803,14 +811,136 @@ func (r *NovaReconciler) ensureCell(
...
+
+	if len(cellTemplate.NovaComputeTemplates) != 0 && cell.Status.Conditions.IsTrue(novav1.NovaAllComputesReadyCondition) {
```
gibizer, line 820: if you move this if into the `ensureNovaComputeDiscover` call then you can use the same pattern as we use for ensureCellMapped. So the return value of the `ensureNovaComputeDiscover` can be `nova.CellComputeDiscoveryReady` if there is nothing else needed (either as the job was done, or there was no need for a job) and then you can translate nova.CellComputeDiscoveryReady to  nova.CellReady here below

controllers/nova_controller.go
```diff
@@ -803,14 +811,136 @@ func (r *NovaReconciler) ensureCell(
...
+			ConfigOptions:      templateParameters,
+			Labels:             cmLabels,
+			CustomData:         map[string]string{},
+			Annotations:        map[string]string{},
+			AdditionalTemplate: extraTemplates,
+		},
+	}
+
+	configHash := make(map[string]env.Setter)
+	err := secret.EnsureSecrets(ctx, h, cell, cms, &configHash)
```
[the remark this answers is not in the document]
mrKisaoLamb: there is also difference that cell mapping is also running for cell0 so config looks differently

gibizer:
recheck
kuttl is fixed now

controllers/nova_controller.go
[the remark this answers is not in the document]
gibizer: True, the cell0 case is a specific thing for cell mapping. However discover hosts will never run on cell0 as cell0 cannot have computes, therefore this cell0 case is not important for the discover hosts job perspective. I believe we can directly reuse the `"%s-config-data", cell.Name+"-manage"` Secret in `ensureNovaComputeDiscover`.  I would probably pull out the generation of those Secrets from `ensureCellMapped` to before it. And then pass the name of the Secret to both ensureCellMapped and ensureNovaComputeDiscover calls.

test/functional/nova_multicell_test.go
gibizer, line 1: could you please add coverage for the case when a new compute is added (or a replica is changed) and therefore the hash is changing and the Job is rerun?

controllers/nova_controller.go
```diff
@@ -802,14 +810,145 @@ func (r *NovaReconciler) ensureCell(
 	status, err := r.ensureCellMapped(
 		ctx, h, instance,
 		cell, cellTemplate, apiDB.GetDatabaseHostname(), cellTransportURL)
+
 	if status == nova.CellMappingReady {
-		// As mapping is the last step if that is ready then the cell is ready
+		// As we not always need discover we can set already ready
 		status = nova.CellReady
 	}
```
gibizer, line 817: you can drop this now the L828 will take care of this now

gibizer:
recheck
Some DHCP agent related tempest case failed, nothing related to the PR itself

api/v1beta1/nova_types.go
```diff
@@ -133,6 +133,10 @@ type NovaStatus struct {
 	// nova_api database with a value that is the hash of the given cell
 	// configuration.
 	RegisteredCells map[string]string `json:"registeredCells,omitempty"`
+
+	// DiscoveredCells is a map keyed by cell names that have discovered all computes in cell
```
gibizer, line 137: Maybe we can say "all kubernetes managed computes" to avoid confusion with EDPM computes.

controllers/nova_controller.go
```diff
@@ -634,6 +642,94 @@ func (r *NovaReconciler) initConditions(
 	return nil
 }
 
+func (r *NovaReconciler) ensureNovaJobsSecret(
```
gibizer, line 645: nit: I would call it ensureNovaManageJobSecret

controllers/nova_controller.go
```diff
@@ -634,6 +642,94 @@ func (r *NovaReconciler) initConditions(
...
+	ctx context.Context, h *helper.Helper, instance *novav1.Nova,
+	cell *novav1.NovaCell,
+	cellTemplate novav1.NovaCellTemplate,
+	apiDBHostname string,
+	cellTransportURL string,
+) (map[string]env.Setter, string, string, error) {
+	scriptName := ""
+	configName := ""
+	ospSecret, _, err := secret.GetSecret(ctx, h, instance.Spec.Secret, instance.Namespace)
```
gibizer, line 654: We can pass the Secret object in from Reconcile as it is loaded already at https://github.com/openstack-k8s-operators/nova-operator/blob/d371f09abbc69b32b9cda473b724c1fa01f2f095/controllers/nova_controller.go#L200

pkg/nova/host_discovery.go
```diff
@@ -0,0 +1,93 @@
...
+	scriptName string,
+	inputHash string,
+	labels map[string]string,
+) *batchv1.Job {
+	runAsUser := int64(0)
+
+	args := []string{
+		"-c",
+		"/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/host_discover.sh",
+	}
```
gibizer, line 38: Please adapt to the recently merged change https://github.com/openstack-k8s-operators/nova-operator/pull/567 and use StopJob here as well

test/functional/nova_reconfiguration_test.go
```diff
@@ -435,6 +440,35 @@ var _ = Describe("Nova reconfiguration", func() {
...
+			discoverJob := th.GetJob(cell1.HostDiscoveryJobName)
+			oldJobInputHash := GetEnvVarValue(
+				discoverJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
+
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				zero := int32(0)
+				cell1 := nova.Spec.CellTemplates["cell1"]
+				ironicTemplate := cell1.NovaComputeTemplates[ironicComputeName]
+				ironicTemplate.Replicas = &zero
```
gibizer, line 454: `ptr.To(0)` is a nicer way to put it. 

Also either go up from 1 to 2 replicas, or note here that we go to 0 to simplify the test as the job is rerun in any replica changes even if it is a decrease.

gibizer (commented):
Thanks for pulling out the config generation. Overall it looks good. I can be convinced to merge this and fix my comments in a follow up.

gibizer:
recheck
Infra issue

gibizer:
recheck
Dhcp tempest test failes again. I need to check this o  main separately.

PR #567, opened 2023-10-12
Introduce Debug stopJob to replace stopDBSync                

bogdando:
Add Debug.StopJob into CR specs to control execution of
a job.

Added that, StopDBSync now becomes redundant as it controls db sync,
which is also a job. Having a generic StopJob looks better.
Remove StopDBSync and change the db sync job to use StopJob as well.

The 2nd user of the stopJob debug interface is the nova cellmapper job.

We can add more jobs in the future, which might be useful to debug as
well.

Also, remove implementation specific questions from the spec fileds
descriptions.

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -76,23 +76,22 @@ type NovaServiceBase struct {
 }
 
 // Debug allows enabling different debug option for the operator
+// QUESTION(gibi): Not all CR will run a service, dbsync, or cells mappings
+// jobs.  Should we have per CR Debug struct, or keep this generic one and
+// ignore fields in the controller that are not applicable?
 type Debug struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
-	// StopDBSync allows stopping the init container before running db sync
-	// to apply the DB schema
-	// QUESTION(gibi): Not all CR will run dbsync, should we have per CR
-	// Debug struct or keep this generic one and ignore fields in the
-	// controller that are not applicable
-	StopDBSync bool `json:"stopDBSync"`
+	// StopService allows stopping the service container before staring the
+	// openstack service binary. If the CR does not start any service, then the
+	// value of the field is ignored.
+	StopService bool `json:"stopService"`
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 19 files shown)


api/v1beta1/common_types.go
```diff
@@ -94,6 +94,11 @@ type Debug struct {
 	StopService bool `json:"stopService"`
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
+	// StopJob allows stopping the jobs containers before staring
+	// the openstack service binary
```
gibizer, line 98: nit: ...before starting the script of the job

api/v1beta1/common_types.go
```diff
@@ -94,6 +94,11 @@ type Debug struct {
...
+	StopJob bool `json:"stopJob"`
```
gibizer, line 99: Now we have StopDBSync and StopJob and db sync is also a job. Also looking at the comment at StopDBSync I think having a generic StopJob is better. So I suggest to remove StopDBSync and change the db sync job to use  StopJob as well.

gibizer (commented):
I have one comment on the Debug CRD field but otherwise this looks good.

api/v1beta1/common_types.go
```diff
@@ -76,25 +76,22 @@ type NovaServiceBase struct {
...
-	// StopDBSync allows stopping the init container before running db sync
-	// to apply the DB schema
-	// QUESTION(gibi): Not all CR will run dbsync, should we have per CR
-	// Debug struct or keep this generic one and ignore fields in the
-	// controller that are not applicable
-	StopDBSync bool `json:"stopDBSync"`
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
 	// StopService allows stopping the service container before staring
 	// the openstack service binary
```
gibizer, line 86: As the question is removed now, I would state the current behavior so add:
..starting the openstack service binary. If the CR does not start any service then the value of the field is ignored.

gibizer (commented):
only nits about the CR documentation. But the logic looks good. Let me know if you want to fix the nit separately in a follow up and I will approve this PR.

PR #570, opened 2023-10-17
Add nova-compute samples

mrKisaoLamb:
Add missing samples for nova-compute (both FakeDriver and IronicDriver)


gibizer (commented):
Could you please add these to the envtest sample tests?

mrKisaoLamb:
> Could you please add these to the envtest sample tests?

Good catch! Thanks, done

gibizer (approved):
Looks good. Thanks!

PR #574, opened 2023-10-17
[compute]Configure live migration

gibizer:
The default nova-compute config is changed 
* to contain a live_migration_uri config that is suitable for the current EDPM deployment model.
* to set skip_cpu_compare_on_dest = True workaround flag to avoid an upstream nova bug https://bugs.launchpad.net/nova/+bug/2039803

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -162,6 +162,11 @@ disable_fallback_pcpu_query=true
 {{ if eq .service_name "nova-compute"}}
 enable_qemu_monitor_announce_self=true
 reserve_disk_resource_for_image_cache=true
+
+# NOTE(gibi): We need this as live migration does not work with
+# cpu_mode=host-model . See https://bugs.launchpad.net/nova/+bug/2039803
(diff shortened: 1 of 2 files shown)


SeanMooney (approved):
we have an upstream nova bug to track this and this is an acceptable temporary workaround.

skip_cpu_compare_on_dest should not be used in production long term as it leads to very late live migration failures which can be expensive for large deployments but for now this is ok

gibizer:
recheck retry

SeanMooney (approved):
we still need to wait for the deconstierstion patch and another edpm-ansible patch but ill approve this for now and we can then just remove the hold once they have landed.

this looks fine to me as is so once the deps are ready we should be good.

gibizer:
/unhold 

This won't break anything. For the live migration to work we need https://github.com/openstack-k8s-operators/edpm-ansible/pull/467 to land and we need to bump everything to openstack-operator. After that I will enable the live-migration tempest tests

PR #578, opened 2023-10-20
Default [upgrade_levels]compute=auto for all services

gibizer:
This is important for upgrade and adoption support and safe to use in greenfield as well.

### DIFF
--- a/pkg/nova/cellmapping.go
+++ b/pkg/nova/cellmapping.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	cellMappingCommand      = "/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/ensure_cell_mapping.sh"
+	cellMappingCommand = "/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/ensure_cell_mapping.sh"
 )
 
 func CellMappingJob(
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -286,3 +286,6 @@ project_domain_name = {{ .default_project_domain }}
 user_domain_name = {{ .default_user_domain}}
 {{ end }}
 {{ end }}
+
+[upgrade_levels]
+compute = auto
--- a/test/functional/nova_compute_ironic_controller_test.go
+++ b/test/functional/nova_compute_ironic_controller_test.go
@@ -152,6 +152,8 @@ var _ = Describe("NovaCompute controller", func() {
 				configData := string(configDataMap.Data["01-nova.conf"])
 				Expect(configData).Should(
 					ContainSubstring("transport_url=rabbit://cell1/fake"))
+				Expect(configData).Should(
+					ContainSubstring("[upgrade_levels]\ncompute = auto"))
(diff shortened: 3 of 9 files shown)


pkg/nova/cellmapping.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	cellMappingCommand      = "/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/ensure_cell_mapping.sh"
+	cellMappingCommand = "/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/ensure_cell_mapping.sh"
```
gibizer, line 14: go fmt is a mystery

SeanMooney: :)

SeanMooney (approved):
thanks

SeanMooney (approved):

gibizer:
> @SeanMooney @gibi During adoption, this auto-upgrades Nova control plane services right after starting it antelope containers. I believe the pre-requisite for that is EDPM compute nodes also starting upgraded to Antelope, which means they need this config option out of box in edpm ansible. We can do that here firstly, but then need to enable `disable_compute_service_check_for_ffu=false` as well

disable_compute_service_check_for_ffu is something that should be false in an cloud except during upgrade. It is defaulted to false, so it is not set here. At the start of adoption it need to be set to true, then at the end of the adoption it need to set to false (or remove as the default is false).

SeanMooney:
> @SeanMooney @gibi During adoption, this auto-upgrades Nova control plane services right after starting it antelope containers. I believe the pre-requisite for that is EDPM compute nodes also starting upgraded to Antelope, which means they need this config option out of box in edpm ansible. We can do that here firstly, but then need to enable `disable_compute_service_check_for_ffu=false` as well

not quite.

auto means that the rpc clint will auto discover the new common rpc version supported by all comptue services.
provided we are not crossing a RPC major version bump this will find a common version.

disable_compute_service_check_for_ffu=false is the default
so we don't need to set it.
we should not be generating 
disable_compute_service_check_for_ffu=true by default.

disable_compute_service_check_for_ffu=true can be used when doing an unsupported upgrade such as adoption or FFU which exceed the supported upgrade range form an upstream online migration standpoint.

disable_compute_service_check_for_ffu=true disables the safety check for unsupproted/untested configs. even though in theory nova can manage any compute in the same major version of the rpc API we only test n > n+1 before antelop and n>n+2 after 

adoption is doing n>n+4 which is is not an upgrade path nova support even though it can work hence the need for disable_compute_service_check_for_ffu=true

gibizer:
recheck
```
ConflictException: 409: Client Error for url: https://compute.rdo.mtl2.vexxhost.net/v2.1/servers/c0b23e50-10e4-4a1b-b9c5-fcf525ffdbd9/os-interface, Port a931ca07-7327-4f18-b59f-78d55771acfa is still in use.
```

gibizer:
recheck
tempest was fixed

PR #584, opened 2023-10-27
[tempest]Disable swift tests

gibizer:
The object storage tests started to fail with Unauthorized errors. So this PR disables them. We might keep them disabled for the long run as nova-operator and nova does not depends on swift.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -205,6 +205,9 @@
         - test_server_connectivity_cold_migration
         - test_server_connectivity_live_migration
         - test_server_connectivity_resize
+      # Swift test failing with unauthorized errors
+        - tempest.api.object_storage
+        - tempest.scenario.test_object_storage
 - project:
     name: openstack-k8s-operators/nova-operator
     github-check:


gibizer:
Example failures https://github.com/openstack-k8s-operators/nova-operator/pull/583

gibizer:
So swift tries to use the public keystone endpoint but we recently enabled TLS there. 
```
2023-10-26T22:30:43.749023107+00:00 stderr F proxy-server: Unable to validate token: Could not find versioned identity endpoints when attempting to authenticate. Please check that your auth_url is correct. SSL exception connecting to https://keystone-public-openstack.apps-crc.testing/: HTTPSConnectionPool(host='keystone-public-openstack.apps-crc.testing', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)')))
```

gibizer:
expected fix for swift is https://github.com/openstack-k8s-operators/swift-operator/pull/63

mrKisaoLamb (approved):
lgtm

PR #587, opened 2023-11-01
Disable HTTP2 in webhooks

dprince:
Also bumps golang.org/x/net to v0.17.0

Jira: [OSPRH-468](https://issues.redhat.com//browse/OSPRH-468)

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


api/go.mod
```diff
@@ -67,3 +67,6 @@ require (
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+// Bump golang.org/x/net to avoid Rapid Reset CVE
+replace golang.org/x/net => golang.org/x/net v0.17.0 //allow-merging
```
gibizer, line 72: Do we really need this? The x/net is already on 0.17 in L45

go.mod
```diff
@@ -83,6 +83,9 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
+// Bump golang.org/x/net to avoid Rapid Reset CVE
+replace golang.org/x/net => golang.org/x/net v0.17.0 //allow-merging
```
gibizer, line 87: Ditto x/net is already on 0.17

api/go.mod
dprince: I think the consensus was to do it this way to eliminate the older x/net references from the go.sum files

dprince: removed it here. We have done this differently in other operators (and it is perhaps safer to be explicit as this is a CVE) but it is set to 0.17 so should be fine.

PR #596, opened 2023-11-14
Add cleaning nova-scheduler down services from db

mrkisaolamb:


controllers/novascheduler_controller.go
```diff
@@ -400,3 +416,76 @@ func (r *NovaSchedulerReconciler) ensureDeployment(
...
+			APIDatabasePasswordSelector,
+			CellDatabasePasswordSelector,
+			TransportURLSelector,
+		},
+		h.GetClient(),
+		&instance.Status.Conditions,
+		r.RequeueTimeout,
+	)
+	if err != nil {
+		l.Error(err, "Failed geting secret")
```
gibizer, line 480: * nit: getting
* Please add a hint to the log that in this case the nova-scheduler service is not removed from the nova_cell0 DB.

controllers/novascheduler_controller.go
```diff
@@ -400,3 +416,76 @@ func (r *NovaSchedulerReconciler) ensureDeployment(
...
+		return nil
```
gibizer, line 481: OK so we return here as we cannot delete the service without the secret but we don't want to block the reconcileDelete and the CR deletion as that would be worse.

controllers/novascheduler_controller.go
```diff
@@ -400,3 +416,76 @@ func (r *NovaSchedulerReconciler) ensureDeployment(
...
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		// TODO(gibi): add keystoneAuthURL here is that is also passed via
+		// the Secret. Also add DB and MQ user name here too if those are
+		// passed via the Secret
+		[]string{
+			ServicePasswordSelector,
+			APIDatabasePasswordSelector,
+			CellDatabasePasswordSelector,
+			TransportURLSelector,
```
gibizer, line 473: I think we only need the ServicePassword here and not the rest as the gopher client only depend on that password.

controllers/novascheduler_controller.go
```diff
@@ -400,3 +416,76 @@ func (r *NovaSchedulerReconciler) ensureDeployment(
...
+
+	allServices, err := services.ExtractServices(allPages)
+	if err != nil {
+		return err
+	}
+	for _, service := range allServices {
+		if service.State == "down" {
+			services.Delete(computeClient, service.ID)
+		}
+	}
```
gibizer, line 450: This is now called from the reconcileDelete so when the NovaScheduler CR is deleted. That probably means the Pods of the nova-scheduler service is also just being deleted. So there was no time for the nova-scheduler service to time out.

So I think during reconcileDelete it is OK to delete "up" services as well as we know that all the related Pods are deleted too. We probably has a race condition between the Pod delete and the reconcileDelete here. So it is possible that the Pod still runs when we execute this reconcileDelete cleanup code and hence the nova-scheduler binary can re-register itself.

Currently you are not calling cleanServiceFromNovaDb from any periodic and not calling during scale down of the StatefulSet. So after a such scale down the related nova-scheduler service is not removed from the DB.

mrkisaolamb: I removed reconcileDelete because scheduler is not optional so removing it will left ctr plane in not healthy state so probably always will be at least 1 scheduler. Now cleaning will be called only during reconciliation but still this operation is not blocking. In case when we will have some problems with reporting correct state in nova-api during reconcilation with scaling down scheduler the next run will fix this issue and clean db.

gibizer (commented):
Follow ups:
* refactor lib-common to be able to reuse the providerClient creation from there instead of copy pasting it

controllers/novascheduler_controller.go
gibizer: I agree. scheduler is not optional.

controllers/common.go
```diff
@@ -90,6 +96,9 @@ const (
 	// TransportURLSelector is the name of key in the internal cell
 	// Secret for the cell message bus transport URL
 	TransportURLSelector = "transport_url"
+
+	// defaultRequestTimeout is the default timeout duration for requests
+	defaultRequestTimeout = 10 * time.Second
```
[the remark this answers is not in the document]
mrkisaolamb: done

controllers/novascheduler_controller.go
```diff
@@ -197,6 +208,7 @@ func (r *NovaSchedulerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&novav1.NovaScheduler{}).
 		Owns(&v1.StatefulSet{}).
+		Owns(&keystonev1.KeystoneService{}).
```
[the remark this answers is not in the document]
mrkisaolamb: This is leftover so removed

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+}
+
+func (f *NovaAPIFixture) RecordRequest(r *http.Request) {
+	f.APIRequests = append(f.APIRequests, map[string]string{
+		"method": r.Method,
+		"path":   r.URL.Path,
+	})
+}
+
+func (f *NovaAPIFixture) GetRequests() []map[string]string {
```
gibizer, line 83: This is unused now and probably not needed if APIRequests field remains public

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+	fixture := AddNovaAPIFixture(log, server)
+	fixture.OwnsServer = true
+	return fixture
+}
+
+func (f *NovaAPIFixture) RecordRequest(r *http.Request) {
+	f.APIRequests = append(f.APIRequests, map[string]string{
+		"method": r.Method,
+		"path":   r.URL.Path,
+	})
```
gibizer, line 80: It feels enough for our needs so I'm OK with it. Later this can be extended or even changed to record the whole http.Request object (but maybe that is a bit much)

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+}
+
+func (f *NovaAPIFixture) GetRequests() []map[string]string {
+	return f.APIRequests
+}
+
+func (f *NovaAPIFixture) FindRequest(method string, path string) bool {
```
gibizer, line 87: This is OK for now. But later if more complex asserts are needed then I suggest to simply use the ginkgo matchers on the APIRequests field instead of building a lot of different helpers like this.

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+
+func (f *NovaAPIFixture) registerHandler(handler api.Handler) {
+	f.Server.AddHandler(f.URLBase+handler.Pattern, handler.Func)
+}
+
+func (f *NovaAPIFixture) registerNormalHandlers() {
+	f.registerHandler(api.Handler{Pattern: "/os-services/", Func: f.ServicesList})
+}
+
+// ServicesList responds with a valid keystone token
```
gibizer, line 118: copy paste docstring :)

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+func (f *NovaAPIFixture) ServicesList(w http.ResponseWriter, r *http.Request) {
+	f.LogRequest(r)
+	f.RecordRequest(r)
```
gibizer, line 121: As we discussed both of these can move to lib common as a wrapper applied in FakeAPIServer.AddHandler when you move the recoding fuctionality there.

test/functional/base_test.go
```diff
@@ -42,6 +45,194 @@ const (
...
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(204)
+	default:
+		f.UnexpectedRequest(w, r)
+		return
+	}
+}
+
+// HandleToken responds with a valid keystone token
+func ResponseHandleToken(keystoneURL string, computeURL string) string {
```
gibizer, line 187: as we discussed this can move to the KeystoneAPIFixture during a subsequent refactoring.

test/functional/nova_scheduler_test.go
```diff
@@ -512,3 +550,47 @@ var _ = Describe("NovaScheduler controller", func() {
...
+		novaAPIServer.Setup()
+		f := keystone_helper.NewKeystoneAPIFixtureWithServer(logger)
+		text := ResponseHandleToken(f.Endpoint(), novaAPIServer.Endpoint())
+		f.Setup(
+			api.Handler{Pattern: "/", Func: f.HandleVersion},
+			api.Handler{Pattern: "/v3/users", Func: f.HandleUsers},
+			api.Handler{Pattern: "/v3/domains", Func: f.HandleDomains},
+			api.Handler{Pattern: "/v3/auth/tokens", Func: func(w http.ResponseWriter, r *http.Request) {
+				switch r.Method {
+				case "POST":
```
gibizer, line 569: I expect this to be simplified once the `ResponseHandleToken` is moved to be part of the keystoneAPIFixture.

test/functional/nova_scheduler_test.go
```diff
@@ -512,3 +550,47 @@ var _ = Describe("NovaScheduler controller", func() {
...
+			k8sClient.Delete, ctx, CreateInternalTopLevelSecret(novaNames))
+		spec := GetDefaultNovaSchedulerSpec(novaNames)
+		spec["keystoneAuthURL"] = f.Endpoint()
+		DeferCleanup(
+			th.DeleteInstance, CreateNovaScheduler(novaNames.SchedulerName, spec))
+		DeferCleanup(novaAPIServer.Cleanup)
+	})
+	When("NovaScheduler down service is removed from api", func() {
+		It("during reconciling", func() {
+			th.SimulateStatefulSetReplicaReady(novaNames.SchedulerStatefulSetName)
```
gibizer, line 586: I would note in a comment that:
```
// We can assert the service cleanup behavior or the reconciler as the cleanup is triggered
// unconditionally at the end of every successful reconciliation. So we don't actually need
// to trigger a scaledown to see the down services are deleted. The novaAPIFixture is 
// set up in a way that it always respond with a  list of  services in down state.
```

test/functional/nova_scheduler_test.go
```diff
@@ -512,3 +550,47 @@ var _ = Describe("NovaScheduler controller", func() {
...
+			th.ExpectCondition(
+				novaNames.SchedulerName,
+				ConditionGetterFunc(NovaSchedulerConditionGetter),
+				condition.ReadyCondition,
+				corev1.ConditionTrue,
+			)
+			Expect(novaAPIServer.FindRequest("DELETE", "/compute/os-services/3")).To(BeTrue())
+		})
```
gibizer, line 594: I would assert that no other services are deleted.
e.g:
```
			Expect(novaAPIServer.FindRequest("DELETE", "/compute/os-services/1")).To(BeFalse())
			Expect(novaAPIServer.FindRequest("DELETE", "/compute/os-services/2")).To(BeFalse())
            // service 4 is a compute service so it should not be deleted by the scheduler reconciler
			Expect(novaAPIServer.FindRequest("DELETE", "/compute/os-services/4")).To(BeFalse())
```
// later 
for some reason the last assert fails. We have a bug somewhere

// later
I think I see it. It is not a bug in the code. The code first lists the services **with a server side filter for nova-scheduler** and then delete all the down services in that list. But the test fixture ignores the filter and returns not just nova-scheduler but nova-compute services. 

We have multiple options.
A) verify that the code asked a filtered list, this need addition to record the query params in APIRequest, and ignore the problem that it deletes service 4 with some comments in the test case
B) make the novaAPIFixture smarter to detect filter and return a filtered list.

In the long run B) will be better as it will allow us to us the same fixture code for each service type deletion test case.

gibizer (approved):
The code looks correct so I approve this. We pre-agreed on bunch of test refactors and I found some extra during the test code review (i.e. server side filtering).

test/functional/base_test.go
gibizer: We might need the Query params though :) See below

gibizer:
The pre-commit result is interesting
```
# github.com/openstack-k8s-operators/nova-operator/test/functional_test
test/functional/nova_scheduler_test.go:216:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
test/functional/nova_scheduler_test.go:409:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
test/functional/nova_scheduler_test.go:449:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
test/functional/nova_scheduler_test.go:506:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
test/functional/nova_scheduler_test.go:527:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
test/functional/nova_scheduler_test.go:534:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
```

gibizer:
> The pre-commit result is interesting
> 
> ```
> # github.com/openstack-k8s-operators/nova-operator/test/functional_test
> test/functional/nova_scheduler_test.go:216:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> test/functional/nova_scheduler_test.go:409:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> test/functional/nova_scheduler_test.go:449:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> test/functional/nova_scheduler_test.go:506:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> test/functional/nova_scheduler_test.go:527:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> test/functional/nova_scheduler_test.go:534:5: L001: 'Expect' in the 'Eventually' block should be called via a local 'Gomega' parameter. Change the declaration of the function passed to the 'Eventually' block to take a parameter with type 'Gomega' and use that to call 'Expect'
> ```

The error does not make sense based on the code
https://github.com/openstack-k8s-operators/nova-operator/pull/596/files#diff-da53d7e777cd5259b17811d67ecb8d5045ec076783904c4426c6ce9e76e0a118R531-R536

And I cannot reproduce the same error locally with pre-commit

test/functional/nova_scheduler_test.go
```diff
@@ -15,12 +15,18 @@ package functional_test
 
 import (
 	"encoding/json"
+	"fmt"
+	"net/http"
 
 	networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 	. "github.com/onsi/gomega"
```
gibizer, line 24: This together causing the operator-lint error.

test/functional/nova_scheduler_test.go
```diff
@@ -33,9 +39,38 @@ var _ = Describe("NovaScheduler controller", func() {
...
+		DeferCleanup(f.Cleanup)
+		keystone_name := keystone.CreateKeystoneAPIWithFixture(novaNames.NovaName.Namespace, f)
+		keystone_instance := keystone.GetKeystoneAPI(keystone_name)
+		keystone_instance.Status = keystonev1.KeystoneAPIStatus{
+			APIEndpoints: map[string]string{
+				"public":   f.Endpoint(),
+				"internal": f.Endpoint(),
+			},
+		}
+		gomega.Expect(th.K8sClient.Status().Update(th.Ctx, keystone_instance.DeepCopy())).Should(gomega.Succeed())
```
gibizer, line 68: copy paste, we don't need to import and use gomega with name as it is dot imported.

test/functional/nova_scheduler_test.go
gibizer: fixed.

test/functional/nova_scheduler_test.go
gibizer: fixed.

gibizer:
I fixed the pre-commit operator-lint and make tidy issues to be able to unblock this to merge.

gibizer (approved):
Re-approved after trivial lint fixes. With the pending followups recorded in https://github.com/openstack-k8s-operators/nova-operator/pull/596#pullrequestreview-1747668435

PR #598, opened 2023-11-16
Do not run nova services as root

gibizer:
Implements: https://issues.redhat.com/browse/OSPRH-1374

### DIFF
--- a/pkg/novaconductor/dbsync.go
+++ b/pkg/novaconductor/dbsync.go
@@ -26,11 +26,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-const (
-	// cellDBSyncCommand - the command to be used to run db sync for the cell DB
-	cellDBSyncCommand = "/usr/local/bin/kolla_set_configs && /bin/sh -c /var/lib/openstack/bin/dbsync.sh"
+	"k8s.io/utils/ptr"
 )
 
 // CellDBSyncJob - define a batchv1.Job to be run to apply the cel DB schema
@@ -39,13 +35,11 @@ func CellDBSyncJob(
 	labels map[string]string,
 	annotations map[string]string,
 ) *batchv1.Job {
-	runAsUser := int64(0)
-
 	args := []string{"-c"}
 	if instance.Spec.Debug.StopJob {
 		args = append(args, common.DebugCommand)
 	} else {
-		args = append(args, cellDBSyncCommand)
+		args = append(args, nova.KollaServiceCommand)
 	}
 
 	envVars := map[string]env.Setter{}
@@ -81,13 +75,13 @@ func CellDBSyncJob(
 							Args:  args,
 							Image: instance.Spec.ContainerImage,
 							SecurityContext: &corev1.SecurityContext{
-								RunAsUser: &runAsUser,
+								RunAsUser: ptr.To(nova.NovaUserID),
 							},
 							Env: env,
 							VolumeMounts: []corev1.VolumeMount{
 								nova.GetConfigVolumeMount(),
 								nova.GetScriptVolumeMount(),
-								nova.GetKollaConfigVolumeMount("nova-conductor"),
+								nova.GetKollaConfigVolumeMount("nova-conductor-dbsync"),
 							},
 						},
 					},
--- a/templates/nova-manage/config/cell-mapping-config.json
+++ b/templates/nova-manage/config/cell-mapping-config.json
@@ -1,5 +1,5 @@
 {
-    "command": "/usr/sbin/httpd -DFOREGROUND",
+    "command": "/bin/ensure_cell_mapping.sh",
     "config_files": [
         {
             "source": "/var/lib/openstack/config/nova-blank.conf",
@@ -21,22 +21,10 @@
             "optional": true
         },
         {
-            "source": "/var/lib/openstack/config/httpd.conf",
-            "dest": "/etc/httpd/conf/httpd.conf",
-            "owner": "apache",
-            "perm": "0644"
-        }
-    ],
-    "permissions": [
-        {
-            "path": "/var/log/nova",
-            "owner": "nova:apache",
-            "recurse": true
-        },
-        {
-            "path": "/etc/httpd/run/",
-            "owner": "nova:apache",
-            "recurse": true
+            "source": "/var/lib/openstack/bin/ensure_cell_mapping.sh",
+            "dest": "/bin/",
+            "owner": "nova",
+            "perm": "0700"
         }
     ]
 }
--- a/templates/novametadata/config/httpd.conf
+++ b/templates/novametadata/config/httpd.conf
@@ -18,6 +18,8 @@ LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" combine
 LogFormat "%{X-Forwarded-For}i %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" proxy
 
 SetEnvIf X-Forwarded-For "^.*\..*\..*\..*" forwarded
+ErrorLog /dev/stderr
+TransferLog /dev/stdout
 CustomLog /dev/stdout combined env=!forwarded
 CustomLog /dev/stdout proxy env=forwarded
 ## set default apache log level to info from warning
--- a/templates/novametadata/config/nova-metadata-config.json
+++ b/templates/novametadata/config/nova-metadata-config.json
@@ -32,6 +32,11 @@
       "path": "/var/log/nova",
       "owner": "nova:apache",
       "recurse": true
+    },
+    {
+      "path": "/etc/httpd/run/",
+      "owner": "nova:apache",
+      "recurse": true
     }
   ]
 }
--- a/pkg/nova/cellmapping.go
+++ b/pkg/nova/cellmapping.go
@@ -4,14 +4,11 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	common "github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/v1beta1"
-)
-
-const (
-	cellMappingCommand = "/usr/local/bin/kolla_set_configs && /var/lib/openstack/bin/ensure_cell_mapping.sh"
 )
 
 func CellMappingJob(
@@ -22,13 +19,11 @@ func CellMappingJob(
 	inputHash string,
 	labels map[string]string,
 ) *batchv1.Job {
-	runAsUser := int64(0)
-
 	args := []string{"-c"}
 	if cell.Spec.Debug.StopJob {
 		args = append(args, common.DebugCommand)
 	} else {
-		args = append(args, cellMappingCommand)
+		args = append(args, KollaServiceCommand)
 	}
 
 	envVars := map[string]env.Setter{}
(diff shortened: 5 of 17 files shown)


gibizer:
db sync job fails but must gather does not collect the job's log :/ I need to try this with a local build

gibizer:
@SeanMooney you mentioned that kolla set config needs sudo and that the sudo is managed by rootwrap in the k8s nova images.

```
❯ oc exec -t nova-cell0-conductor-0 -- cat /etc/sudoers.d/nova
Defaults:nova !requiretty

nova ALL = (root) NOPASSWD: /usr/bin/nova-rootwrap /etc/nova/rootwrap.conf *
nova ALL = (root) NOPASSWD: /usr/bin/privsep-helper *
```

```
❯ oc exec -t nova-cell0-conductor-0 -- grep filters_path /etc/nova/rootwrap.conf
filters_path=/etc/nova/rootwrap.d,/usr/share/nova/rootwrap
```

```
❯ oc exec -t nova-cell0-conductor-0 -- ls /etc/nova/rootwrap.d /usr/share/nova/rootwrap
ls: cannot access '/etc/nova/rootwrap.d': No such file or directory
ls: cannot access '/usr/share/nova/rootwrap': No such file or directory
command terminated with exit code 2
```

It seems to me that the sudo config is incomplete as we have no filters defined. For dbsync we directly call `kolla_set_configs` at 
https://github.com/openstack-k8s-operators/nova-operator/blob/main/pkg/novaconductor/dbsync.go#L33 
For the service containers we call `/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start` 

If I understand correctly kolla_set_config needs sudo. How should we provide it in this env?

gibizer:
> @SeanMooney you mentioned that kolla set config needs sudo and that the sudo is managed by rootwrap in the k8s nova images.
> 
> ```
> ❯ oc exec -t nova-cell0-conductor-0 -- cat /etc/sudoers.d/nova
> Defaults:nova !requiretty
> 
> nova ALL = (root) NOPASSWD: /usr/bin/nova-rootwrap /etc/nova/rootwrap.conf *
> nova ALL = (root) NOPASSWD: /usr/bin/privsep-helper *
> ```
> 
> ```
> ❯ oc exec -t nova-cell0-conductor-0 -- grep filters_path /etc/nova/rootwrap.conf
> filters_path=/etc/nova/rootwrap.d,/usr/share/nova/rootwrap
> ```
> 
> ```
> ❯ oc exec -t nova-cell0-conductor-0 -- ls /etc/nova/rootwrap.d /usr/share/nova/rootwrap
> ls: cannot access '/etc/nova/rootwrap.d': No such file or directory
> ls: cannot access '/usr/share/nova/rootwrap': No such file or directory
> command terminated with exit code 2
> ```
> 
> It seems to me that the sudo config is incomplete as we have no filters defined. For dbsync we directly call `kolla_set_configs` at https://github.com/openstack-k8s-operators/nova-operator/blob/main/pkg/novaconductor/dbsync.go#L33 For the service containers we call `/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start`
> 
> If I understand correctly kolla_set_config needs sudo. How should we provide it in this env?

Ahh there is an extra line the sudoers file:
```
# anyone in the kolla group may run /usr/local/bin/kolla_set_configs as the
# root user via sudo without password confirmation
%kolla ALL=(root) NOPASSWD: /usr/local/bin/kolla*
```
that suggest that we only need a user in the kolla group for the kolla commands to work

gibizer:
So lets try to switch from root(0) to nova(42436) in the pods

SeanMooney:
so this is a bug we should never call kolla_set_configs directly 

we have two options

cellDBSyncCommand = "sudo /usr/local/bin/kolla_set_configs && /bin/sh -c /var/lib/openstack/bin/dbsync.sh"

as you said

or

cellDBSyncCommand = "/usr/local/bin/kolla_start" and provide a kolla config.json file with 
"/bin/sh -c /var/lib/openstack/bin/dbsync.sh" as the command for kolla start to run after it has set up the config.

that would be my preference.

i.e. instead of bypassing the kolla image API we should correctly provide the kolla config.json to the job container and specify the command as normal in that.

gibizer:
> so this is a bug we should never call kolla_set_configs directly
> 
> we have two options
> 
> cellDBSyncCommand = "sudo /usr/local/bin/kolla_set_configs && /bin/sh -c /var/lib/openstack/bin/dbsync.sh"
> 
> as you said
> 
> or
> 
> cellDBSyncCommand = "/usr/local/bin/kolla_start" and provide a kolla config.json file with "/bin/sh -c /var/lib/openstack/bin/dbsync.sh" as the command for kolla start to run after it has set up the config.
> 
> that would be my preference.
> 
> i.e. instead of bypassing the kolla image API we should correctly provide the kolla config.json to the job container and specify the command as normal in that.

Thanks. I will rework this.

FYI, we have an even worse situation in placement-operator https://github.com/openstack-k8s-operators/placement-operator/pull/107#issuecomment-1819031459 as I think the placement image itself is broken (placement user is no in the kolla group)

gibizer:
This is now using sudo for kolla_set_config just to see if that make the rest of the pieces work. If so I can add another commit to use kolla config files to execute the job commands.

gibizer:
recheck

gibizer:
recheck zuul nodepool should be back in business

gibizer:
Still no success even with sudo. And must gather still not provide job logs (and the related pod is deleted) so I need to go back and deploy it locally to see the error.

pkg/nova/volumes.go
```diff
@@ -28,7 +28,9 @@ const (
 
 var (
 	configMode int32 = 0640
-	scriptMode int32 = 0740
+	// NOTE(gibi): scripts are mounted as root but kolla runs them with
+	// the nova user. Hence the mode needs to be wide open here.
+	scriptMode int32 = 0777
```
SeanMooney, line 33: thank you for documenting that for future us

templates/nova-manage/config/cell-mapping-config.json
```diff
@@ -0,0 +1,24 @@
+{
+    "command": "/var/lib/openstack/bin/ensure_cell_mapping.sh",
```
SeanMooney, line 2: with that siad if this is the script that need 777 to execute you can use
the permsions section of the file to update it to nova

https://docs.openstack.org/kolla/latest/admin/kolla_api.html#format-of-the-configuration-file

  "permissions": [
        {
            "path": "/var/lib/openstack/bin",
            "owner": "nova:nova",
            "recurse": true
        }

templates/novametadata/config/nova-metadata-config.json
```diff
@@ -32,6 +32,11 @@
       "path": "/var/log/nova",
       "owner": "nova:apache",
       "recurse": true
+    },
+    {
+      "path": "/etc/httpd/run/",
+      "owner": "nova:apache",
+      "recurse": true
```
SeanMooney, line 39: +1
that makes sense

templates/novametadata/config/httpd.conf
```diff
@@ -18,6 +18,8 @@ LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" combine
 LogFormat "%{X-Forwarded-For}i %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" proxy
 
 SetEnvIf X-Forwarded-For "^.*\..*\..*\..*" forwarded
+ErrorLog /dev/stderr
+TransferLog /dev/stdout
```
SeanMooney, line 22: out of interest what addtioanl logs are these

i assume they are the generic apache logs unrelated to the virtual hosts config.

SeanMooney (commented):
im +1.5 on this im just wondering if we really need to change the script mode or not

pkg/novaconductor/dbsync.go
```diff
@@ -39,13 +35,11 @@ func CellDBSyncJob(
...
 	annotations map[string]string,
 ) *batchv1.Job {
-	runAsUser := int64(0)
-
 	args := []string{"-c"}
 	if instance.Spec.Debug.StopJob {
 		args = append(args, common.DebugCommand)
 	} else {
-		args = append(args, cellDBSyncCommand)
+		args = append(args, nova.KollaServiceCommand)
```
bogdando, line 42: where goes the missing ``/bin/sh -c /var/lib/openstack/bin/dbsync.sh`` part now?

templates/nova-manage/config/cell-mapping-config.json
gibizer: I tried that it does not work. It say it is on a read only file system

pkg/novaconductor/dbsync.go
bogdando: og I see it's in nova-conductor-dbsync-config.json

templates/novametadata/config/httpd.conf
gibizer: Yes, generic apache log. This is here just to avoid another log file to be created in another directory that has no nova persmissions. I think we want to have logs on stdout/err instead of in a file so we can easily collect them

templates/nova-manage/config/cell-mapping-config.json
SeanMooney: ah right the configmaps are we would have to use kolla to copy the script using the "config_files" element instead.

i.e. copy them to /bin with the correct permissions so that they can be run

templates/novametadata/config/httpd.conf
SeanMooney: apache  logs are fine to have on stdout/stderr yes as long as the sidecar is used for the wsgi app logs so we can keep those separate so I'm cool with this

bogdando:
lgtm

pkg/novaconductor/dbsync.go
gibizer: in the command section of the relevant kolla config file

gibizer: https://github.com/openstack-k8s-operators/nova-operator/pull/598/files#diff-39d323540d0d9d8440e05e672dd08ebf4c73e85657fd5938c05981da31adeb64R2

templates/nova-manage/config/cell-mapping-config.json
gibizer: I did what you suggested and this way we can fall back to a normal script mount permission.

SeanMooney (approved):

gibizer:
Setting scc to nonroot-v2 (or nonroot) does not work with kolla + sudo. It seem sudo needs uid 0 to exists
```
❯ oc logs nova-cell0-conductor-db-sync-bbtdl
+ sudo -E kolla_set_configs
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
```

bogdando:
> Setting scc to nonroot-v2 (or nonroot) does not work with kolla + sudo. It seem sudo needs uid 0 to exists
> 
> ```
> ❯ oc logs nova-cell0-conductor-db-sync-bbtdl
> + sudo -E kolla_set_configs
> sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
> ```

To have a user with uid 0 mapped to some (nova) id user in the host, usernsmappings should work for cri-o in openshift the similar thay it works for podman, see https://github.com/cri-o/cri-o/issues/5294

gibizer:
> nonroot-v2

Tried the manual mapping in https://github.com/openstack-k8s-operators/nova-operator/pull/605 it does not seem to work.

gibizer:
> > Setting scc to nonroot-v2 (or nonroot) does not work with kolla + sudo. It seem sudo needs uid 0 to exists
> > ```
> > ❯ oc logs nova-cell0-conductor-db-sync-bbtdl
> > + sudo -E kolla_set_configs
> > sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
> > ```
> 
> To have a user with uid 0 mapped to some (nova) id user in the host, usernsmappings should work for cri-o in openshift the similar thay it works for podman, see [cri-o/cri-o#5294](https://github.com/cri-o/cri-o/issues/5294)

Tried the manual mapping in https://github.com/openstack-k8s-operators/nova-operator/pull/605 it does not seem to work.

bogdando:
> Tried the manual mapping in #605 it does not seem to work.

there is still something to research in that regard

PR #599, opened 2023-11-16
Do not use sidecar for logging for non apache services

gibizer:
Neither novncproxy nor nova-compute need sidecar for logging


SeanMooney (approved):
yep that is correct they do not.

we only need that when running under a webserver like apache

mrKisaoLamb (approved):
lgtm

PR #606, opened 2023-11-27
[api]ensure sidecar terminates on SIGTERM

gibizer:
We noticed that nova-api and nova-metadata pods do not immediately terminate on delete as the logging sidecar ignores SIGTERM. This PR adds the usage of dump-init to the sidecar command to ensure it handles SIGTERM properly

Fixes: [OSPRH-770](https://issues.redhat.com//browse/OSPRH-770)

### DIFF
--- a/pkg/novaapi/deployment.go
+++ b/pkg/novaapi/deployment.go
@@ -129,9 +129,16 @@ func StatefulSet(
 						{
 							Name: instance.Name + "-log",
 							Command: []string{
-								"/bin/bash",
+								"/usr/bin/dumb-init",
 							},
-							Args:  []string{"-c", "tail -n+1 -F /var/log/nova/nova-api.log"},
+							Args: []string{
+								"--single-child",
+								"--",
+								"/usr/bin/tail",
+								"-n+1",
+								"-F",
+								"/var/log/nova/nova-api.log",
+							},
 							Image: instance.Spec.ContainerImage,
 							SecurityContext: &corev1.SecurityContext{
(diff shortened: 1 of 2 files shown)


pkg/novametadata/deployment.go
```diff
@@ -131,7 +131,14 @@ func StatefulSet(
 							Command: []string{
 								"/bin/bash",
```
karelyatin, line 132: this should be also switched.

gibizer: Done.

mrkisaolamb (approved):
lgtm

karelyatin (commented):
/lgtm

pkg/novaapi/deployment.go
```diff
@@ -129,9 +129,16 @@ func StatefulSet(
 						{
 							Name: instance.Name + "-log",
 							Command: []string{
-								"/bin/bash",
```
SeanMooney, line 132: so the entrypoint in the kolla contianer is

    Entrypoint: "dumb-init --single-child --"

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/tcib/base/base.yaml#L44C18-L44C45

so i was expecting command to map to docker/podmans command

it would appear that k8s/openshfit is not following the normal rules
for how entrypoint and command interact

https://stackoverflow.com/a/66967227

so when i orgianlly wrote this i was expect it to be wrapped in dumb init already

SeanMooney (approved):
yes this is required

PR #608, opened 2023-11-28
[Make]Run kubefwd when operator runs outside of OCP

gibizer:


gibizer:
```
2023-11-28T11:26:28.399+0100	ERROR	Failed cleaning services from nova db	{"controller": "novascheduler", "controllerGroup": "nova.openstack.org", "controllerKind": "NovaScheduler", "NovaScheduler": {"name":"nova-scheduler","namespace":"openstack"}, "namespace": "openstack", "name": "nova-scheduler", "reconcileID": "7de45ff9-c5b4-4d81-8a4c-06c140f41933", "error": "Get \"https://nova-public-openstack.apps-crc.testing/v2.1/os-services?binary=nova-scheduler\": x509: certificate signed by unknown authority"}
```

hack/run_with_kubefwd.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -ex
+
+pushd bin
```
SeanMooney, line 4: is there a reason you choose to do it this way instead of using a make target like we have for operator-lint or the SDK?

SeanMooney (commented):
i think for my own usage i would continue to use
the routing approach from
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs/contributing/development_environment.md#route-networks
since if i run locally its almost exclusively via vscode

although i do use this target to inially configure the webhooks so i don't exclusively use vscode

hack/run_with_kubefwd.sh
gibizer: no specific reason, it is started as a hack :)

PR #610, opened 2023-11-29
Refactor using of gophercloud client

mrkisaolamb:
Improve testing and code

Depends on: https://github.com/openstack-k8s-operators/lib-common/pull/409


test/functional/nova_scheduler_test.go
```diff
@@ -578,18 +578,22 @@ var _ = Describe("NovaScheduler controller cleaning", func() {
...
 			th.DeleteInstance, CreateNovaScheduler(novaNames.SchedulerName, spec))
-		DeferCleanup(novaAPIServer.Cleanup)
+		DeferCleanup(novaAPIFixture.Cleanup)
 	})
 	When("NovaScheduler down service is removed from api", func() {
 		It("during reconciling", func() {
+			// We can assert the service cleanup behavior or the reconciler as the cleanup is triggered
+			// unconditionally at the end of every successful reconciliation. So we don't actually need
+			// to trigger a scaledown to see the down services are deleted. The novaAPIFixture is
+			// set up in a way that it always respond with a  list of  services in down state.
```
gibizer, line 588: supernit: s/of  services/of services/

gibizer (commented):
good step forward. Looks good just waiting for the dependency to land

controllers/common.go
```diff
@@ -40,7 +37,6 @@ import (
 	"github.com/openstack-k8s-operators/nova-operator/pkg/nova"
 
 	gophercloud "github.com/gophercloud/gophercloud"
-	gophercloud_openstack "github.com/gophercloud/gophercloud/openstack"
```
SeanMooney, line 43: you likely inted to keep this?

controllers/common.go
```diff
@@ -544,73 +540,9 @@ func getNovaClient(ctx context.Context,
...
-	if err != nil {
-		return nil, ctrl.Result{}, err
-	}
-
-	providerClient.HTTPClient = httpClient
-	providerClient.HTTPClient.Transport = transport
-
-	// authenticate the client
-	err = gophercloud_openstack.Authenticate(providerClient, opts)
+	computeClient, err := openstack.GetNovaOpenStackClient(cfg)
```
SeanMooney, line 543: openstack is now undefiend.

SeanMooney (changes requested):

controllers/common.go
SeanMooney: ah "github.com/openstack-k8s-operators/lib-common/modules/openstack"

so this is defiend but the function is not found because the github action does not understand depends on.

mrkisaolamb:
lib-common function need small refactor to work fully with internal and public endpoints. Currently even if keystone is internal we still use public endpoints

controllers/common.go
```diff
@@ -537,82 +533,24 @@ func (r *ReconcilerBase) ensureMetadataDeleted(
...
-	err = gophercloud_openstack.Authenticate(providerClient, opts)
-	if err != nil {
-		return nil, ctrl.Result{}, err
-	}
-
 	// create the compute client using previous providerClient
 	endpointOpts := gophercloud.EndpointOpts{
-		Region: cfg.Region,
+		Region:       cfg.Region,
+		Availability: gophercloud.AvailabilityInternal,
```
gibizer, line 549: Cool, so this will ensure that the client uses the nova internal endpoint.

PR #611, opened 2023-11-30
Fix issue with not verified cert during cleaning

mrkisaolamb:
Apply similar fix to: https://github.com/openstack-k8s-operators/keystone-operator/commit/2ca06317fed3d31fd938016bee64a3df3e6c1f03

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/http"
+	"net/url"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


controllers/common.go
```diff
@@ -537,12 +538,24 @@ func (r *ReconcilerBase) ensureMetadataDeleted(
...
 	h *helper.Helper, authURL string, adminUser string, authPassword string, timeout time.Duration) (*gophercloud.ServiceClient, ctrl.Result, error) {
 
+	parsedAuthURL, err := url.Parse(authURL)
+	if err != nil {
+		return nil, ctrl.Result{}, err
+	}
+
+	tlscfg := &openstack.TLSConfig{}
+	if parsedAuthURL.Scheme == "https" {
+		// TODO: (ksambor) for now just set to insecure, when keystone got
```
SeanMooney, line 548: if we merge this we need to file an jira bug or gitbub issue to fix this and target it for Q1 if we cant fixt it properly by the end of the year.

SeanMooney: this is related to https://issues.redhat.com/browse/OSP-19151
so i would like to have a bug that is set as a blocker for that.
the intent is to complete that by the end for q4
@stuggi @olliewalsh does that sound ok to ye ^

PR #612, opened 2023-12-01

mrkisaolamb:


gibizer (approved):

mrkisaolamb:
recheck

mrkisaolamb:
recheck

.zuul.yaml
```diff
@@ -134,6 +134,8 @@
       cifmw_tempest_tests_skipped:
       # We don't have a running cinder-volume
         - cinder_tempest_plugin.
+         # TODO(samborka): remove this once we have a running barbican
+        - barbican_tempest_plugin.
```
gibizer, line 138: * Please move this either above L136 or below one of our sections. Each section is separated with a dedented comment describing the reason why the set of tests are disabled. Now you added this in the middle of the cinder section.
* this skip alone won't be enough. We need to skip the other test cases that started to run due to barbican is now enabled in the deployment: `tempest.scenario.test_encrypted_cinder_volumes.TestEncryptedCinderVolumes`

mrkisaolamb:
recheck

gibizer:
Force merging this as we suspect that the image build failure blocking this is goes away with this fix.

PR #613, opened 2023-12-01

mrkisaolamb:
Cherry pick from https://github.com/openstack-k8s-operators/nova-operator/pull/612


gibizer (changes requested):
I think we need to change this as well
https://github.com/openstack-k8s-operators/nova-operator/blob/main/config/default/manager_default_images.yaml#L24-L25

mrkisaolamb:
> I think we need to change this as well https://github.com/openstack-k8s-operators/nova-operator/blob/main/config/default/manager_default_images.yaml#L24-L25

in dev-preview2 we don't have this value https://github.com/openstack-k8s-operators/nova-operator/blob/dev-preview2/config/default/manager_default_images.yaml

gibizer (approved):

PR #615, opened 2023-12-01
[zuul]Use the common content provider

gibizer:


.zuul.yaml
```diff
@@ -1,20 +1,4 @@
...
-      This job builds the nova-operator and provides it to child jobs
-      It also rebuilds the openstack-operator using the updated nova operator.
-    provides:
-      - "nova-operator-content"
-    vars:
-      cifmw_operator_build_org: openstack-k8s-operators
-      cifmw_operator_build_operators:
-        - name: "openstack-operator"
-          src: "~/src/github.com/{{ cifmw_operator_build_org }}/openstack-operator"
-          image_base: nova
```
SeanMooney, line 15: we use this one to specifically ensure we build the opesntack operator with the new content form the nova-operator patch under test.

i don't think we can make this change and have the ci work correctly unless they have change that job to rebuild every operator in the required projects automatically ?

SeanMooney (changes requested):

gibizer:
This was a failed attempt to troubleshoot the build failure.

PR #620, opened 2023-12-04
 Clean conductors from db during rescaling

mrkisaolamb:


controllers/novaconductor_controller.go
```diff
@@ -446,6 +453,41 @@ func (r *NovaConductorReconciler) ensureDeployment(
 	return ctrl.Result{}, nil
 }
 
+func (r *NovaConductorReconciler) cleanServiceFromNovaDb(
```
gibizer, line 456: This can be generalized now as it only differs from the scheduler one by `Binary: "nova-conductor"` so that can be passed as a parameter and also the keystoneAuthURL can be taken as a parameter.

test/functional/novaconductor_controller_test.go
```diff
@@ -637,3 +641,46 @@ var _ = Describe("NovaConductor controller", func() {
...
+	When("NovaScheduler down service is removed from api", func() {
+		It("during reconciling", func() {
+			th.SimulateStatefulSetReplicaReady(cell0.ConductorStatefulSetName)
+			th.ExpectCondition(
+				cell0.ConductorName,
+				ConditionGetterFunc(NovaConductorConditionGetter),
+				condition.DeploymentReadyCondition,
+				corev1.ConditionTrue,
+			)
+			Expect(novaAPIServer.FindRequest("DELETE", "/compute/os-services/3", "")).To(BeTrue())
```
gibizer, line 683: Now that fixture records the query param please assert that the code queried a list of conductor services.

controllers/novaconductor_controller.go
mrkisaolamb: done

test/functional/novaconductor_controller_test.go
mrkisaolamb: done a68cc8cf75f2e0b78231211bc559d4b6647f5fc9

controllers/common.go
```diff
@@ -102,6 +103,35 @@ type conditionsGetter interface {
 	GetConditions() condition.Conditions
 }
 
+func cleanNovaServiceFromNovaDb(ctx context.Context,
+	h *helper.Helper, authURL string, adminUser string,
+	authPassword string, timeout time.Duration, l logr.Logger, serviceName string) error {
+	computeClient, _, err := getNovaClient(ctx, h, authURL, adminUser, authPassword, defaultRequestTimeout, l)
```
gibizer, line 109: * It seems that the result return value of getNovaClient is unused so that can also be removed. 
* also as far as I see the `defaultRequestTimeout` is also unused in getNovaClient so the whole constant can be removed

controllers/novaconductor_controller.go
```diff
@@ -462,28 +461,8 @@ func (r *NovaConductorReconciler) cleanServiceFromNovaDb(
...
-	if err != nil {
-		return err
-	}
-	for _, service := range allServices {
-		if service.State == "down" {
-			services.Delete(computeClient, service.ID)
-		}
-	}
+	err := cleanNovaServiceFromNovaDb(ctx, h, instance.Spec.KeystoneAuthURL,
+		instance.Spec.ServiceUser, authPassword, defaultRequestTimeout, l, "nova-conductor")
```
gibizer, line 465: I think we can express that the authURL and the user name is coming from the same instance with some refactoring and shrink the parameter list.

mrkisaolamb:
> I see a list of code cleanup possibilities. I did them in [a81aba2](https://github.com/openstack-k8s-operators/nova-operator/commit/a81aba228a1a49d7d3d7d5cc1de165af13132675) If you like it feel free to squash it in.

I cherry-pick commit

gibizer (approved):

PR #622, opened 2023-12-08
Improve naming validation to allign with k8s

mrkisaolamb:


gibizer (approved):

gibizer:
recheck
it seems tempest ended abruptly but while tests was still in progress, but the logs does not show that it was as test case or an full tempest timeout

PR #624, opened 2023-12-08
[tempest]Reduce running tests

gibizer:
Only run the api tests for compute and scenario tests.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -123,32 +123,21 @@
             compute.min_compute_nodes: 3
             compute-feature-enabled.vnc_console: true
             compute-feature-enabled.stable_rescue: true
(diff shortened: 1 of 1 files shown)


gibizer:
/hold
I see strange tempest failures where test remaining inprogress and the tempest exits with error. I guess there is a global timeout somewhere we started to hit. Let's see if I reduce the number of test cases then we can be below that limit

gibizer:
the tempest part of the run in this PR took 38 minutes, on main it takes 49 minutes and fails.

gibizer:
/unhold
test reduction helpes the tempest test run in now 38 (down from 49) minutes

mrkisaolamb (approved):
lgtm

PR #630, opened 2023-12-15
[envtest]Cover the size of nova-metadata-neutron-config

gibizer:
We had a regression recently due to not checking in the test that the nova-metadata-neutron-config should have a single key.

This is adding the missing test coverage. The fix was done in #628


stuggi (approved):
/lgtm

PR #635, opened 2023-12-20
[must-gather]Add GMR file trigger support

gibizer:
The API services can only use the file trigger for GMR as apache
disables signal handling. This PR adds the file trigger config for
every server.
GMR can be triggered by
```
  oc exec  <service pod> -- touch /var/lib/nova
```
Optionally non apache based services produce GMR also via
```
  oc exec <service pod> -- kill -USR2 1
```

Related-to: https://issues.redhat.com/browse/OSPRH-730

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -309,3 +309,7 @@ user_domain_name = {{ .default_user_domain}}
 
 [upgrade_levels]
 compute = auto
+
+[oslo_reports]
+# api services need file based GMR trigger as apache disables signal handling
+file_event_handler=/var/lib/nova


templates/nova.conf
```diff
@@ -309,3 +309,10 @@ user_domain_name = {{ .default_user_domain}}
 
 [upgrade_levels]
 compute = auto
+
+
+{{if eq .service_name "nova-api" "nova-metadata"}}
```
SeanMooney, line 314: io would prefer ot just add this for all services.
even if the conductor/schduler and comptue agent can use the SIGUSR2
its cleaner to document one way to do this.

also enabling the file trigger does not disable the SIGUSR2

templates/novaapi/config/nova-api-config.json
```diff
@@ -25,6 +25,12 @@
             "dest": "/etc/httpd/conf/httpd.conf",
             "owner": "apache",
             "perm": "0644"
+        },
+        {
+            "source": "/var/lib/openstack/config/nova-blank.conf",
+            "dest": "/var/log/nova/gmr_trigger",
+            "owner": "nova",
+            "perm": "0666"
```
SeanMooney, line 33: i guess that is ok
we dont actually need this to exist when its first touched it should work too but sure that is fine too.

SeanMooney (changes requested):

templates/nova.conf
gibizer: Good  point. If we want to manually trigger GMRs then having a single way is better. I will switch every service to file trigger

templates/novaapi/config/nova-api-config.json
gibizer: Unfortunately not. GMR fails to initialize during service startup if the file trigger does not exists

templates/nova.conf
gibizer: Done

templates/novaapi/config/nova-api-config.json
gibizer: I figured that we can use use dirs as a trigger. So I switched to /var/lib/nova directory itself as the trigger.

SeanMooney (approved):
i was not aware we coudl jsut use a dir but ok that also works

perhaps not as intuitive as 
/var/lib/nova/gmr_trigger but it does not require changes to edpm-ansible or the operator to create an empty file.

so over all i like this approach more thanks gibi

PR #636, opened 2023-12-21
Improve log msg for nova compute service to avoid confusion

mrkisaolamb:


SeanMooney:
recheck

api/v1beta1/conditions.go
```diff
@@ -50,8 +50,8 @@ const (
 	// NovaComputeServiceConfigReady indicates when the compute service config
 	// is ready for the given NovaCell
 	NovaComputeServiceConfigReady condition.Type = "NovaComputeServiceConfigReady"
-	// NovaAllComputesReadyCondition indicates that every defined NovaCompute is ready but undiscovered
-	NovaAllComputesReadyCondition condition.Type = "NovaAllComputesReady"
+	// NovaAllControlPlaneComputesReadyCondition indicates that every defined NovaCompute is ready but undiscovered
+	NovaAllControlPlaneComputesReadyCondition condition.Type = "NovaAllComputesReady"
```
SeanMooney, line 54: the kuttl test failure might be real so i recheck it to get new results

ill pull this locally and see if i can repodcue

mrkisaolamb: It appears that the issue is not related to these changes. The only failure is in the tempest job, and it is related to selinux.

gibizer, line 54: I would consider changing the type value as that is shown in the CR status to clarify what computes are ready.

mrkisaolamb: thanks done, sorry for lag

gibizer (approved):

PR #638, opened 2023-12-21
Remove CI-framework ci_framework dir references

pablintino:
Since the framework is removing that unnecessary level, that is now a symlink of the root level dirs, we can safely get rid off that directory usages before they are gone in the framework repo.


pablintino:
Duplication of https://github.com/openstack-k8s-operators/nova-operator/pull/637

PR #645, opened 2024-01-10
update container images for fips

SeanMooney:
This change modifes the builder image use the fully qualified name
but maintains the current version of go 1.19

A new GO_BUILD_EXTRA_ENV_ARGS build argument is added to allow
the env parmaters set during the go build step to be
modifed when building the image.

it defaults to "CGO_ENABLED=0 GO111MODULE=on"

we use CGO_ENABLED=0 by default so that the manager binary
is staticly linked. this allows us to use the
gcr.io/distroless/static:nonroot image as our base image.

prow jobs will be added to test building with
CGO_ENABLED=1 and a fips enabled toolchain

The makefile is extened to support passing DOCKER_BUILD_ARGS
to the build command executed by docker-build

Closes: [OSPRH-3106](https://issues.redhat.com//browse/OSPRH-3106)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG GOLANG_BUILDER=golang:1.19
+ARG GOLANG_BUILDER=quay.io/projectquay/golang:1.19
 ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 # Build the manager binary
@@ -13,10 +13,11 @@ ARG REMOTE_SOURCE_SUBDIR=
 ARG DEST_ROOT=/dest-root
 
 ARG GO_BUILD_EXTRA_ARGS=
-
+# note we set CGO_ENABLED=0 to force a static build so that we can use
+# distroless/static as our base image
+ARG GO_BUILD_EXTRA_ENV_ARGS="CGO_ENABLED=0 GO111MODULE=on"
 COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR
 WORKDIR $REMOTE_SOURCE_DIR/$REMOTE_SOURCE_SUBDIR
-
 RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 
 # cache deps before building and copying source so that we don't need to re-download as much
@@ -24,7 +25,8 @@ RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 RUN if [ ! -f $CACHITO_ENV_FILE ]; then go mod download ; fi
 
 # Build manager
-RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; CGO_ENABLED=0  GO111MODULE=on go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
+RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; env ${GO_BUILD_EXTRA_ENV_ARGS} go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
+
 
 RUN cp -r templates ${DEST_ROOT}/templates
 
--- a/Makefile
+++ b/Makefile
@@ -160,9 +160,13 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	/bin/bash hack/clean_local_webhook.sh
 	go run ./main.go -metrics-bind-address ":$(METRICS_PORT)" -health-probe-bind-address ":$(HEALTH_PORT)"
 
+
+# Extra vars which will be passed to the Docker-build
+DOCKER_BUILD_ARGS ?=
+
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	podman build -t ${IMG} .
+	podman build -t ${IMG} . ${DOCKER_BUILD_ARGS}
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can use [openshift-local](https://developers.redhat.com/products/openshift-l
 
 ### golang
 
-this repo currently uses go 1.18
+This repo currently uses go 1.19+
 
 ### pre-commit
 


Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG GOLANG_BUILDER=golang:1.19
-ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
+ARG GOLANG_BUILDER=quay.io/projectquay/golang:1.19
+ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/cc-debian12:nonroot
```
SeanMooney, line 2: the chagne form static to cc-debian12 will provide openssl to make the tool happy but we should not need this.

we are executing a go binary which is always statically linked to all libs and we never spawn a shell to invoke the openssl binaies.

so rather then changing the base image we likely should fix the tool.

SeanMooney: actully we might have to use gcr.io/distroless/base-debian11 i need to confirm if cc-debian12 has openssl.
i would strongly prefer to keep using static here since fips is only relevant for downstream images and this is a argument that can be overridden when the contiener is built so ill likely drop this if we can reach agreement that this is not required for the quay images.

it is more secure to use static then the other variants as we don't need the extra packages the other variants intoduce.

Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG GOLANG_BUILDER=golang:1.19
-ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
+ARG GOLANG_BUILDER=registry.access.redhat.com/ubi9/go-toolset:1.20
+ARG OPERATOR_BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-minimal:latest
```
SeanMooney, line 2: before proceeding with thse base image chnages we should discuss if
the trade off in image size is worth it.

this will increase the final image form 25.1MB to 155 MB
gcr.io/distroless/static                              nonroot           01b5d1d2ca0f  20 hours ago    3.87 MB
registry.access.redhat.com/ubi9/ubi-minimal           latest            1cd5a3561bdb  5 weeks ago     97.1 MB
quay.io/openstack-k8s-operators/nova-operator         latest            94090de25043  50 minutes ago      155 MB

SeanMooney:
runinig payload check on thisi get


```
---- Failure Report
+-------------------------+--------------------+--------------------------------------+--------------------------------------+
| OPERATOR NAME           | EXECUTABLE NAME    | STATUS                               | IMAGE                                |
+-------------------------+--------------------+--------------------------------------+--------------------------------------+
| nova-operator-container | /usr/sbin/ldconfig | executable is not dynamically linked | quay.io/smooney/nova-operator:latest |
+-------------------------+--------------------+--------------------------------------+--------------------------------------+
---- Warning Report
+-------------------------+-----------------+-----------------------------------------------------------------+--------------------------------------+
| OPERATOR NAME           | EXECUTABLE NAME | STATUS                                                          | IMAGE                                |
+-------------------------+-----------------+-----------------------------------------------------------------+--------------------------------------+
| nova-operator-container | /manager        | go binary has no build tags set (should have strictfipsruntime) | quay.io/smooney/nova-operator:latest |
+-------------------------+-----------------+-----------------------------------------------------------------+--------------------------------------+
```

https://termbin.com/wr7i

gibizer:
Is there a CI job running the tool on all build?

SeanMooney:
to fully verfy this work we will need to modify the prow image building job to add the fips check

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG_BUILDER=golang:1.19
+ARG GOLANG_BUILDER=quay.io/projectquay/golang:1.20
```
gibizer, line 1: can we bump to golang 1.20 now?

Dockerfile
```diff
@@ -13,18 +13,18 @@ ARG REMOTE_SOURCE_SUBDIR=
 ARG DEST_ROOT=/dest-root
 
 ARG GO_BUILD_EXTRA_ARGS=
-
+ARG GO_BUILD_EXTRA_ENV_ARGS="CGO_ENABLED=1 GO111MODULE=on"
```
gibizer, line 16: can we simply reuse the existing `GO_BUILD_EXTRA_ARGS`

Dockerfile
```diff
@@ -13,18 +13,18 @@ ARG REMOTE_SOURCE_SUBDIR=
...
 RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN if [ ! -f $CACHITO_ENV_FILE ]; then go mod download ; fi
 
 # Build manager
-RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; CGO_ENABLED=0  GO111MODULE=on go build ${GO_BUILD_EXTRA_ARGS} -a -o ${DEST_ROOT}/manager main.go
+RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi ; env ${GO_BUILD_EXTRA_ENV_ARGS} go build -a -o ${DEST_ROOT}/manager main.go
+
```
SeanMooney, line 27: oh i forgot to re add  ${GO_BUILD_EXTRA_ARGS}

Dockerfile
SeanMooney: no there is a postioanl depency
i forgot to readd ${GO_BUILD_EXTRA_ENV_ARGS}

Dockerfile
SeanMooney: i have been using 1.21 locally for quite a while so i belive so

SeanMooney: lets chat about that seperatly

i was doing both in this pr because i thought the bump to golang 1.20 was overdue and required for some of the renovate changes and the new runtime version ectra
i could pull the bump out to a separate pr

SeanMooney: we agreed to do the go bump in a diffent pr so ill adress that shortly

gibizer (approved):

SeanMooney:
https://review.rdoproject.org/zuul/build/7890caff097e4d948edeb543c35a162b

the kuttle job passed.
this version just updates the commit message but is otherwise its identical

PR #650, opened 2024-01-15
Revert "Enable CPU power management by default for libvirt compute"

gibizer:
This reverts commit 3d19fcc042878f923f745f9a9addec07a9dabd6f. As there are multiple errors when cpu_dedicate_set is defined.


SeanMooney (approved):
approving until we have time to debug this in detail

Issue #652, opened 2024-01-18
Validate transport_url and check if every cell have separate one


mrkisaolamb:
Probably we need to validate it in webhook or in nova reconciler

gibizer:
I think we can validate that every CellTemplate in NovaSpec.cellTemplates has a unique CellTemplate.CellMessageBusInstance field value

PR #653, opened 2024-01-18
Validate cellMessageBusInstance to be uniq per cell

mrkisaolamb:
Closes: #652

### DIFF
--- a/api/v1beta1/nova_webhook.go
+++ b/api/v1beta1/nova_webhook.go
@@ -151,13 +151,28 @@ func (r *NovaSpec) ValidateCellTemplates(basePath *field.Path) field.ErrorList {
 		)
 	}
 
+	cellMessageBusNames := make(map[string]string)
+
 	for name, cell := range r.CellTemplates {
 		cellPath := basePath.Child("cellTemplates").Key(name)
 		errors = append(
 			errors,
 			ValidateCellName(cellPath, name)...,
 		)
+		if name != Cell0Name {
+			if dupName, ok := cellMessageBusNames[cell.CellMessageBusInstance]; ok {
+				errors = append(errors, field.Invalid(
+					cellPath.Child("cellMessageBusInstance"),
+					cell.CellMessageBusInstance,
+					fmt.Sprintf(
+						"RabbitMqCluster CR need to be uniq per cell. It's duplicated with cell: %s",
+						dupName),
+				),
+				)
+			}
 
+			cellMessageBusNames[cell.CellMessageBusInstance] = name
+		}
 		if *cell.MetadataServiceTemplate.Enabled && *r.MetadataServiceTemplate.Enabled {
 			errors = append(
 				errors,
--- a/test/functional/validation_webhook_test.go
+++ b/test/functional/validation_webhook_test.go
@@ -567,4 +567,36 @@ var _ = Describe("Nova validation", func() {
 			),
 		)
 	})
+	It("check Cell validation with duplicate cellMessageBusInstance", func() {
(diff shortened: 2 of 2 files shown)


api/v1beta1/nova_webhook.go
```diff
@@ -151,13 +151,25 @@ func (r *NovaSpec) ValidateCellTemplates(basePath *field.Path) field.ErrorList {
...
 
+	cellMessageBusNames := make(map[string]int)
+
 	for name, cell := range r.CellTemplates {
 		cellPath := basePath.Child("cellTemplates").Key(name)
 		errors = append(
 			errors,
 			ValidateCellName(cellPath, name)...,
 		)
+		if name != Cell0Name {
```
gibizer, line 162: The message bus of cell0 also need to be unique across cells so you have to check cell0 too

api/v1beta1/nova_webhook.go
```diff
@@ -151,13 +151,25 @@ func (r *NovaSpec) ValidateCellTemplates(basePath *field.Path) field.ErrorList {
...
+			if _, ok := cellMessageBusNames[cell.CellMessageBusInstance]; ok {
+				errors = append(errors, field.Invalid(
+					cellPath.Child("cellMessageBusInstance"),
+					cell.CellMessageBusInstance,
+					"RabbitMqCluster CR need to be uniq per cell"),
```
gibizer, line 167: If you want you can be a bit more precise in the error message and mention the two cell name that uses the same message bus. For that you need to change the cellMessageBusNames to be map[string]string and store the cell name in the value. Then here `cellMessageBusNames[cell.CellMessageBusInstance]` gives you one of the wrongly configured cell and `name` gives you the name of the other wrongly configured cell.

test/functional/validation_webhook_test.go
```diff
@@ -567,4 +567,36 @@ var _ = Describe("Nova validation", func() {
 			),
 		)
 	})
+	It("check Cell validation with duplicate cellMessageBusInstance", func() {
+		spec := GetDefaultNovaSpec()
+		cell0 := GetDefaultNovaCellTemplate()
+		cell1 := GetDefaultNovaCellTemplate()
+		cell2 := GetDefaultNovaCellTemplate()
+		cell1["cellMessageBusInstance"] = "rabbitmq-of-caerbannog"
```
gibizer, line 575: +1 
:rabbit: 
:hole:

api/v1beta1/nova_webhook.go
mrkisaolamb: For cell 0 we are not using cellMessageBusInstance, and cell0 doesn't have separate MQ https://github.com/openstack-k8s-operators/nova-operator/blob/main/controllers/nova_controller.go#L354

api/v1beta1/nova_webhook.go
mrkisaolamb: good point, done

api/v1beta1/nova_webhook.go
gibizer: Sorry you are right. I just looked at a deployed Nova CR and saw that the field is filled. But you are correct, whathever the user passed in that field to Nova CR it gets ignored as noted in https://github.com/openstack-k8s-operators/nova-operator/blob/8f204945b887a562229143f40901c4540925e9cd/api/v1beta1/novacell_types.go#L45-L50

SeanMooney (approved):
looks good to me also

i was going to say you should use a set not a map

and then use cell.name for the errmessage but the memory overhead is trivial so meh this is fine too

PR #663, opened 2024-02-01
Drop Debug fields

gibizer:
These fields were redundant. Crashing pods can be debugged via oc debug.

PreserveJobs field moved to the NovaSpec level from the Debug struct to
align with most of our service operators. We keep this sort of debug
option as this cannot be simulated with oc commands at the moment.

Implements: [OSPRH-4290](https://issues.redhat.com//browse/OSPRH-4290)


mrkisaolamb (approved):
lgtm

PR #672, opened 2024-02-06
Fix operator crash at compute removal

gibizer:
When a compute definition is removed from the novaComputeTemplates the nova-operator crashed due to a wrong logging statement. This is fixed now and the missing test coverage is added.


mrkisaolamb (approved):
lgtm

PR #675, opened 2024-02-07
[webhook]Fix compute template rejection messages

gibizer:
Some of the validation code used wrong yaml path when referring to the invalid value in the rejection message. This is fixed now and a missing test case is added to cover ironic replicas.


mrkisaolamb (approved):
lgtm

PR #671, opened 2024-02-06
Controller runtime update

bshephar:
This changes updates the controller-runtime and fixes the initialisation of objects to comply with the new controller-runtime standards. Namely:

We fix the function signatures of the webhook to include admissions.Warnings:
https://github.com/kubernetes-sigs/controller-runtime/commit/8770b4d3b5425b156f48c971ee140ea25f9c1bfa

We fix the initialisation of the manager to declare the webhook following the new schema:
https://github.com/kubernetes-sigs/controller-runtime/commit/91cdd8c9539c4860c5bacad36ebf75e4271f726d

We fix the Watches function to comply with the new signature definition:
https://github.com/kubernetes-sigs/controller-runtime/commit/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2

Add context to EventHandlers:
https://github.com/kubernetes-sigs/controller-runtime/commit/2464a9d7af4911c6bf6e0baaaf8dc0ab08c7452c

PR #676, opened 2024-02-07
DNM: testing #671 with temporary replace lines

gibizer:
Just tesing https://github.com/openstack-k8s-operators/nova-operator/pull/671


gibizer:
The env test passed for me locally with this PR so I think we are OK but we need a rebase on #671 to see the same result here.

PR #678, opened 2024-02-09
Refactor test only CreateCellInternalSecret

mrkisaolamb:
Not all services need same internal secret
now in tests we can create minimal secret

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	. "github.com/onsi/gomega"
+	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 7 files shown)


test/functional/base_test.go
```diff
@@ -877,16 +878,19 @@ func GetDefaultNovaNoVNCProxySpec(cell CellNames) map[string]interface{} {
 	}
 }
 
-func CreateCellInternalSecret(cell CellNames) *corev1.Secret {
+func CreateCellInternalSecret(cell CellNames, additionalValues map[string][]byte) *corev1.Secret {
```
gibizer, line 881: Based on the current usage I would add two helpers to avoid repeating the metadata secret.

CreateCellInternalSecretWithExtra(cell CellName, additionalValues)

CreateCellInternalSecret(cell CellName)
CreateCellInternalSecretWithMetadata(cell CellName)

// later
Or will the memcached change add more, different usage of the additionalValues ?

mrkisaolamb: I change it because same thing will be with memcached instances when metadata don't need this in secret

gibizer (approved):

PR #681, opened 2024-02-12
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount

zzzeek:
This moves nova to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

https://github.com/openstack-k8s-operators/mariadb-operator/pull/184/files

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] Use configsecrets for database URLs; remove from job hash  - already was like that
15. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed


zzzeek:
recheck

controllers/nova_controller.go
```diff
@@ -196,18 +196,29 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.APIDatabaseAccount,
+		instance.Namespace, false,
+	)
+
+	if err != nil {
+		return ctrl.Result{}, err
```
zzzeek, line 211: we can add a status condition here, see https://github.com/openstack-k8s-operators/neutron-operator/blame/13c5c990d917ab2825bda8e44b6f631474cc9fe1/controllers/neutronapi_controller.go#L925

zzzeek (commented):
this also needs GetDatabaseByName updated for the finalizer removal

controllers/nova_controller.go
```diff
@@ -817,16 +852,30 @@ func (r *NovaReconciler) ensureCellDB(
...
+
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err := mariadbv1.EnsureMariaDBAccount(
+		ctx, h, cellTemplate.CellDatabaseAccount,
+		instance.Namespace, false, "nova_"+cellName,
```
gibizer, line 864: we have a helper for that getCellDatabaseName

controllers/nova_controller.go
```diff
@@ -817,16 +852,30 @@ func (r *NovaReconciler) ensureCellDB(
...
 	)
+
+	if err != nil {
+		return nil, nova.DBFailed, err
+	}
+
+	cellDB := mariadbv1.NewDatabaseForAccount(
+		cellTemplate.CellDatabaseInstance, // mariadb/galera service to target
+		"nova_"+cellName,                  // name used in CREATE DATABASE in mariadb
```
gibizer, line 873: we have a helper for that getCellDatabaseName.

controllers/nova_controller.go
```diff
@@ -880,7 +929,7 @@ func (r *NovaReconciler) ensureCell(
 	}
 	if cellTemplate.HasAPIAccess {
 		cellSpec.APIDatabaseHostname = apiDB.GetDatabaseHostname()
-		cellSpec.APIDatabaseUser = instance.Spec.APIDatabaseUser
+		cellSpec.APIDatabaseAccount = instance.Spec.APIDatabaseAccount
```
gibizer, line 932: OK we still use the hostname to signal if the api DB is accessible to the cell or not. Not super nice that we default the account in both case but that is not new, we did that with the APIDatabaseUser too. So I think it is good as is.

test/functional/nova_controller_test.go
```diff
@@ -989,6 +1015,217 @@ var _ = Describe("Nova controller", func() {
...
+			}, timeout, interval).Should(Succeed())
+
+		},
+		// delete the CR, allowing tests that exercise finalizer removal
+		DeleteCR: func() {
+			th.DeleteInstance(GetNova(novaNames.NovaName))
+		},
+	}
+
+	mariadbCellSuite.RunBasicSuite()
```
gibizer, line 1228: nice. Thanks for the generalized test cases

test/functional/nova_multicell_test.go
```diff
@@ -32,6 +33,28 @@ import (
...
+			cell1.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+		DeferCleanup(k8sClient.Delete, ctx, cell1Account)
+		DeferCleanup(k8sClient.Delete, ctx, cell1Secret)
+
+		cell2Account, cell2Secret := mariadb.CreateMariaDBAccountAndSecret(
+			cell2.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+		DeferCleanup(k8sClient.Delete, ctx, cell2Account)
+		DeferCleanup(k8sClient.Delete, ctx, cell2Secret)
+
+	})
```
gibizer, line 57: Is it intentional that we pre-create the accounts in these tests? As we create Nova CR and not some subCR in this tests we could let the nova controller's ensure call do the creation. Of course it is valid to pre-create them to simulate that it is done by the openstack-operator or the user.

If this is intentional then please leave a comment in the code nothing it as this is now different from the nova_controller_test.go and that might confuse the future reader

test/functional/nova_reconfiguration_test.go
```diff
@@ -50,22 +51,42 @@ func CreateNovaWith3CellsAndEnsureReady(novaNames NovaNames) {
 	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell1.MariaDBDatabaseName.Namespace, cell1.MariaDBDatabaseName.Name, serviceSpec))
 	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell2.MariaDBDatabaseName.Namespace, cell2.MariaDBDatabaseName.Name, serviceSpec))
 
+	apiMariaDBAccount, apiMariaDBSecret := mariadb.CreateMariaDBAccountAndSecret(
```
gibizer, line 54: ditto, we could let the nova controller create these. If this is intentional then please leave a comment

gibizer (commented):
I think all my comments can be fixed in a follow up. So let me know if you want to do that and I will approve this.

test/functional/nova_multicell_test.go
zzzeek: if we move to openstack-operator being responsible for mariadbaccount creation then we'd be glad to have these here.   I have a feeling that EnsureMariaDBAccount is going to be generating username/passwords for a long time though because it's very convenient that it takes care of itself.    

given the discussion at https://github.com/openstack-k8s-operators/mariadb-operator/pull/205#pullrequestreview-1916137443 I think I want to add Eventually() to the RunURLAssertSuites here before we merge and I can try removing some of these extraneous account creates.    Between the many test suites here as well as all the changes made by @stuggi there are likely more mariadbaccoutn creates than needed.

controllers/nova_controller.go
```diff
@@ -683,14 +714,24 @@ func (r *NovaReconciler) ensureNovaManageJobSecret(
...
+	apiDatabaseAccount, apiDbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, instance.Spec.APIDatabaseAccount, instance.Namespace)
+	if err != nil {
+		return nil, "", "", err
+	}
+
+	cellDatabaseAccount, cellDbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, cell.Spec.CellDatabaseAccount, instance.Namespace)
+	if err != nil {
+		return nil, "", "", err
+	}
+
```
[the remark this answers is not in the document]
zzzeek: yes these should be calling from db.GetAccount()  / db.GetSecret()

test/functional/nova_multicell_test.go
zzzeek: update: I am not using any RunURLAssertSuites.   so we can merge this as is and address things in followup, given how much effort it is for me to keep rebasing this each time someone changes something.

gibizer (approved):
I deployed this and did a cell1 password rotation. Then I checked that the new user / pass was added to the cell mapping table automatically and I could deploy an instance to celll1 too.

the open comments can be fixed in a followup

PR #694, opened 2024-02-18
Remove go version from README.md

gibizer:
The supported golang version can be seen in the go.mod file easily so duplicating that information in the readme just calls for divergence. (As it did during the golang 1.20 bump)


mrkisaolamb (approved):
lgtm

PR #698, opened 2024-02-20
Add DB purge CronJob

gibizer:
We will have one CronJob per cell to do db archiving and purging per
cell.

Pointer types are needed for nested struct field defaulting.

Operator-lint update is needed to handle the pointer types properly in
the omitempty check and to handle an edge case with generic types.

Implements: [OSPRH-104](https://issues.redhat.com//browse/OSPRH-104)


gibizer:
kuttl failure seems to be unrelated but probably persistent
```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'cifmw_edpm_prepare_update_os_containers | bool' failed. The error was: error while evaluating conditional (cifmw_edpm_prepare_update_os_containers | bool): 'cifmw_edpm_prepare_update_os_containers' is undefined. 'cifmw_edpm_prepare_update_os_containers' is undefined\n\nThe error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/nova-operator/ci/nova-operator-base/playbooks/pre.yaml': line 128, column 20, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n              --for=jsonpath='{.status.phase}'=Complete --timeout=20m\n                   ^ here\n\nThere appears to be both 'k=v' shorthand syntax and YAML in this task. Only one syntax may be used.\n"}
```

templates/novaconductor/bin/dbpurge.sh
```diff
@@ -0,0 +1,38 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -x
+export ARCHIVE_AGE=${ARCHIVE_AGE:?"Please specify ARCHIVE_AGE variable."}
+export PURGE_AGE=${PURGE_AGE:?"Please specify PURGE_AGE variable."}
+archive_before=$(date --date="${ARCHIVE_AGE} day ago" +%Y-%m-%d)
+purge_before=$(date --date="${PURGE_AGE} day ago" +%Y-%m-%d)
```
SeanMooney, line 20: i guess this works.

we could also do the datetime conversion in go and pass in the iso datetime
since the locale should be set in these images "day ago" should not break on non English systems as we will be usein an en_US or C locale for this container regardless of the underlying hosts locale

test/functional/base_test.go
```diff
@@ -338,6 +339,9 @@ func CreateNovaWithCell0(name types.NamespacedName) client.Object {
 				"cell0": map[string]interface{}{
 					"cellDatabaseUser": "nova_cell0",
 					"hasAPIAccess":     true,
+					"dbPurge": map[string]interface{}{
+						"schedule": "1 0 * * *",
```
SeanMooney, line 343: so the actual default will be 0 0 * * *

and your using a non default default to test that that works?

test/functional/nova_controller_test.go
```diff
@@ -135,6 +136,14 @@ var _ = Describe("Nova controller", func() {
 			Expect(instance.Status.RegisteredCells).To(BeEmpty())
 		})
 
+		It("defaults Spec fields", func() {
```
SeanMooney, line 139: this is not really default its the test defaults

test/functional/novaconductor_controller_test.go
```diff
@@ -58,6 +59,13 @@ var _ = Describe("NovaConductor controller", func() {
 			Expect(instance.Status.ReadyCount).To(Equal(int32(0)))
 		})
 
+		It("defaults Spec fields", func() {
+			instance := GetNovaConductor(cell0.ConductorName)
+			Expect(instance.Spec.DBPurge.Schedule).To(Equal(ptr.To("0 0 * * *")))
```
SeanMooney, line 64: this is the actual defalt

pkg/novaconductor/dbpurge.go
```diff
@@ -0,0 +1,93 @@
...
+
+func DBPurgeCronJob(
+	instance *novav1.NovaConductor,
+	labels map[string]string,
+	annotations map[string]string,
+) *batchv1.CronJob {
+	args := []string{"-c", nova.KollaServiceCommand}
+
+	envVars := map[string]env.Setter{}
+	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
```
SeanMooney, line 24: this is a point of divergance form kolla

they removed this since COPY_ALWAYS was the default and no one used
COPY_ONCE so we probably should do that eventually.

this is not wrong its just unnecessary for us to support anything else in tcib.

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/set_configs.py#L371-L386 can be simplified

templates/novaconductor/bin/dbpurge.sh
```diff
@@ -0,0 +1,38 @@
...
+
+nova-manage db purge --verbose --before "${purge_before}"
+ret=$?
+# 0 means no error and something is deleted
+# 3 means no error and nothing is deleted
+if [[ $ret -eq 1 || $ret -eq 2 || $ret -gt 3 ]]; then
+    exit $ret
+fi
+
+exit 0
```
SeanMooney, line 38: ack so one job for both and purge only runs if archive does not error.
i guess that is fine too.

i wonder if there is a native way to get notified if this cron job fails in k8s? i.e. how will a human operator discover that they need to investigate why the archive filed and correct it.

that's out of scope of this change but we likely need to document that somewhere

SeanMooney (commented):
overall i think this looks good. im a little confused why some of the enve tests are suing "1 0 * * *" instead of "0 0 * * *"

templates/novaconductor/bin/dbpurge.sh
gibizer: at last it works for me, and I don't see less trouble with timezones even if we convert it in go 

```
Running command: '/bin/dbpurge.sh'
+ umask 0022
+ exec /bin/dbpurge.sh
+ export ARCHIVE_AGE=1
+ ARCHIVE_AGE=1
+ export PURGE_AGE=1
+ PURGE_AGE=1
++ date '--date=1 day ago' +%Y-%m-%d
+ archive_before=2024-02-25
++ date '--date=1 day ago' +%Y-%m-%d
+ purge_before=2024-02-25
+ nova-manage db archive_deleted_rows --verbose --until-complete --task-log --before 2024-02-25
Archiving...complete
Nothing was archived.
+ ret=0
+ '[' 0 -gt 1 ']'
+ nova-manage db purge --verbose --before 2024-02-25
DB: Unable to purge table shadow_alembic_version because it has no timestamp column
+ ret=3
+ [[ 3 -eq 1 ]]
+ [[ 3 -eq 2 ]]
+ [[ 3 -gt 3 ]]
+ exit 0
```

test/functional/base_test.go
gibizer: testing that propagating value works through the CRDs

templates/novaconductor/bin/dbpurge.sh
gibizer: I'm not sure but I guess a k8s monitoring tool can watch and report cronjob health. I added a sentence about it in the Jira, so the doc update is not forgotten

test/functional/nova_controller_test.go
gibizer: I can drop the assert of the schedule filed as that is not the default, but the two age fields are defaults here

templates/novaconductor/bin/dbpurge.sh
SeanMooney: we have set the locale in the container image so yes it should work fine.
I was just wondering if we needed to worry about that or not as the time zone of the db and the container image might be different but lets keep this as is.

gibizer: if the timezone of the DB is different from ocp then neither go nor bash will help, then we need a parameter form MariadbDatabase, to know what tz to use

gibizer:
> the funcitonal test failure is
> 
> not to contain substring : novncproxy_base_url In [It] at: /go/src/github.com/openstack-k8s-operators/operator/test/functional/novacell_controller_test.go:405 @ 02/25/24 20:49:23.25
> 
> flaky test?

probably yes. but the kuttl job is broken on main

gibizer:
recheck

gibizer:
kuttl test failure is relevant, probably we have a different name in kuttl
```
2024-02-26 13:34:48.219366 | controller |     case.go:364: failed in step 1-deploy
2024-02-26 13:34:48.219432 | controller |     case.go:366: cronjobs.batch "nova-cell0-conductor-db-purge" not found
2024-02-26 13:34:48.219464 | controller |     case.go:366: cronjobs.batch "nova-cell1-conductor-db-purge" not found
```

gibizer:
rebased after memcache support landed

gibizer:
kuttl is finally green \o/

pkg/novaconductor/dbpurge.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// add CA cert if defined
+	if instance.Spec.TLS.CaBundleSecretName != "" {
+		volumes = append(volumes, instance.Spec.TLS.CreateVolume())
+		volumeMounts = append(volumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
+	}
+
+	cron := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      instance.Name + "-db-purge",
```
mrkisaolamb, line 50: because this is job per-cell shoudn't we have name created based on cell name?

gibizer: instance is a conductor, so its name already cell name based, and unique across cells. E.g. a job for cell0 will be named `nova-kuttl-cell0-conductor-db-purge`. Do you suggest to drop the conductor part from the name?

https://github.com/openstack-k8s-operators/nova-operator/pull/698/files/6990025cbdeba2f3895ca70e20a905c4ac31715e#diff-10280178933dbd65fee3107932d754f291ac6cb1fe1a952c043e982d4be5232eR521

api/v1beta1/novacell_types.go
```diff
@@ -188,6 +192,34 @@ type NovaCellSpec struct {
...
+	// ArchiveAge defines the minimuma age of the records in days that can be
+	// moved to the shadow tables.
+	ArchiveAge *int `json:"archiveAge"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=90
+	// +kubebuilder:validation:Minimum=1
+	// PurgeAge defines the minimum age of the records in days that can be
+	// deleted from the shadow tables
+	PurgeAge *int `json:"purgeAge"`
```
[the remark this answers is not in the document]
gibizer: I'm reluctant to open up the interface that much. The max rows and sleep params are helpful to limit the db load during archiving high volume of data. However that can be solved by doing frequent archiving of small amount of data via the existing params. And we actually suggest to do that frequent archiving.

For the extreme cases we cannot predict today, I would rather let the customer file an RFE for us about the case, so we can learn about it. I predict in many of those cases the solution will not be a tweak of the unexposed params, but a change in other parts of the system. Still if it turns out, that I'm wrong and these params truly needed, such RFE will be a good learning about something that is unknown today. Also for that hypothetical RFE we can provide a CronJob yaml def as a WA until the API change is delivered.

pkg/novaconductor/dbpurge.go
mrkisaolamb: yes, to use only instance.Spec.CellName to shorten name plus IMO it's more accurate

api/v1beta1/novacell_types.go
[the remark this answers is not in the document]
mrkisaolamb: My thinking was more inclined towards creating a backdoor for a quick fix within the system, even in the future when we may have more options. However, the assumption is that we wouldn't recommend this approach to the client. But I like being opinionated because it provides more predictability from a developer's point of view. So, even if we don't open up for this workaround entirely, perhaps allowing at least the configuration of max rows and sleep at this point could address some issues. But, generally, I'm okay with the current approach.

pkg/novaconductor/dbpurge.go
gibizer: Done

mrkisaolamb (approved):

PR #702, opened 2024-02-22
Revert "Reapply "Enable CPU power management by default for libvirt compute""

gibizer:
This reverts commit 41497aaed81e537bce2e9f62d7175bc593177b9a.


SeanMooney (approved):
we found 2 new bugs in the power managment feature when trying to enable whitebox testing.

one is related to live migation and the other to the isolate emulator threads policy.

as a result we are goign to hold off on enabling htis by default until we can fix that

PR #708, opened 2024-02-26
remove unused zuul jobs and fix kuttle

SeanMooney:
the nova base job was calling set_openstack_containers
based on the edpm_prepare job to enable the
nova-operator-compute-kit job to eventurally deploy edpm nodes.

the nova-operator-compute-kit is not currenly used since added a
diffent tempet job so both the nova-operator-compute-kit job and
invocation of set_openstack_containers in nova-operator-base is
removed.


gibizer (approved):
simplification is welcome

PR #710, opened 2024-02-27
[envtest]remove unnecessary and racy db creation

gibizer:
The c1c816062d66d1c8a8af8f6b9ade671d37562d05 added MariaDBDatabase and
Account creation to the reconfiguration tests that uses Nova CRD. The
nova controller creates these resources automatically so the test
should not do it. This causes a race where the test case tries to create
and already existing db, when the conroller creates it first.


mrkisaolamb (approved):
lgtm

SeanMooney (approved):

PR #712, opened 2024-02-28
[envtest]fix metadata shared secret reconfig test

gibizer:
With unfortunate timing the shared secret might be already up to date
with the new password but the Metadata.Status.Hash is not yet showing
the new hash of the secret. This causes the test to fail as it only
waits for the secret to be updated and assumes that the hash is updated
too.

The test instability can be reproduced with sleep:
```diff
diff --git a/controllers/novametadata_controller.go b/controllers/novametadata_controller.go
index dc3ddfa9..3e6e8630 100644
--- a/controllers/novametadata_controller.go
+++ b/controllers/novametadata_controller.go
@@ -21,6 +21,7 @@ import (
        "fmt"
        "net/url"
        "strings"
+       "time"

        v1 "k8s.io/api/apps/v1"
        corev1 "k8s.io/api/core/v1"
@@ -792,6 +793,7 @@ func (r *NovaMetadataReconciler) generateNeutronConfigs(
                return err
        }

+       time.Sleep(2 * time.Second)
        // TODO(gibi): can we make it simpler?
        a := &corev1.EnvVar{}
        hashes[configName](a)
```


mrkisaolamb (approved):
lgtm

PR #722, opened 2024-03-12
[CRD]Move container images to top level

gibizer:
To simplify service container image defaulting and upgrade this patch removes the container image fields from the service templates and the cell template. Instead it adds container image URL fields to the NovaSpec and the NovaCellSpec. The former is expected to be set by the openstack-operator the latter is set by the nova reconciler. We keep the service Spec container image fields and they are now populated from the NovaSpec or NovaCellSpec respectively.

This limits the ability to define different container image for the same service type in different cells. If such capability is needed, e.g.: for debugging, then oc debug can be used. More permanent hotfix on an individual cell level will not be possible, the fix needs to be deployed to every cell in the controlplane.


api/v1beta1/common_types.go
```diff
@@ -89,18 +89,82 @@ type PasswordSelector struct {
...
 
+type NovaImages struct {
+	// +kubebuilder:validation:Required
+	// APIContainerImageURL
+	APIContainerImageURL string `json:"apiContainerImageURL"`
+
+	// +kubebuilder:validation:Required
+	// SchedulerContainerImageURL
+	SchedulerContainerImageURL string `json:"schedulerContainerImageURL"`
+
```
mrkisaolamb, line 100: Shoudn't we have here also metadata image for case of global metadata?

gibizer: as it includes the NovaCellImages struct it has metadata image url as well

SeanMooney: thanks for working on this

and ya since you are doing an anonomus embed of the NovaCellImages
struct the metadata image will be directly in NovaImages automatically

we can see this here https://github.com/openstack-k8s-operators/nova-operator/pull/722/files#diff-faae6079d2ee9fc600482b30c93a7791eb19189fcbc8031abde8258e6337e58aR1662

SeanMooney (commented):
this look quite promising to me

gibizer:
kuttl test failure is relevant.
```
    case.go:366: resource Nova:nova-kuttl-default/nova-kuttl: .spec.apiServiceTemplate.containerImage: key is missing from map
```

gibizer:
> kuttl test failure is relevant.
> 
> ```
>     case.go:366: resource Nova:nova-kuttl-default/nova-kuttl: .spec.apiServiceTemplate.containerImage: key is missing from map
> ```

Fixed

mrkisaolamb (approved):
lgtm

PR #723, opened 2024-03-12
Add NovaSpecCore struct 

gibizer:
This version of the struct (called "core") is meant to
be used via the OpenStackControlplane. It is the same
as NovaSpec only it is missing the container image URL fields.

Also we expose Default and Validate functions on NovaSpecCore.

The higher level meta operator will include the NovaSpecCore now in the
OpenStackControlPlane CRD and therefore it needs to default and validate
that struct in its webhooks, so the necessary functions are exposed.
This patch manages to avoid duplicating the validation logic because the
the NovaSpecCore is now directly embedded in the NovaSpec and therefore
validations written for NovaSpecCore can be also called on NovaSpec
too.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


SeanMooney (approved):
assuming ci agrees that this is correct, this is much more self contained and has little  long term maintance overhead compared to the iteration before we uplevled the images.

i am much happier with this iteration +2

PR #725, opened 2024-03-13
update mariadb to include finalizer secret logic

zzzeek:
this PR updates mariadb-operator to include [1], which will add / remove finalizers on the Secret objects that are associated with MariaDBAccount objects.  This ensures that GetDatabaseByNameAndAccount will successfully locate a MariaDBDatabase/MariaDBAccount/Secret trio, rather than returning not found, allowing ``Database.DeleteFinalizer()`` to be called when a resource is in reconcile delete.

[1] https://github.com/openstack-k8s-operators/mariadb-operator/pull/210


gibizer:
#719 will pull in this mariadb version among other bumps

zzzeek:
cool, i just wanted to make sure tests pass

PR #733, opened 2024-03-27
Add Status.ObservedGeneration to Nova CR

dprince:
Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)


dprince:
recheck

gibizer (approved):
Looks good. We agree that we will use the new observedGeneration + Ready condition together in the openstack-operator to decide when an update on Nova is fully reconciled. 

The current implementation of when to set the observedGeneration is a good first approximation(i.e. it is not worst than only having the ReayCondition), and we will improve on it later internally.

PR #738, opened 2024-04-15
use shared content provider job

SeanMooney:
This change replaces the nova-operator-content-provider
with the openstack-k8s-operators-content-provider.

The openstack-k8s-operators-content-provider has all the
same functionality of the nova-operator-content-provider
plus it support depends on to other operators i.e.
the neuton-operator which was not supported by the
nova-operator-content-provider

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,20 +1,5 @@
 ---
 - job:
-    name: nova-operator-content-provider
-    parent: content-provider-base
-    description: |
-      This job builds the nova-operator and provides it to child jobs
-      It also rebuilds the openstack-operator using the updated nova operator.
-    provides:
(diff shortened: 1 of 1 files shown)


gibizer (approved):
thanks

SeanMooney:
the tempest job failed ot install cert manager...
ill check with the ci folks after that reports before recheckign
i know this was a problem in the past so i don't want to waste ci time with a recheck if there is a known issue currently 

we can review the logs and confirm the inventory is correct.
the kuttle test seams to be working so we can also confirm the correct image form the content provider was used there.

SeanMooney:
check-rdo 6b896a47cbb1ef7aab7ee504af1e372a59d845fd has been reverted

PR #739, opened 2024-04-17
Add ObservedGeneration to the sub Resources

mrkisaolamb:
Depends-on: https://github.com/openstack-k8s-operators/mariadb-operator/pull/222


controllers/novaapi_controller.go
gibizer, line 1: we need to check the api statefulset observedgeneration when we look at the readyCount and set the DeploymentReady condition

controllers/novacompute_controller.go
gibizer, line 1: we need to check the statefulset observeredGeneration when we look at the ReadyCount and setting the DeploymentReady condition

gibizer:
b9b9aec6d7e57e4a4498612ac57c729a1e91c1d6 can be dropped form the PR as it is independently merged to main already.

gibizer:
> [b9b9aec](https://github.com/openstack-k8s-operators/nova-operator/commit/b9b9aec6d7e57e4a4498612ac57c729a1e91c1d6) can be dropped form the PR as it is independently merged to main already.

you amended the new test changes into that commit so it cannot be dropped.

gibizer (commented):
I needed to break down this change to smaller independent steps to be able the follow what test changes needed because of what code changes. I did not finished the hole breakdown but it is here for reference https://github.com/gibizer/nova-operator/commits/pr-739-smaller-steps/ Each commit passes the tests except the unfinished last two.

gibizer:
9ef6ea9768a8b6ef512c038e67068dc6fa8798fa already has 2 test failing for me
```
Summarizing 2 Failures:
  [FAIL] NovaNoVNCProxy controller when starts zero replicas [It] and deployment is Ready
  /home/gibi/go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.3.1-0.20240429052447-09a614506ca6/test/helpers/conditions.go:53
  [FAIL] NovaCell controller when NovaCell/cell1 with metadata is reconfigured [It] applies zero replicas to NovaNoVNCProxy if requested
  /home/gibi/go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.3.1-0.20240429052447-09a614506ca6/test/helpers/conditions.go:53
```

mrkisaolamb:
To resolve test issue we need to merge and bump two patches https://github.com/openstack-k8s-operators/mariadb-operator/pull/222 and https://github.com/openstack-k8s-operators/lib-common/pull/502

controllers/nova_controller.go
```diff
@@ -1117,13 +1118,15 @@ func (r *NovaReconciler) ensureAPI(
...
-	c := api.Status.Conditions.Mirror(novav1.NovaAPIReadyCondition)
-	// NOTE(gibi): it can be nil if the NovaAPI CR is created but no
-	// reconciliation is run on it to initialize the ReadyCondition yet.
-	if c != nil {
-		instance.Status.Conditions.Set(c)
+	if api.Generation == api.Status.ObservedGeneration {
+		c := api.Status.Conditions.Mirror(novav1.NovaAPIReadyCondition)
+		// NOTE(gibi): it can be nil if the NovaAPI CR is created but no
+		// reconciliation is run on it to initialize the ReadyCondition yet.
+		if c != nil {
```
gibizer, line 1125: the funny thing is that I think this cannot happen any more as if the NovaAPI CR is not reconciled yet then nothing bumped the ObservedGeneration in it to be equal to the Generation. Don't fix it here but in a follow up PR. Also I think this is true for all the similar code places.

test/functional/novacell_controller_test.go
```diff
@@ -1015,6 +1015,7 @@ var _ = Describe("NovaCell controller", func() {
 			}, timeout, interval).Should(Succeed())
 
 			Eventually(func(g Gomega) {
+				th.SimulateStatefulSetReplicaReady(cell1.NoVNCProxyStatefulSetName)
```
gibizer, line 1020: One thing occurred to me. Why we put this in the Eventually block. Then I realized that we need that as we don't know when the controller will set the 0 replica on the StatefulSet so running it outside of Eventually would be racy if it run before the controller was able to set 0 replica on the StatefulSet. So this is good as is.

test/functional/nova_compute_ironic_controller_test.go
```diff
@@ -529,6 +529,7 @@ var _ = Describe("NovaCompute with ironic diver controller", func() {
 		It("and deployment is Ready", func() {
 			ss := th.GetStatefulSet(cell1.NovaComputeStatefulSetName)
 			Expect(int(*ss.Spec.Replicas)).To(Equal(0))
+			th.SimulateStatefulSetReplicaReady(cell1.NovaComputeStatefulSetName)
```
gibizer, line 532: here we don't need eventually as the GetStatefulSet has eventually and in this scenario the StatefulSet is created with 0 replicas as the only update on it.

gibizer (commented):
I reviewed the first few commits. The next on my list is "Add ObservedGeneration logic too top lvl services"

test/functional/nova_compute_ironic_controller_test.go
gibizer: so this is also good as is

test/functional/nova_multicell_test.go
```diff
@@ -531,6 +531,7 @@ var _ = Describe("Nova multi cell", func() {
 						novaNames.APIMariaDBDatabaseName.Name, novaNames.Namespace)),
 			)
 
+			SimulateReadyOfNovaTopServices()
```
gibizer: between this call and the previous call to SimulateReadyOfNovaTopServices() I don't see any test code that changes any nova related CRs. So only one of the call should be needed.

test/functional/nova_multicell_test.go
```diff
@@ -552,6 +553,7 @@ var _ = Describe("Nova multi cell", func() {
 				corev1.ConditionTrue,
 			)
 			Eventually(func(g Gomega) {
+				SimulateReadyOfNovaTopServices()
```
gibizer: This also seems strange as no change was triggered on the CRs since the last call to SimulateReadyOfNovaTopServices also nothing changes CRs in the Eventually block either.

test/functional/nova_reconfiguration_test.go
```diff
@@ -145,15 +142,15 @@ func CreateNovaWith3CellsAndEnsureReady(novaNames NovaNames) {
...
-	th.SimulateStatefulSetReplicaReady(novaNames.MetadataStatefulSetName)
 
 	th.ExpectCondition(
 		novaNames.NovaName,
 		ConditionGetterFunc(NovaConditionGetter),
 		novav1.NovaAllCellsReadyCondition,
 		corev1.ConditionTrue,
 	)
+	SimulateReadyOfNovaTopServices()
+	th.SimulateStatefulSetReplicaReady(novaNames.MetadataStatefulSetName)
```
gibizer, line 153: I think SimulateReadyOfNovaTopServices already calls SimulateStatefulSetReplicaReady on the metadata so we don't need an explicit call.

gibizer (commented):
Left comments on commit  "Add ObservedGeneration logic too top lvl services" in the series.

gibizer:
using the latest from here and the merged mariadb pseudoversion of the harness fix I still see unstable tests.
I did 3 runs. One passed, the other two each failed with a single tc:
```
  [FAIL] Nova controller when Nova CR instance is created with NetworkAttachment and ExternalEndpoints [It] creates all the sub CRs and passes down the network parameters
  /home/gibi/go/pkg/mod/github.com/openstack-k8s-operators/keystone-operator/api@v0.3.1-0.20240525111221-1e3ee314289c/test/helpers/crd.go:241
```

```
  [FAIL] Nova reconfiguration [BeforeEach] when the service password in the osp secret is changed updates the hash in the statefulsets to trigger the restart with the new config
  /home/gibi/upstream/git/openstack-k8s-operators/nova-operator/test/functional/base_test.go:1034
```
in both failures the keystoneendpoint was not found.

test/functional/nova_multicell_test.go
```diff
@@ -255,7 +255,7 @@ var _ = Describe("Nova multi cell", func() {
 
 			Expect(configData).To(ContainSubstring("transport_url=rabbit://cell0/fake"))
 
-			th.SimulateStatefulSetReplicaReady(novaNames.APIStatefulSetName)
+			SimulateReadyOfNovaTopServices()
```
gibizer, line 258: I saw a failure here as well and dig into it. The sequence of evens from the log https://privatebin.corp.redhat.com/?6311ba861ac8041f#4RT6JhpDyXXW3ot8HjA993FbpZvxZzoKHBhyXeUD5d4a is the following: 
1. cell0 is ready but and the test set the cell mapping job to success too L216 but the Nova controller not yet reconciled to see the cell mapping job success
2. As cell0 is ready NovaAPI is created and creates the StatefulSet but that is not ready yet
3. The SimulateReadyOfNovaTopServices call makes the API statefulset ready and waits for the keystone endpoint. The NovaAPI controller not yet seen the updated statefulset
4. The Nova controller sees the cell mapping job success and updates the NovaAPI with the new registered cell hash
5. NovaAPI controller updates the statefulset as the input hash changed (that includes the registered cells) so the statefulset generation is now bigger than when the test made it ready
6. NovaAPI never creates the KeystoneEndpoint as the statefulset never be ready
7. Test times out


So here what missing is waiting for the registered cells information to propagate to the NovaAPI before the test makes the statefulset ready

mrkisaolamb:
I reorganized the bumping of generation and added additional readiness checks. I was able to run the entire test suite 40 times without errors, but I will try to add an additional check for cell0 and repeat the whole process.

gibizer:
In my first local run I still get a failure:
```
••••••••••••••••••••••••••••••••••••••••

Summarizing 1 Failure:
  [FAIL] Nova controller when Nova CR instance is created with NetworkAttachment and ExternalEndpoints [It] creates all the sub CRs and passes down the network parameters
  /home/gibi/go/pkg/mod/github.com/openstack-k8s-operators/keystone-operator/api@v0.3.1-0.20240525111221-1e3ee314289c/test/helpers/crd.go:241

Ran 322 of 322 Specs in 25.866 seconds
FAIL! -- 321 Passed | 1 Failed | 0 Pending | 0 Skipped

coverage: 81.9% of statements
composite coverage: 81.9% of statements

Ginkgo ran 1 suite in 28.582927435s

Test Suite Failed
make: *** [Makefile:145: test] Error 1
❯ git log -p | head -n 1
commit 659c667ed944aade0255088ec16d07588763703e
```

full testcase log for the failure https://privatebin.corp.redhat.com/?ddfbc6c15cfb0b90#XZEs6UDePFY4mDND6RtXKEKSRidpM8RvcZCTdAC9sc6

mrkisaolamb:
This PR should pass. The last errors were related to too short timeouts, and the database didn't have time to reach the correct state.

gibizer (approved):
The logic in the controllers looks good to me. The CI is green and the envtest is stable form me after a more than 70 runs.

couple of follow ups:
* remove the nil checking logic when we mirroring conditions from nova subresources as we have a generation check round it and observedGeneration >= 1 means the subresource controller already saw the CR and therefore initialized the condition list
* refine how much we simulate success on all the things. We will do this one by one to go from green CR to green CR.


Thank you @mrkisaolamb for pushing through it.

gibizer:
recheck kuttl retry limit

mrkisaolamb:
recheck

gibizer:
> recheck

CI is blocked until https://github.com/openstack-k8s-operators/openstack-operator/pull/834 merges.

mrkisaolamb:
recheck

gibizer (approved):
Just a clean rebase. I also run this version through 30 iteration of envtest and it is stable.

Thanks @mrkisaolamb!

PR #740, opened 2024-04-18
Ensure golangci-lint runs on all files

gibizer:
This needed the following adjustments:

* rename variables to follow style
* removed unused function parameters
* changed error string to start with lower case
* added missing error handling
* ignore revive dot-imports rule for ginkgo and gomega
* added missing assert of Eventually block
* improved gomega assert usage


mrkisaolamb (approved):
lgtm

PR #745, opened 2024-04-22
tempest skip list reduction

SeanMooney:
- remove auto calultated values for tempest
- [WIP] remove outdated test skips

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -90,14 +90,12 @@
       # require cinder or swift
       cifmw_tempest_tempestconf_config:
           overrides: |
-            compute.min_compute_nodes 3
+            identity.v3_endpoint_type public
             compute-feature-enabled.vnc_console true
             compute-feature-enabled.stable_rescue true
             compute-feature-enabled.hostname_fqdn_sanitization true
             compute-feature-enabled.live_migration true
             compute-feature-enabled.block_migration_for_live_migration true
-            validation.run_validation true
-            identity.v3_endpoint_type public
             service_available.swift false
             service_available.cinder false
       cifmw_test_operator_tempest_include_list: |
@@ -141,9 +139,11 @@
       cifmw_run_test_role: test_operator
       cifmw_test_operator_concurrency: 4
       cifmw_test_operator_timeout: 7200
+      # NOTE(gibi): identity.v3_endpoint_type override is a WA to force the
+      # publicURL as otherwise tempest gets configured with adminURL and that
+      # causes test instability.
       cifmw_tempest_tempestconf_config:
           overrides: |
-            compute.min_compute_nodes 3
             identity.v3_endpoint_type public
             compute-feature-enabled.vnc_console true
             compute-feature-enabled.stable_rescue true
@@ -151,7 +151,6 @@
             compute-feature-enabled.live_migration true
             compute-feature-enabled.block_migration_for_live_migration false
             volume.storage_protocol ceph
-            validation.run_validation true
             service_available.swift false
             service_available.cinder true
       cifmw_test_operator_tempest_include_list: |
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -82,46 +82,30 @@
...
             compute-feature-enabled.vnc_console true
             compute-feature-enabled.stable_rescue true
             compute-feature-enabled.hostname_fqdn_sanitization true
             compute-feature-enabled.live_migration true
             compute-feature-enabled.block_migration_for_live_migration true
-            validation.run_validation true
-            identity.v3_endpoint_type public
             service_available.swift false
             service_available.cinder false
+            image_ssh_user cirros
```
SeanMooney, line 100: these shoudl be prefixerd with validation.

.zuul.yaml
```diff
@@ -88,40 +88,29 @@
       # NOTE(efoley): swift and cinder are not disabled, we're just telling
       # tempest that they're unavailable so that it'll skip the tests that
       # require cinder or swift
+      # TODO(sean-k-mooney): we are currently using cirros 0.5.2 we should be using 0.6.2
+      # for now we need to use the old password cubswin:) but we should update this to gocubsgo!
+      # once we update the image
       cifmw_tempest_tempestconf_config:
           overrides: |
-            compute.min_compute_nodes 3
+            identity.v3_endpoint_type public
```
SeanMooney, line 96: note gibi's comment appares to still be valid here.
downstream we have been asked to removed this and i think we need to revert that and ensure this si set downstream too.

gibizer: could you please move the comment here as well?

SeanMooney: the comment is there on line 85

i have some local changes that ill push shortly as i am filing bugs for the two tests that failed.

SeanMooney:
test_encrypted_cinder_volumes_cryptsetup fails because of 

cinder.exception.VolumeDriverException: Volume driver reported an error: Provider plain not supported.

so that is not supported with ceph.

we may be able do disable that via a tempest config option if not ill just skip it.

tempest.scenario.test_server_volume_attachment.TestServerVolumeAttachScenarioOldVersion.test_old_versions_reject and tempest.scenario.test_server_volume_attachment.TestServerVolumeAttachmentScenario.test_server_detach_rules

both fail with 

Details: {'code': 401, 'title': 'Unauthorized', 'message': 'The request you have made requires authentication.'}

this might be because of service tokens ill double check that

SeanMooney:
yep tempest.scenario.test_server_volume_attachment.TestServerVolumeAttachmentScenario.test_server_detach_rules

failed befause of


cinder.exception.ConflictNovaUsingAttachment: Detach volume from instance 3a52c1fe-07e8-4837-84e0-db48429e610a using the Compute API

so this is because of the cve fix.

i cant find logs to confirm this but looking at the implementation of tempest.scenario.test_server_volume_attachment.TestServerVolumeAttachScenarioOldVersion.test_old_versions_reject

https://github.com/openstack/tempest/blob/790b8c82f63f3c3a303a028f0b44cd169f4f82b9/tempest/scenario/test_server_volume_attachment.py#L197-L203

it is invoking _call_with_fake_service_token which uses a service token withut the service role

which we now block in the nova api and others.

so this looks like a tempest bug.

it is conditionally adding unauthorised i.e. 401 when the user token is invliad but it should always be adding it since we return 401 when the service token does not have the service roles and we have enabled enforcement of that.

SeanMooney:
check-rdo

gibizer:
/test functional

```
Summarizing 1 Failure:
  [FAIL] Samples when nova_v1beta1_nova-compute-ironic.yaml sample is applied [DeferCleanup (Each)] Nova is created
  /go/pkg/mod/github.com/openstack-k8s-operators/lib-common/modules/common@v0.3.1-0.20240424083053-a69e2d692e30/test/helpers/instance.go:66
```
This is unrelated. The test fails during cleanup

PR #755, opened 2024-05-06
Fix service cleanup

gibizer:
This contains multiple related fixes for the service cleanup logic.
* When TLS is enabled the gopherclient needs a TLS config to be able to connect to the openstack REST APIs
* The gopherclient by default used nova API microversion 2.1. Before 2.53 the DELETE /os-services/<id> REST API call used the integer primary key as the service ID which is ambiguous as each cell DB has its own services table. The default microversion is bumped to 2.95 (Antelope maximum) to use the UUID for service delete.
* We had no test coverage in kuttl that actually covered the feature with real openstack REST APIs hence we missed these bugs. Now kuttl test coverage is added that uses real keystone and nova openstack services.

Implements: [OSPRH-6259](https://issues.redhat.com//browse/OSPRH-6259)


gibizer:
I expect that the new test coverage fails until we pass the tls config to the client

gibizer:
kuttl failed as I expected:
```
2024-05-06 07:18:55.888851 | controller |     logger.go:42: 07:18:55 | scale-tests/3-scale-down-nova | running command: [sh -c (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service conductor -f value | grep cell1 | wc -l) == 1 ))
2024-05-06 07:18:55.888928 | controller |         ]
2024-05-06 07:18:58.967640 | controller |     logger.go:42: 07:18:58 | scale-tests/3-scale-down-nova | test step failed 3-scale-down-nova
2024-05-06 07:18:58.967708 | controller |     case.go:364: failed in step 3-scale-down-nova
2024-05-06 07:18:58.967716 | controller |     case.go:366: exit status 1
```
https://logserver.rdoproject.org/55/755/98d737f457f4b6480efea08bccae46cd81bd7cfc/github-check/nova-operator-kuttl/191c4a2/job-output.txt

SeanMooney:
> This contains multiple related fixes for the service cleanup logic.
> 
>     * When TLS is enabled the gopherclient needs a TLS config to be able to connect to the openstack REST APIs
> 
>     * The gopherclient by default used nova API microversion 2.1. Before 2.53 the DELETE /os-services/ REST API call used the integer primary key as the service ID which is ambiguous as each cell DB has its own services table. The default microversion is bumped to 2.95 (Antelope maximum) to use the UUID for service delete.
> 
>     * We had no test coverage in kuttl that actually covered the feature with real openstack REST APIs hence we missed these bugs. Now kuttl test coverage is added that uses real keystone and nova openstack services.

I'm a little conflicted about this as kuttl is meant to test the behaviour of 
operator not the behaviour of nova.

so while i agree with the sentiment to do some level of testing here we should not be creating nova pods or db instances in kuttl

> 
> 
> TODO:
> 
>     * check if envtest coverage can be added to cover some of the differences between the TLS and non TLS cases
> 
> 
> Implements: [OSPRH-6259](https://issues.redhat.com//browse/OSPRH-6259)

gibizer:
> > This contains multiple related fixes for the service cleanup logic.
> > ```
> > * When TLS is enabled the gopherclient needs a TLS config to be able to connect to the openstack REST APIs
> > 
> > * The gopherclient by default used nova API microversion 2.1. Before 2.53 the DELETE /os-services/ REST API call used the integer primary key as the service ID which is ambiguous as each cell DB has its own services table. The default microversion is bumped to 2.95 (Antelope maximum) to use the UUID for service delete.
> > 
> > * We had no test coverage in kuttl that actually covered the feature with real openstack REST APIs hence we missed these bugs. Now kuttl test coverage is added that uses real keystone and nova openstack services.
> > ```
> 
> I'm a little conflicted about this as kuttl is meant to test the behaviour of operator not the behaviour of nova.
> 
> so while i agree with the sentiment to do some level of testing here we should not be creating nova pods or db instances in kuttl
> 
> ## later
> well creating pods is fine what i don't really think would be correct is creating flavors/instance ie resource in the deployed nova.
> 
> I'm not sure how i feel about using the openstck client to look at the services which could be a sideffect fo the cards.
> 
> ill need to think about that a little more
> 

kuttl always used a real control plane depoyment, with real nova-api and keystone services. The change here is that the service cleanup feature of the nova-operator had no kuttl test before. And the env test coverage of the feature did not catch these bugs. So the new kuttl test step is actually verifying a nova-operator behavior as the nova-operator deletes the scaled in nova-conductor and nova-scheduler services automatically.

SeanMooney:
> 
> kuttl always used a real control plane depoyment, with real nova-api and keystone services. The change here is that the service cleanup feature of the nova-operator had no kuttl test before. And the env test coverage of the feature did not catch these bugs. So the new kuttl test step is actually verifying a nova-operator behavior as the nova-operator deletes the scaled in nova-conductor and nova-scheduler services automatically.

yes but we have been avoiding using the bash command style assert until now and we are now implicitly relying on the creating for the osc pod.

that is fine I'm just hesitating because I'm not sure how reliable doing the oc exec will be

I'm not saying this is a bad test just i hoped to never need to use the bash assert features fo kuttl

SeanMooney:
ok i tought we have previosuly agreed to not use bash asserts in this repo but if we we have already don't that for the policy stuff then i guess we can proceed

gibizer:
> ok i tought we have previosuly agreed to not use bash asserts in this repo but if we we have already don't that for the policy stuff then i guess we can proceed

I still prefer resource matching instead of shell scripting when it is possible.

test/kuttl/test-suites/default/scale-tests/04-assert.yaml
```diff
@@ -1,239 +1,35 @@
...
-    metadata:
-      labels:
-        cell: cell1
-        service: nova-conductor
+# Scaled down to 1 replicas so the other scheduler services should be
+# removed from the DB.
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+metadata:
+  name: scheduler-cleaned
```
gibizer, line 7: based on the job log I see that kuttl only runs one of the TestAsserts here. I suspect that it is due to the fact that the TestAssert was not needed and therefore the 3 def override each other. Lets see if naming them helps

SeanMooney: ah you think its its collapising commonly named asserts into one and keeping only the first or last definition

ya perhaps 
having separate names is better for debugging anyway if one of these fails.

gibizer: Still only running one assert https://logserver.rdoproject.org/55/755/30d5a8ff7a81b370b95c1445ff04f6a31a8c1a56/github-check/nova-operator-kuttl/bf780c1/job-output.txt
I either need to put all the commands in a single TestAssert or split the assert into multiple stepsh

SeanMooney: maybe you can use the test steps 

https://kuttl.dev/docs/testing/steps.html#shell-scripts


so one assert file with multiple steps

SeanMooney: based on https://kuttl.dev/docs/testing/reference.html#testassert commands is a list of command so ya you r opriotn are a single test assert with multiple commands or multiple files.

test/kuttl/test-suites/default/scale-tests/04-assert.yaml
```diff
@@ -6,6 +6,7 @@ kind: TestAssert
 namespaced: true
 commands:
   - script: |
+      oc exec -n $NAMESPACE openstackclient -- openstack compute service list
       (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-scheduler -f value | wc -l) == 1 ))
-      (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-conductor -f value | grep cell1 | wc -l) == 1 ))
+      (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service conductor -f value | grep cell1 | wc -l) == 1 ))
```
gibizer, line 11: intentionally breaking the check to make sure the script can fail in the middle

gibizer: OK, we need set -e as kuttl passed even though the check should fail

gibizer: euxo pipefail make sense.

gibizer:
Learnings form kuttl so far:
* you cannot have multiple TestAssert in a same assert file, only the last will be executed even if you name the TestAsserts differently
* When there are multiple TestStep in a step file then they are applied in parallel not sequentially. So inserting sleeps between TestSteps within the same step file is ineffective. The steps after the sleep needs to be in the next step file to form a sequence.

gibizer:
now the sleep is proper and the kuttl failed on the middle command in the script where I intentionally injected the bug. So now kuttl works as expected. So I removed the last commit with the fault injection.

test/kuttl/test-suites/default/scale-tests/05-trigger-service-cleanup.yaml
```diff
@@ -0,0 +1,27 @@
+# The service cleanup happens eventually at the next reconciliation of the
+# given service sub CR. To make the test more stabile we directly triggering
+# reconciliation for the conductor and scheduler CRs by manipulating
+# annotations on them.
```
SeanMooney, line 4: hum ok i guess thats a nice trick.
also ya it will make it faster too

controllers/common.go
```diff
@@ -504,23 +513,59 @@ func (r *ReconcilerBase) ensureMetadataDeleted(
...
 	GetKeystoneUser() string
+	GetCABundleSecretName() string
 }
 
 func getNovaClient(
-	keystoneAuth keystoneAuth, password string,
+	ctx context.Context,
+	h *helper.Helper,
+	auth clientAuth,
+	password string,
```
SeanMooney, line 526: eventually we should take the micro-version asn in input here.
we can defer that to the point when we actually need to use different microversion for different requests.

controllers/common.go
```diff
@@ -126,9 +129,15 @@ func cleanNovaServiceFromNovaDb(
...
 	}
+
 	for _, service := range allServices {
 		if service.State == "down" {
-			services.Delete(computeClient, service.ID)
+			rsp := services.Delete(computeClient, service.ID)
+			if rsp.Err != nil {
+				l.Error(rsp.Err, "Failed to delete service", "service", service, "response", rsp)
+				return rsp.Err
+			}
```
SeanMooney, line 139: ah ok we are passing in the logger to log the response

I guess that would be required for debugging so that makes sense

test/functional/base_test.go
```diff
@@ -78,22 +78,22 @@ func NewNovaAPIFixtureWithServer(log logr.Logger) *NovaAPIFixture {
...
-	f.APIRequests = append(f.APIRequests, map[string]string{
-		"method": r.Method,
-		"path":   r.URL.Path,
-		"query":  r.URL.RawQuery,
-	})
+	f.APIRequests = append(f.APIRequests, *r)
 }
 
-func (f *NovaAPIFixture) FindRequest(method string, path string, query string) bool {
+func (f *NovaAPIFixture) GetRequest(method string, path string, query string) *http.Request {
```
SeanMooney, line 84: nit: GetRequest make me think of a request with a GET HTTP method.

FindRequest is more descriptive but you are using it for a different reasons

could we keep this as fined request

test/functional/base_test.go
```diff
@@ -78,22 +78,22 @@ func NewNovaAPIFixtureWithServer(log logr.Logger) *NovaAPIFixture {
...
+			if request.URL.RawQuery == query || query == "" {
+				return &request
 			}
 		}
 	}
-	return false
+	return nil
+}
+
+func (f *NovaAPIFixture) FindRequest(method string, path string, query string) bool {
```
SeanMooney, line 95: and rename this to HasRequest

i wold be ok with this also being done in a follow up if you like but
i find the proposed naming kind of confusing.

SeanMooney (commented):
+1 some comments inline

controllers/common.go
```diff
@@ -530,7 +575,12 @@ func getNovaClient(
 	if err != nil {
 		return nil, err
 	}
-	return computeClient.GetOSClient(), nil
+
+	client := computeClient.GetOSClient()
+	// NOTE(gibi): We use Antelope maximum because we can. In reality we only
+	// need 2.53 to be able to delete services based on UUID.
+	client.Microversion = "2.95"
```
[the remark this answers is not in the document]
gibizer: This can serve as a good minimum going forward as we never need to support older than Antelope openstack with the operators.

test/functional/base_test.go
gibizer: Done

test/functional/base_test.go
gibizer: Done

SeanMooney (approved):
thanks this is eaiser to follow.

ill approve this for now and then you can keep/remove the hold once we are outside fo the freeze period

SeanMooney:
hum actully looking at the runnign ci job https://review.rdoproject.org/zuul/build/dd60a7399a554184b4e0170df2d57cc9

this is failing with 

 openstack resource provider trait list: error: the following arguments are required: <uuid>

SeanMooney:
actully no i think the error is 

 logger.go:42: 13:07:13 | scale-tests/5-trigger-service-cleanup | NovaScheduler:nova-kuttl-default/nova-kuttl-scheduler updated
    logger.go:42: 13:07:13 | scale-tests/5-trigger-service-cleanup | running command: [sh -c set -euxo pipefail
        (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-scheduler -f value | wc -l) == 1 ))
        (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-conductor -f value | grep cell1 | wc -l) == 1 ))
        (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-conductor -f value | grep cell0 | wc -l) == 1 ))
        ]
    logger.go:42: 13:07:13 | scale-tests/5-trigger-service-cleanup | ++ oc exec -n nova-kuttl-default openstackclient -- openstack compute service list --service nova-scheduler -f value
    logger.go:42: 13:07:13 | scale-tests/5-trigger-service-cleanup | ++ wc -l
    logger.go:42: 13:08:24 | scale-tests/5-trigger-service-cleanup | Unable to establish connection to http://nova-public.nova-kuttl-default.svc:8774/v2.1/os-services: HTTPConnectionPool(host='nova-public.nova-kuttl-default.svc', port=8774): Max retries exceeded with url: /v2.1/os-services?binary=nova-scheduler (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fae05130c10>: Failed to establish a new connection: [Errno 113] No route to host'))
    logger.go:42: 13:08:24 | scale-tests/5-trigger-service-cleanup | command terminated with exit code 1

test/kuttl/test-suites/default/scale-tests/05-assert.yaml
```diff
@@ -0,0 +1,12 @@
...
+# removed from the DB.
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
+      set -euxo pipefail
+      (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-scheduler -f value | wc -l) == 1 ))
+      (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-conductor -f value | grep cell1 | wc -l) == 1 ))
+      (( $(oc exec -n $NAMESPACE openstackclient -- openstack compute service list --service nova-conductor -f value | grep cell0 | wc -l) == 1 ))
```
SeanMooney, line 12: hum looking at your fork you scale down in step 3

https://github.com/openstack-k8s-operators/nova-operator/blob/8f5d9ff5a121c2768f06adc59760e4a078ad3be4/test/kuttl/test-suites/default/scale-tests/03-scale-down-nova.yaml

and then in step 4 reconfigure
the downtime.

then you assert there.

in ci this never gets to 1 and stays at 3

gibizer:
Kuttl can always surprise me:
```
2024-05-08 13:06:48.171258 | controller |     logger.go:42: 13:06:48 | scale-tests/4-reconfigure-service-downtime-and-wait | starting test step 4-reconfigure-service-downtime-and-wait
2024-05-08 13:06:48.171519 | controller |     logger.go:42: 13:06:48 | scale-tests/4-reconfigure-service-downtime-and-wait | running command: [sh -c sleep 25
2024-05-08 13:06:48.171539 | controller |         ]
2024-05-08 13:07:13.256400 | controller |     logger.go:42: 13:07:13 | scale-tests/4-reconfigure-service-downtime-and-wait | Nova:nova-kuttl-default/nova-kuttl updated
2024-05-08 13:07:13.256507 | controller |     logger.go:42: 13:07:13 | scale-tests/4-reconfigure-service-downtime-and-wait | test step completed 4-reconfigure-service-downtime-and-wait
2024-05-08 13:07:13.256554 | controller |     logger.go:42: 13:07:13 | scale-tests/5-trigger-service-cleanup | starting test step 5-trigger-service-cleanup
```
Look at the timestamps of step4. It shows that it is not really parallel as I assumed. It does do a sequential apply of the steps but it does it in a random order from the step file. And this run the sleep was run first, then after the full sleep the service_down_time reconfiguration run.  I need to write up a patch to our kuttl guide to note these behaviors.

However this does not explain why the schedulers are not cleaned from the DB. As the replica scale down was in step 3, step 4 did the reconf and waited enough to all the scaled down service replicas to go to down state. So the reconcile trigger in step 5 should have cleaned them.

As you noted the first nova-api request in step 5 resulted in an error
```
2024-05-08 13:08:24.525619 | controller |     logger.go:42: 13:08:24 | scale-tests/5-trigger-service-cleanup | Unable to establish connection to http://nova-public.nova-kuttl-default.svc:8774/v2.1/os-services: HTTPConnectionPool(host='nova-public.nova-kuttl-default.svc', port=8774): Max retries exceeded with url: /v2.1/os-services?binary=nova-scheduler (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fae05130c10>: Failed to establish a new connection: [Errno 113] No route to host'))
```
This suggest that since the NovaAPI reconfiguration in step3 the nova-api service did not come up successfully. (even though there was a 25 sec sleep in between). This could affect the service cleanup code in the the operator as well as it also tries to call the nova-api. 

Indeed a similar error visible in the operator log:
```
2024-05-08T17:07:16.644Z	ERROR	Controllers.NovaScheduler	Failed cleaning services from nova db	{"controller": "novascheduler", "controllerGroup": "nova.openstack.org", "controllerKind": "NovaScheduler", "NovaScheduler": {"name":"nova-kuttl-scheduler","namespace":"nova-kuttl-default"}, "namespace": "nova-kuttl-default", "name": "nova-kuttl-scheduler", "reconcileID": "6c0fe372-0f46-4556-b697-07563334503b", "error": "Get \"http://nova-internal.nova-kuttl-default.svc:8774/v2.1/os-services?binary=nova-scheduler\": read tcp 10.217.0.101:38274->10.217.4.196:8774: read: connection reset by peer"}
```

So the question is why the nova-api is not up after 25 seconds of the last reconfiguration of the NovaAPI. (Note that the scheduler controller does not retry the service cleanup on its own, it only tries it if for any other reason there is a reconciliation of the scheduler. So after the initial failure of the cleanup no retry happens during the test.) 

Ahh I see now. My first observation in this comment about the reordering of the sleep is relevant. The nova-api service_down_time reconfiguration happens right before the step 5 triggering the reconcile of the scheduler and conductor. So the nova-api service is being redeployed during step5 causing the api access errors and therefore the failure of the scheduler and conductor cleanup. 

The test passed before when the random ordering of kuttl tasks within the step happened to do the nova-api reconf first, then the sleep, and fails now with the other ordering. Either I have to move the sleep to its own step file to further enforce ordering, or I need to add an assert to step4 that checks that the nova-api is up after the reconfiguration. The latter would require to asset both readyness of the NovaAPI resource and the up-to-date-ness of it via comparing its Generation with its Status.ObservedGeneration. I will go with the former as ObservedGeneration is not available yet on the NovaAPI subresorce.

gibizer:
Documented my learnings about kuttl in https://github.com/openstack-k8s-operators/dev-docs/pull/108

test/kuttl/test-suites/default/scale-tests/05-wait-for-service-downtime.yaml
```diff
@@ -0,0 +1,7 @@
+# Wait a bit more than the service_down_time to make sure that scaled down
+# services are considered down.
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      sleep 25
```
SeanMooney, line 7: instead of doing this let do it properly and add an assert to stage 4 to confirm the nova CR is ready and that the custom service config is set.

that will make use wiait without this potential race

gibizer: Even if the nova CR is ready and the custom service_down_time is set, the test needs to wait for the service_down_time amount of time before triggering the reconcile on Conductor and Scheduler as the service cleanup only deletes down service form the DB.

SeanMooney: right but kuttl also wait for the assert to be true automtiacaly for a given period of time.

it basically keeps reapply ing the assert until we hit the timeout which we set to 300 seconds
https://github.com/openstack-k8s-operators/nova-operator/blob/main/test/kuttl/test-suites/default/config.yaml#L7
so you should not need to have this sleep because the next stage will try for up to 5 mins for the assert to pass

gibizer: But we don't only need to re-apply the assert 6 after 25 periodically. We need to trigger a reconcile (step 6) after 25 seconds of the replica change (step 3) and after the service_down_time is set to 20 (step 4)

gibizer: Do you suggest to move the triggering of the reconcile from step 6 into the 06-assert.yaml?

gibizer: We discussed this with @SeanMooney over today's call and agreed that we can improve the actual implementation separately to instead of deleting down services delete the services by name and delete them if the index of the end of the name is higher than the current ready StatefulSet replica count. I will file a Jira for this.

gibizer: Filed https://issues.redhat.com/browse/OSPRH-6912 to track the impl improvement

gibizer:
/unhold
beta freeze is over

SeanMooney (approved):
yep tracker for the follow up is filed https://issues.redhat.com/browse/OSPRH-6912

and this si functional even if we can refine it more later so i agree that its good to proceed with this for now

PR #763, opened 2024-05-20
Remove ensureSecret and use one from lib-common

### LINKED ISSUE #649: Move ensureSecret to lib-common and use it in nova
ensureSecret is similar both in placement and in nova so it can be moved to central place

mrkisaolamb:
Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/507


bogdando (commented):
is this still needed?

PR #764, opened 2024-05-20
Set fips-compliant CRD feature flag

bshephar:
This change adds the feature flag to indicate that our operator is fips compliant.


gibizer:
@bshephar same question as in the placement-operator commit. Do we need to land this in Beta?

gibizer (approved):

PR #773, opened 2024-05-27
Make the ceph job non voting until fixed

gibizer:
The job might be back to green after
openstack-k8s-operators/ci-framework#1757
and openstack-k8s-operators/dataplane-operator#896 lands


SeanMooney (approved):
in general i do not like skiping this for more then a day or two but i agree we can make it non voting until we resolve this.

approving since Martin's patch is currently in merge conflict.

PR #776, opened 2024-05-29
Wait for DB writes to propagate (causality checks)

gibizer:
Because we deploy the database in multi-master mode we can have cases
where reading data from the DB fails after the data was written to
another galera master node.

We saw this happening during Flavor creation where nova first creates
the flavor in the DB then the notification sending code tries to read
back the flavor projects from the DB and the query fails as the flavor
is not found.

In this patch we change the default value on the DB engine to wait for
previous write to sync before read.

This will have a performance impact.

We use `mysql_wsrep_sync_wait` from oslo.db [1] setting it to 1 as per
the documented values of this parameter in the DBMS [2][3].

[1]: https://opendev.org/openstack/oslo.db/commit/009d23df45969036c70e4cf59eb4019aaace9a55
[2]: https://mariadb.com/docs/server/ref/mdb/system-variables/wsrep_sync_wait/
[3]: https://galeracluster.com/library/documentation/mysql-wsrep-options.html

Fixes: https://issues.redhat.com/browse/OSPRH-7198

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -211,12 +211,22 @@ cpu_power_management=true
 {{if (index . "cell_db_address")}}
 [database]
 connection = mysql+pymysql://{{ .cell_db_user }}:{{ .cell_db_password}}@{{ .cell_db_address }}/{{ .cell_db_name }}?read_default_file=/etc/my.cnf
+
+# Wait for writes to complete when doing a read.
+# Relevant for multi-master galera deployments
+# https://mariadb.com/docs/server/ref/mdb/system-variables/wsrep_sync_wait/
+mysql_wsrep_sync_wait = 1
 {{end}}
 
 
 {{if (index . "api_db_address")}}
 [api_database]
 connection = mysql+pymysql://{{ .api_db_user }}:{{ .api_db_password }}@{{ .api_db_address }}/{{ .api_db_name }}?read_default_file=/etc/my.cnf
+
+# Wait for writes to complete when doing a read.
+# Relevant for multi-master galera deployments
+# https://mariadb.com/docs/server/ref/mdb/system-variables/wsrep_sync_wait/
+mysql_wsrep_sync_wait = 1
 {{end}}
 
 [keystone_authtoken]
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -319,7 +319,6 @@ var _ = Describe("Nova controller", func() {
 					cell0Account.Spec.UserName, cell0Secret.Data[mariadbv1.DatabasePasswordSelector],
 					novaNames.Namespace)),
 			)
-
 			apiAccount := mariadb.GetMariaDBAccount(novaNames.APIMariaDBDatabaseAccount)
 			apiSecret := th.GetSecret(types.NamespacedName{Name: apiAccount.Spec.Secret, Namespace: novaNames.APIMariaDBDatabaseAccount.Namespace})
 
@@ -327,6 +326,9 @@ var _ = Describe("Nova controller", func() {
 				ContainSubstring(fmt.Sprintf("[api_database]\nconnection = mysql+pymysql://%s:%s@hostname-for-openstack.%s.svc/nova_api?read_default_file=/etc/my.cnf",
 					apiAccount.Spec.UserName, apiSecret.Data[mariadbv1.DatabasePasswordSelector], novaNames.Namespace)),
 			)
+
+			Expect(configData).To(ContainSubstring("mysql_wsrep_sync_wait = 1"))
+
 			// NOTE(gibi): cell mapping for cell0 should not have transport_url
 			// configured. As the nova-manage command used to create the mapping
 			// uses the transport_url from the nova.conf provided to the job


zzzeek (approved):

templates/nova.conf
```diff
@@ -211,12 +211,22 @@ cpu_power_management=true
 {{if (index . "cell_db_address")}}
 [database]
 connection = mysql+pymysql://{{ .cell_db_user }}:{{ .cell_db_password}}@{{ .cell_db_address }}/{{ .cell_db_name }}?read_default_file=/etc/my.cnf
+
+# Wait for writes to complete when doing a read, update, or insert
+# Relevant for multi-master galera deployments
+# https://mariadb.com/docs/server/ref/mdb/system-variables/wsrep_sync_wait/
+mysql_wsrep_sync_wait = 7
```
SeanMooney, line 218: ack based on the doc 7 is

Performs causality checks on read, update, and insert operations, such as SELECT, UPDATE, DELETE, INSERT, and REPLACE statements

that looks correct to me.

zzzeek (changes requested):
let's try starting with the setting at 1, which is the equivalent of wsrep_causal_reads = true.   we should not need a wsrep check on INSERTs and unlikely for UPDATE/DELETE

test/functional/nova_controller_test.go
```diff
@@ -319,14 +319,16 @@ var _ = Describe("Nova controller", func() {
...
-
 			apiAccount := mariadb.GetMariaDBAccount(novaNames.APIMariaDBDatabaseAccount)
 			apiSecret := th.GetSecret(types.NamespacedName{Name: apiAccount.Spec.Secret, Namespace: novaNames.APIMariaDBDatabaseAccount.Namespace})
 
 			Expect(configData).To(
 				ContainSubstring(fmt.Sprintf("[api_database]\nconnection = mysql+pymysql://%s:%s@hostname-for-openstack.%s.svc/nova_api?read_default_file=/etc/my.cnf",
 					apiAccount.Spec.UserName, apiSecret.Data[mariadbv1.DatabasePasswordSelector], novaNames.Namespace)),
 			)
+
+			Expect(configData).To(ContainSubstring("mysql_wsrep_sync_wait = 7"))
```
SeanMooney, line 330: ack this is really the only thing we can test
+1

SeanMooney (approved):

zzzeek:
https://forums.percona.com/t/stability-of-8-0-x/13005/5

> wsrep_sync_wait = 7 is quite intense and will certainly cause slowness in your application. Have you tested with this disabled or maybe =1?

templates/nova.conf
zzzeek: are we doing INSERTs that rely on existing data being read?    Whats the INSERT scenario here?   note that this flag started in earlier galera as wsrep_causal_reads=1, which is equivalent to "1" here.   I dont think we should dampen performance ahead of what's needed

zzzeek:
> > let's try starting with the setting at 1, which is the equivalent of wsrep_causal_reads = true. we should not need a wsrep check on INSERTs and unlikely for UPDATE/DELETE
> 
> for db sync perhaps but insert are likely important for online migrations

an INSERT is a row with new data in it.   doing a causality check for an INSERT seems unnecessary for anything I can think of.     if the goal is that an INSERT with conflicting data should fail, that will happen anyway when the write-sets try to reconcile.    note that setting this to 1 will give you causality check for all SELECTs.

SeanMooney:
if we belive 1 is sufficent to ensure when we do the min comptue service version check  as part of adoption the I'm ok with starting with 1

Nova in general is not expecting Galara to be eventually consistent
it may work but it will impact scheduling,  provided transactions are still atomic
i.e. we will never use outdated data to make decisions because reads force synchronisation
then it might be fine.

why I expected 7 was to ensure when we claim a PCI device or similar in the db we want to ensure those are visible to the scheduler.

SeanMooney:
based on your last respocne 1 is probably sufficient as the read to get the aviabel PCI devices will enforce the causality check.

our expectation however is that if we create a writer transaction and do an insert form 2 different process in parallel only one of them will complete. and the other will raise a db conflict

zzzeek:
So I'd discourage the term "Eventually consistent" since galera is not really an "eventually" model, where "eventually" is usually taken to mean "can be any amount of time, seconds, minutes".   it's not strictly incorrect however I think is misleading; this is a very brief latent period in all cases.

we should start with SELECT being the tool we are using for visibility; an INSERT just pushes a new row.   an UPDATE or DELETE is also questionable as to whether it needs to see an up to date cluster.    if we do "UPDATE ..WHERE col='foobar'", even without multi-master, other processes can set/unset "foobar" in the database after this UPDATE takes place.

zzzeek:
> our expectation however is that if we create a writer transaction and do an insert form 2 different process in parallel only one of them will complete. and the other will raise a db conflict

this will happen anyway without wsrep_causal_reads / sync_wait.   the write-sets do this conflict in any case

cc @dciabrin

zzzeek:
that is, if the conflict is being raised because there's a UNIQUE/INDEX/FOREIGN KEY/CHECK doing a check, those are enforced. galera transactions are consistent across masters

SeanMooney:
ok then we can regert this to 1

all write are done in a transaction so if "galera transactions are consistent across masters"
then all our insert will be consistent.

gibizer:
Set it to 1

zzzeek (approved):

SeanMooney (approved):

karelyatin:
recheck live-migration test fail 6.0 sec

PR #775, opened 2024-05-28
Clean up scaled down services based on hostname

auniyal61:
- Update service objects in functional test api_fixture
- Adds a check to wait for readycount to become equal to replica
- Delete service host from nova DB by hostname on replica count update

Implements: [OSPRH-6912](https://issues.redhat.com//browse/OSPRH-6912)

PR #778, opened 2024-05-31
[envtest]Improve API simulator

gibizer:
Refactored and improved the nova API simulate used in envtest to cover service deletion impl. #775 changes how the server deletion logic works and it needs to be able to provide test dependent service data instead of the hardcoded one.

### DIFF
--- a/test/functional/novaconductor_controller_test.go
+++ b/test/functional/novaconductor_controller_test.go
@@ -16,7 +16,6 @@ package functional_test
 import (
 	"encoding/json"
 	"fmt"
-	"net/http"
 
 	. "github.com/onsi/ginkgo/v2" //revive:disable:dot-imports
 	. "github.com/onsi/gomega"    //revive:disable:dot-imports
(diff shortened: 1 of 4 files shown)


test/functional/novaconductor_controller_test.go
```diff
@@ -798,8 +781,8 @@ var _ = Describe("NovaConductor controller cleaning", func() {
 				corev1.ConditionTrue,
 			)
 			Expect(novaAPIServer.HasRequest("GET", "/compute/os-services/", "binary=nova-conductor")).To(BeTrue())
-			Expect(novaAPIServer.HasRequest("DELETE", "/compute/os-services/3", "")).To(BeTrue())
-			req := novaAPIServer.FindRequest("DELETE", "/compute/os-services/3", "")
+			Expect(novaAPIServer.HasRequest("DELETE", "/compute/os-services/5", "")).To(BeTrue())
```
gibizer, line 784: this was an actual bug because the fixture returned all hardcoded services regardless of the controller asking for a services of a certain binary.

mrkisaolamb (approved):
lgtm

PR #781, opened 2024-06-03
DNM:trigger more logging from nova to test log rotation

gibizer:


gibizer:
OK, the ceph job shows now that the nova-api logs are rotated and the must-gather only get the last part of the nova-log from the test run. Also the sos-report from the ocp worker/master does not have /var/log/ collected.
https://logserver.rdoproject.org/81/781/1c99a10d6dad7a7560db7d7dd04ce7f4d9dc3a3f/github-check/nova-operator-tempest-multinode/9419cc1/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-api-0/logs/

There is however a catch that nova also configures oslo log for rotation at 50MB so I will try to bump that to see if the ocp rotation does the same independently.

PR #785, opened 2024-06-12
Simplify scale-down kuttle tests

auniyal61:
- Removes tests wait to complete pods cleanup on scale-down.
- Move (service replica becomes 1) assertion for scheduler, cell0-conductor and cell1-conductor after pods cleanup.
- Remove manual triggering of reconcile for cleanup after wait.
- Rename later tests.


gibizer:
recheck retry limit

PR #787, opened 2024-06-13
Restore condition transition time at the right place

gibizer:
As we are resetting the condition list in each reconcile we have the
logic to restore the last transition timestamp of the unchanged
conditions in at the end of the reconcile. However in the nova
controller the restore call was after the instance is was persisted and
therefore it was ineffective. This could lead to extra reconciles in
openstack-operator even if the nova-operator needed to update nothing in
the underlying resources.

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -159,8 +159,8 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			instance.Status.Conditions.Set(
 				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
-		err := h.PatchInstance(ctx, instance)
 		condition.RestoreLastTransitionTimes(&instance.Status.Conditions, savedConditions)
+		err := h.PatchInstance(ctx, instance)
 		if err != nil {
 			_err = err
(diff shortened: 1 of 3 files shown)


controllers/nova_controller.go
```diff
@@ -159,8 +159,8 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			instance.Status.Conditions.Set(
 				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
-		err := h.PatchInstance(ctx, instance)
 		condition.RestoreLastTransitionTimes(&instance.Status.Conditions, savedConditions)
+		err := h.PatchInstance(ctx, instance)
```
mrkisaolamb, line 163: shoudn't we move condition.RestoreLastTransitionTimes(&instance.Status.Conditions, savedConditions) into defer func() or we should also correct other places where we have RestoreLastTransitionTimes in defer func()

mrkisaolamb (approved):
lgtm

controllers/nova_controller.go
gibizer: it was and it is in the deferred function

PR #788, opened 2024-06-14
Revert "Wait for DB writes to propagate (causality checks)"

gibizer:
This reverts commit ab95f150cbecb6cdf03795cc56d1516914c9e2fd.

The https://github.com/openstack-k8s-operators/mariadb-operator/pull/229
switched the mariadb-operator to Active/Passive mode instead of
multimaster. This means we don't need to force synchronization from the
client config any more.

Related: [OSPRH-7405](https://issues.redhat.com//browse/OSPRH-7405)


mrkisaolamb (approved):
lgtm

PR #789, opened 2024-06-14
normalize the deployment ready check

auniyal61:


gibizer:
recheck kuttl blocker is fixed

controllers/novaconductor_controller.go
```diff
@@ -283,7 +283,17 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			common.AppSelector: NovaConductorLabelPrefix,
+			CellSelector:       instance.Spec.CellName,
+		}
+		ss := statefulset.NewStatefulSet(novaconductor.StatefulSet(instance, inputHash, serviceLabels, serviceAnnotations), r.RequeueTimeout)
+		statefulSet := ss.GetStatefulSet()
+
+		if statefulSet.ObjectMeta.Generation == statefulSet.Status.ObservedGeneration {
+			Log.Info("Waiting for the ReadyCount to become equal to Replicas before doing service cleanup in nova database.")
+		}
+
```
gibizer, line 296: ensureDeployment runs before above and set the Deployment ready condition to True if both the observedGeneration is up to date and the requested replicas are the same as the readyreplicas. 

https://github.com/openstack-k8s-operators/nova-operator/blob/b8193bd7a63037aea6a6d31253c0acc3e1f9f5a6/controllers/novaconductor_controller.go#L567-L569

So the whole check here can be reduced to checking that the DeploymentReady condition is set to true or not.

You can check the status of the condition with:
```golang
    instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition)
```

auniyal61: ack, updated.

gibizer: You can drop the `*instance.Spec.Replicas != instance.Status.ReadyCount` from the if. The deploymentReady condition is only set to true if the replicas equal to readyreplicas.
https://github.com/openstack-k8s-operators/nova-operator/blob/b8193bd7a63037aea6a6d31253c0acc3e1f9f5a6/controllers/novaconductor_controller.go#L567-L569

The generic rule we want to implement is that we only call cleanup if the deployment is up to date (based on the generations) and ready (based on the readyreplicas). As the deploymentReady condition is exactly representing that we only need to check that condition here.

controllers/novaconductor_controller.go
```diff
@@ -282,8 +282,9 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
-	if *instance.Spec.Replicas != instance.Status.ReadyCount {
+	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
 		Log.Info("Waiting for the ReadyCount to become equal to Replicas before doing service cleanup in nova database.")
```
gibizer, line 286: The log message can be updated now to be more precise. Something along the line of:
```
Waiting for the deployment to be ready before doing service cleanup in the nova database.
```

gibizer (approved):
looks good. Thanks!

auniyal61:
/retest
unrelated fail during pre-commit checks `got an unexpected keyword argument 'timeout'`

auniyal61:
/retest 
deployent issue- stderr: 'error: timed out waiting for the condition on openstackcontrolplanes/openstack-galera-network-isolation'

auniyal61:
recheck
rdoproject.org/github-check needs recheck to run again.

PR #790, opened 2024-06-14
include nova base log collection playbook in all zuul jobs

SeanMooney:
This change addes the extra log colleciton playbook we use
for the kuttl job to all jobs to ensure we have a standard
basline of logs regradless of other ci changes.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -76,6 +76,8 @@
     name: nova-operator-tempest-multinode
     parent: podified-multinode-edpm-deployment-crc-3comp
     dependencies: ["openstack-k8s-operators-content-provider"]
+    post-run:
+      - ci/nova-operator-base/playbooks/collect-logs.yaml
     vars:
       cifmw_run_test_role: test_operator
       cifmw_test_operator_concurrency: 4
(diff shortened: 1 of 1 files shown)


gibizer:
recheck kuttl retry limit

gibizer (approved):
It only adds 2 minutes to the runtime of the jobs so I'm OK to include it.
But since then the must-gather also got fixed and the pods OCP node sos reports are available.

mrkisaolamb:
recheck

SeanMooney:
check-rdo 
the kuttl issues should have been fixed by https://github.com/openstack-k8s-operators/nova-operator/pull/793

SeanMooney:
check-rdo 

the ceph job failed in one test with 
AssertionError: 5.800000000000001 not less than or equal to 5.0 : Downtime of 5.800000000000001 seconds is higher than expected '5.0'

this happened in the past and i thought we tweaked the timeout but perhaps we left it at the default because it happens rarely.

this is tunable by setting 
[validation]
allowed_network_downtime
https://github.com/openstack/tempest/blob/master/tempest/config.py#L923C19-L928

if this fails again we can increase this via cifmw_tempest_tempestconf_config

PR #793, opened 2024-06-15
[CI]Remove DATAPLANE_REPO from kuttl jobdef

gibizer:
That repository is merged into the openstack-operator and install_yamls
does not have the matching variable any more. This causes that the pre
playbook fails with:
```
'cifmw_install_yamls_vars contains a variable that is not defined in install_yamls Makefile nor cifmw_install_yamls_whitelisted_vars: DATAPLANE_REPO'
```


mrkisaolamb (approved):
lgtm

PR #795, opened 2024-06-20
Switch root to apache user for ssl.conf

mrkisaolamb:


stuggi:
Is there an issue with root being the owner? We can do that change, on a default deployment where configs are 644, all files are owned by root. Only if a config should not be readable by anyone the apache group gets set

```
sh-5.1# ls -la /etc/httpd/conf.d/
total 84
drwxr-xr-x. 1 root root      22 Jun 19 08:45 .
drwxr-xr-x. 1 root root      32 Jun 17 06:13 ..
-rw-r--r--. 1 root root      63 Aug  3  2022 auth_mellon.conf
-rw-r-----. 1 root apache 58730 May 15 12:48 auth_openidc.conf
-rw-r--r--. 1 root root    2916 Feb 14 12:32 autoindex.conf
-rw-r--r--. 1 root root     400 Feb 14 12:32 README
-rw-r--r--. 1 root root     655 Jun 19 08:45 ssl.conf
-rw-r--r--. 1 root root    1252 Feb 14 12:29 userdir.conf
-rw-r--r--. 1 root root     653 Feb 14 12:29 welcome.conf
```

Wouldn't root even be more secure (if needed) because when httpd switches the worker processes to the apache user, they could not change the config as they have no write access.

stuggi:
Even if you switch away from running things as root, I don't think this change is required?

SeanMooney:
we dont run as root today we deeply as appche

and no i don think its reasonable for this to be owned by root

SeanMooney:
all files under /etc/httpd/conf.d/ shoudl really be owned by apache

stuggi:
> all files under /etc/httpd/conf.d/ shoudl really be owned by apache

then we have to change the rpms, but I think with apache not being able to write the files its more secure

SeanMooney:
this is really a minor bug but we should not be using root to own service files in the contiers in general they should be owned by the user that they are created for

if you just want to make this 444 or 440 instead of 644 that's fine but we should not be relying on the fact that its world readable today.

stuggi:
> this is really a minor bug but we should not be using root to own service files in the contiers in general they should be owned by the user that they are created for
> 
> if you just want to make this 444 or 440 instead of 644 that's fine but we should not be relying on the fact that its world readable today.

if we do that change to apache being the owner, then we should also change it to 444 or 440.

SeanMooney (approved):

PR #803, opened 2024-06-25
Ensure bundles include all the Required Infrastructure Annotations

mrkisaolamb:
Implements: [OSPRH-7922](https://issues.redhat.com/browse/OSPRH-7922)

### DIFF
--- a/config/manifests/bases/nova-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/nova-operator.clusterserviceversion.yaml
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     operatorframework.io/suggested-namespace: openstack
-    operators.openshift.io/infrastructure-features: '["disconnected"]'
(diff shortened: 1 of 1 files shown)


gibizer (commented):
I cannot comment on the TLS flag. I think we need somebody from security to make a decision. I don't know what are the "well known" tunables, and and also the "any workloads it manages" is a very blanket statement.

config/manifests/bases/nova-operator.clusterserviceversion.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "true"
```
SeanMooney, line 10: we have the same concern here as with placement.
its probably safer to say false but you could argue its true
as we expose service override for all the routes and as a result you should
be able to tweak that aspect of the tls configuration using standard annotations.

config/manifests/bases/nova-operator.clusterserviceversion.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
```
gibizer, line 13: OK this is now in line with what we agreed on the today's podified call.

gibizer (approved):

PR #804, opened 2024-06-26
DNM: test new must-gather with rotated logs

gibizer:


gibizer:
collected logs looks good.

PR #807, opened 2024-06-28
Remove KOLLA_COPY_STRATEGY as COPY_ALWAYS is the default now

### LINKED ISSUE #707: Remove KOLLA_COPY_STRATEGY as COPY_ALWAYS is the default now
this is a point of divergance form kolla

they removed this since COPY_ALWAYS was the default and no one used
COPY_ONCE so we probably should do that eventually.

this is not wrong its just unnecessary for us to support anything else in tcib.

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/set_configs.py#L371-L386 can be simplified

_Originally posted by @SeanMooney in https://github.com/openstack-k8s-operators/nova-operator/pull/698#discussion_r1502385651_

mrkisaolamb:
Closes #707

### DIFF
--- a/pkg/nova/cellmapping.go
+++ b/pkg/nova/cellmapping.go
@@ -21,7 +21,6 @@ func CellMappingJob(
 	args := []string{"-c", KollaServiceCommand}
 
 	envVars := map[string]env.Setter{}
-	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["KOLLA_BOOTSTRAP"] = env.SetValue("true")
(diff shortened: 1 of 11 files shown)


SeanMooney (changes requested):
we cant do this until we update tcib

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/set_configs.py#L371-L386

they did not follow kolla lead in the fork.

we still need to set this today.

bogdando (commented):
given the above concern about forked Kolla in TCIB, is this change request still relevant @mrkisaolamb
 ?

PR #809, opened 2024-07-02
Add posibilities to specify metadata password selector per cell

### LINKED ISSUE #524: Cannot specify metadata password selector per cell
https://github.com/openstack-k8s-operators/nova-operator/blob/8cf10addb78028c2387e0208aeaa9b72d7920226/api/v1beta1/common_types.go#L117-L123

If metadata deployed per cell then we could allow the user to specify the shared secret for each cell metadata service differently.

mrkisaolamb:
Now metadata password can be specified per cell using MetadataTemplate. If there is no defined MetadataSecret secret from top nova secret is used

close #524

### DIFF
--- a/api/bases/nova.openstack.org_nova.yaml
+++ b/api/bases/nova.openstack.org_nova.yaml
@@ -1567,6 +1567,15 @@ spec:
                     description: MetadataSecret - the name of the field to get the
                       metadata secret from the Secret
                     type: string
+                  prefixMetadataCellsSecret:
+                    default: MetadataCellsSecret
+                    description: prefixMetadataCellsSecret - the prefix name of the
+                      field to get the metadata secret from the Secret for cells.
+                      Vale of metadata_proxy_shared_secret information for the nova-metadata
+                      service. This secret is shared between nova and neutron ovn-metadata
+                      inside selected cell and if this is not defined the global metadata_proxy_shared_secret
+                      secret will be used
+                    type: string
                   service:
                     default: NovaPassword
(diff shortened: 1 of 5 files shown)


api/v1beta1/novametadata_types.go
```diff
@@ -84,6 +84,14 @@ type NovaMetadataTemplate struct {
...
 	// TLS - Parameters related to the TLS
 	TLS tls.SimpleService `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// Secret is the name of the Secret instance containing metadata_proxy_shared_secret
+	// information for the nova-metadata service. This secret is shared
+	// between nova and neutron ovn-metadata inside selected cell
+	// and if this is not defined the global metadata_proxy_shared_secret
+	// secret will be used
+	MetadataSecret string `json:"metadataSecret"`
```
gibizer, line 94: This is a deviation from the existing pattern regarding how we get secret data from the user via the Nova CR. So far our interface was a single Nova.spec.secret field and a configurable password selector to select fields out of that secret. I think this patter can be extended without the need on on the user side to create a separate secret per cell to pass the metadata secret.

There could be a per celltemplate or per metadatatemplate password selector to configure which field from the nova.spec.secret should be used for the metadata service in the given cell. We can default that to empty or default it to the same value as the NovaSpec level password selector for the metadata secret "MetadataSecret" to implement that by default all metadata uses the same secret field.

mrkisaolamb: done

PR #810, opened 2024-07-04
Adding patch rbac perm for serviceaccounts

gibizer:
We also took the opportunity and added patch to all the existing rbac
rules that had update already to avoid similar issues in the future

Resolves: https://issues.redhat.com/browse/OSPRH-8363

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -25,6 +25,7 @@ rules:
   - create
   - get
   - list
+  - patch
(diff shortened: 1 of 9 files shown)


gibizer:
this is a blocker for GA. So we need to merge it on main first and then I will cherry-pick it to 18.0.0-proposed.

mrkisaolamb (approved):
lgtm

PR #815, opened 2024-07-08
Remove the icon field entirely from the CSV

mrkisaolamb:
As per an internal conversation, we are not authorized to use the upstream logo and we can save space by just removing the icon.

Related: [OSPRH-8218](https://issues.redhat.com//browse/OSPRH-8218)


gibizer (approved):
We need to backport this to 18.0.0-proposed after this PR lands on main.

PR #823, opened 2024-07-16
Fixes config vars in nova.conf

### LINKED ISSUE #144: Do not render non nova-api specific config for nova-api server
There is a bunch of config rendered for the nova-api service that are not needed for the nova-api and therefore confusing to have it in the config of nova-api. E.g.:
- state_path
- allow_resize_to_same_host
- initial_*allocation_ration
- instance_name_template
- force_config_drive
- disable_fallback_pcpu_query
- enable_qemu_monitor_announce_self
- reserve_disk_resource_for_image_cache

```
[fedora@gibi-f35 install_yamls]$ oc rsh -c nova-api-api pod/nova-api-0
sh-5.1# egrep -e "^[^#]" /etc/nova/nova.conf
[DEFAULT]
state_path = /var/lib/nova
max_logfile_count=5
max_logfile_size_mb=50
log_rotation_type=size
log_file = /var/log/nova/nova-api.log
debug=true
instance_name_template="instance-%(uuid)s"
force_config_drive=True
transport_url=rabbit://fixme
osapi_compute_workers=1
metadata_workers=0
[oslo_concurrency]
lock_path = /var/lib/nova/tmp
[oslo_messaging_rabbit]
amqp_durable_queues=true
amqp_auto_delete=true
heartbeat_in_pthread=true
[oslo_policy]
enforce_new_defaults=true
[api]
use_forwarded_for = true
auth_strategy = keystone
[oslo_middleware]
enable_proxy_headers_parsing = True
[wsgi]
api_paste_config = /etc/nova/api-paste.ini
[upgrade_levels]
compute = auto
driver = noop
[notifications]
[cache]
enabled = True
backend = oslo_cache.dict
[workarounds]
disable_fallback_pcpu_query=true
enable_qemu_monitor_announce_self=true
reserve_disk_resource_for_image_cache=true
[database]
connection = mysql+pymysql://nova_cell0:12345678@openstack/nova_cell0
[api_database]
connection = mysql+pymysql://nova_api:12345678@openstack/nova_api
[keystone_authtoken]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url =  http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
password = 12345678
cafile = 
region_name = regionOne
[placement]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url =  http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal
[glance]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url = http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal
[neutron]
www_authenticate_uri = http://keystone-public-openstack.apps-crc.testing
auth_url = http://keystone-public-openstack.apps-crc.testing
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
username = nova
password = 12345678
cafile = 
region_name = regionOne
valid_interfaces = internal
metadata_proxy_shared_secret = 42
service_metadata_proxy = true
```

auniyal61:
- moves `allow_resize_to_same_host` under nvoa-api
- moves `[workarounds].disable_fallback_pcpu_query=true` under scheduler only and rest under nova-compute
- moves `[libvirt] conf` under `libvirt.LibvirtDriver` driver

Closes #144

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -3,7 +3,9 @@
 # overall then serializing live migrations so set this to 1 explictly
 max_concurrent_live_migrations=1
 state_path = /var/lib/nova
+{{if eq .service_name "nova-api"}}
 allow_resize_to_same_host = true
+{{end}}
 # enable log rotation in oslo config by default
 max_logfile_count=1
 max_logfile_size_mb=20
@@ -25,13 +27,12 @@ instance_usage_audit_period = hour
 initial_cpu_allocation_ratio=4.0
 initial_ram_allocation_ratio=1.0
 initial_disk_allocation_ratio=0.9
-{{end}}
 {{/*using a config drive will void issues with ovn and metadata*/}}
 force_config_drive=True
+{{end}}
 {{ if (index . "transport_url") }}
 transport_url={{.transport_url}}
 {{end}}
-
 {{if eq .service_name "nova-api"}}
 # scaling should be done by running more pods
 osapi_compute_workers=1
@@ -41,7 +42,6 @@ enabled_apis=osapi_compute
 metadata_workers=1
 enabled_apis=metadata
 {{end}}
-
 {{if eq .service_name "nova-novncproxy"}}
 {{ if (index . "SSLCertificateFile") }}
 ssl_only=true
@@ -182,19 +182,22 @@ tls_enabled={{ .MemcachedTLS }}
 backend = oslo_cache.dict
 {{end}}
 
+{{ if eq .service_name "nova-scheduler"}}
 [workarounds]
 disable_fallback_pcpu_query=true
+{{end}}
 {{ if eq .service_name "nova-compute"}}
+[workarounds]
 enable_qemu_monitor_announce_self=true
 reserve_disk_resource_for_image_cache=true
-
 # NOTE(gibi): We need this as live migration does not work with
 # cpu_mode=host-model . See https://bugs.launchpad.net/nova/+bug/2039803
 skip_cpu_compare_on_dest = true
-
 {{end}}
 
-{{ if eq .service_name "nova-compute"}}
+
+{{ if eq .service_name "nova-compute" }}
+{{ if eq .compute_driver "libvirt.LibvirtDriver" }}
 [libvirt]
 live_migration_permit_post_copy=true
 live_migration_permit_auto_converge=true
@@ -213,6 +216,7 @@ live_migration_uri = qemu+ssh://nova@%s/system?keyfile=/var/lib/nova/.ssh/ssh-pr
 # https://issues.redhat.com/browse/OSPRH-8806
 # https://issues.redhat.com/browse/OSPRH-8712
 cpu_power_management=false
+{{end}}
 {{end}}
 
 {{if (index . "cell_db_address")}}


auniyal61:
recheck unrelated keystone 500
tempest.lib.exceptions.IdentityError: Got identity error
Details: Unexpected status code 500

auniyal61:
recheck unrelated fail, network create cmd failed with error:
```
openstack network create --external --default --provider-network-type flat --provider-physical-network datacentre --no-share public
```
Error while executing command: HttpException: 502, The server returned an invalid or incomplete response.: 502 Bad Gateway

templates/nova.conf
```diff
@@ -2,8 +2,15 @@
 # concurrent live migrations are more likely to fail and are slower
 # overall then serializing live migrations so set this to 1 explictly
 max_concurrent_live_migrations=1
+
+{{if eq .service_name "nova-compute"}}
 state_path = /var/lib/nova
+instance_name_template = instance-%08x
+{{end}}
+{{if eq .service_name "nova-api"}}
 allow_resize_to_same_host = true
```
SeanMooney, line 11: this can also be used on the compute in the case if the vmware driver.
it is used by the nova-api in general 
https://github.com/openstack/nova/blob/e82854dc8c514e457528b52834d79176fe5a2135/nova/compute/api.py#L4284-L4285

since we do not support the vmware driver i guess this is fine.

templates/nova.conf
```diff
@@ -2,8 +2,15 @@
 # concurrent live migrations are more likely to fail and are slower
 # overall then serializing live migrations so set this to 1 explictly
 max_concurrent_live_migrations=1
+
+{{if eq .service_name "nova-compute"}}
 state_path = /var/lib/nova
+instance_name_template = instance-%08x
```
SeanMooney, line 8: this is the default value and i dont think we should be setting it here explicitly
its also used on the nova-api i think in some cases.
inanycase we should just remove this.

templates/nova.conf
```diff
@@ -2,8 +2,15 @@
 # concurrent live migrations are more likely to fail and are slower
 # overall then serializing live migrations so set this to 1 explictly
 max_concurrent_live_migrations=1
+
+{{if eq .service_name "nova-compute"}}
 state_path = /var/lib/nova
```
SeanMooney, line 7: state_path is used by all servies that store state. today this is only the compute agent and this is the default os perhaps we should just remove this instead.

SeanMooney (changes requested):

templates/nova.conf
auniyal61: done

templates/nova.conf
auniyal61: done

templates/nova.conf
auniyal61: done

auniyal61:
recheck unrealted

templates/nova.conf
```diff
@@ -25,9 +28,11 @@ instance_usage_audit_period = hour
 initial_cpu_allocation_ratio=4.0
 initial_ram_allocation_ratio=1.0
 initial_disk_allocation_ratio=0.9
-{{end}}
+
 {{/*using a config drive will void issues with ovn and metadata*/}}
 force_config_drive=True
```
mrkisaolamb, line 33: Could these errors from the tempest jobs be related to force_config_drive=True not always being configured?

Details:
```
Fault: {'code': 500, 'created': '2024-07-17T13:21:04Z', 'message': 'No valid host was found.'}
Request ID of server operation performed before checking the server status: req-bbd185cc-987d-4d59-b409-3b6e1b5669b9
``` 
here are links from last two runs: https://logserver.rdoproject.org/23/823/1793b528217d529ef29ad40302f19279f4b401a0/github-check/nova-operator-tempest-multinode/4f834b4/controller/ci-framework-data/tests/test_operator/tempest-tests/stestr_results.html
Second:
https://logserver.rdoproject.org/23/823/1793b528217d529ef29ad40302f19279f4b401a0/github-check/nova-operator-tempest-multinode/b98e0cc/controller/ci-framework-data/tests/test_operator/tempest-tests/stestr_results.html

templates/nova.conf
```diff
@@ -2,8 +2,11 @@
 # concurrent live migrations are more likely to fail and are slower
 # overall then serializing live migrations so set this to 1 explictly
 max_concurrent_live_migrations=1
-state_path = /var/lib/nova
```
SeanMooney, line 5: ok so the reason this is here is because of 

e nova.exception.InvalidNodeConfiguration: Invalid node identity configuration: Unable to write uuid to /usr/lib/python3.9/site-packages/compute_id: [Errno 13] Permission denied: '/usr/lib/python3.9/site-packages/compute_id'

that is using the wrong state path and i think that might be due to how its packaged.
i.e. the rpm content
so you need to revert the removal of this.

SeanMooney (dismissed):

templates/nova.conf
auniyal61: ack, so state_path should be default or only nova-compute ?

auniyal61: I think it should be default - because `The top-level directory for maintaining Nova’s state`
from [here ](https://docs.openstack.org/ocata/config-reference/compute/config-options.html)

auniyal61: reverted

SeanMooney: ya so /var/lib/nova should be the default in code but its not for some reaosn i hav enot found my guess is its to do with how we package the rpm

so we want to set it to /var/lib/nova explictly alwasy to get the correct behavior

auniyal61:
recheck
tempest.lib.exceptions.IdentityError: Got identity error

auniyal61:
recheck 
tempest.lib.exceptions.IdentityError: Got identity error
Details: Unexpected status code 503

gibizer:
This shows that we generate [libvirt] config section for the ironic compute config. So that config can be cleaned up too.

templates/nova.conf
gibizer: we pre-generate that file during greenfield so it does exists when the compute starts. However compute should not try to write to it if it exists.

templates/nova.conf
gibizer: do we want state path defined for other than the computes? right now it is generated for at least the nova-api as well: https://review.rdoproject.org/zuul/build/852d057b604c431596707e27c941d80e/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf#5

gibizer:
Here is the edpm compute conf from the test run: https://review.rdoproject.org/zuul/build/852d057b604c431596707e27c941d80e/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-cell1-compute-config.yaml-01-nova.conf

The content looks OK to me, but as a follow up we should improve the whitespacing of the templating as there are a lot of empty lines in the config file which looks strange.

---

I also checked the nova-api service config
https://review.rdoproject.org/zuul/build/852d057b604c431596707e27c941d80e/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf
*disable_fallback_pcpu_query should be scheduler only afaik: https://review.rdoproject.org/zuul/build/852d057b604c431596707e27c941d80e/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf#95-96
* max_concurrent_live_migrations: as far as I know this is compute only. Also the default is already 1 so we don't need to define it explicitly

SeanMooney:
> #826 shows that we generate [libvirt] config section for the ironic compute config. So that config can be cleaned up too.

by the way what is the motivation for this cleaning in general.
i intentionally didn't make these condition originally because you were orgainly worried about the complexity of the template.

we can make them conditional but I'm just wondering why the change of heart?

gibizer:
> > #826 shows that we generate [libvirt] config section for the ironic compute config. So that config can be cleaned up too.
> 
> by the way what is the motivation for this cleaning in general. i intentionally didn't make these condition originally because you were orgainly worried about the complexity of the template.
> 
> we can make them conditional but I'm just wondering why the change of heart?

good point. My goal is not to have misleading / unused config options in the service configs. If this leads to a complex template then I will probably propose to split the template to per service templates.

SeanMooney:
> > > #826 shows that we generate [libvirt] config section for the ironic compute config. So that config can be cleaned up too.
> > 
> > 
> > by the way what is the motivation for this cleaning in general. i intentionally didn't make these condition originally because you were orgainly worried about the complexity of the template.
> > we can make them conditional but I'm just wondering why the change of heart?
> 
> good point. My goal is not to have misleading / unused config options in the service configs. If this leads to a complex template then I will probably propose to split the template to per service templates.

ok because i really don't want to see use split this into multiple templates.
i don't see the generation as misleading even fi that section is not used.

its why we are intentionally not filtering which service config (cinder, neutorn) are generated to only some services.

gibizer:
> > > > #826 shows that we generate [libvirt] config section for the ironic compute config. So that config can be cleaned up too.
> > > 
> > > 
> > > by the way what is the motivation for this cleaning in general. i intentionally didn't make these condition originally because you were orgainly worried about the complexity of the template.
> > > we can make them conditional but I'm just wondering why the change of heart?
> > 
> > 
> > good point. My goal is not to have misleading / unused config options in the service configs. If this leads to a complex template then I will probably propose to split the template to per service templates.
> 
> ok because i really don't want to see use split this into multiple templates. i don't see the generation as misleading even fi that section is not used.
> 
> its why we are intentionally not filtering which service config (cinder, neutorn) are generated to only some services.

The cinder / neutron sections are less likely to cause a misunderstanding as they not expected to be configured by the user much. However something like `force_config_drive` appearing in our nova-api config https://review.rdoproject.org/zuul/build/f777d3130e1446b5bef440d01dde82a4/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf#17 is more problematic in my eyes. If the user looks at our current config they might thing that it needs to be set in the nova-api config, as it is set there today.

auniyal61:
@gibizer 

> *  there are a lot of empty lines in the config file which looks strange.

fixed 

> * disable_fallback_pcpu_query

moved to scheduler.
I could not find it in [docs](https://docs.openstack.org/ocata/config-reference/compute/config-options.html)

> * max_concurrent_live_migrations

removed

templates/nova.conf
SeanMooney: i kind of want this alwasy defiened even if its not used because we are seeing unexpected behavior based on the rpm install.

it not following the upstream default and i cant see where its been overridden in the rpm spec file at a glance os i want to enforce this otherselves just in case we change we add state to any other service in the future.

templates/nova.conf
```diff
@@ -182,19 +181,21 @@ tls_enabled={{ .MemcachedTLS }}
...
 {{ if eq .service_name "nova-compute"}}
 enable_qemu_monitor_announce_self=true
 reserve_disk_resource_for_image_cache=true
-
 # NOTE(gibi): We need this as live migration does not work with
 # cpu_mode=host-model . See https://bugs.launchpad.net/nova/+bug/2039803
 skip_cpu_compare_on_dest = true
-
+{{end}}
 {{end}}
```
auniyal61, line 197: single line OR and AND condition didn't work

auniyal61:
> 
> > * there are a lot of empty lines in the config file which looks strange.
> 
> fixed

there are still lot of empty lines, because of how its rendered.
normaly in jinja2 templates we put condition inside {% %}and expression or statement inside {{ }}, but here everything is inside {{ }}, so right now its just printing "\S" (space) on every failed condition.

as of now, I tried to use {% %}, but its just printing the whole thing and not parsing it as condition, so no change there.

another way I tried for my tests is to refactor configData string to replace "\n\n\n+" with "\n", before writing to a final file, and it did the what we want, but this may not be a best way.

SeanMooney:
> > > * there are a lot of empty lines in the config file which looks strange.
> > 
> > 
> > fixed
> 
> there are still lot of empty lines, because of how its rendered. normaly in jinja2 templates we put condition inside {% %}and expression or statement inside {{ }}, but here everything is inside {{ }}, so right now its just printing "\S" (space) on every failed condition.
> 
> as of now, I tried to use {% %}, but its just printing the whole thing and not parsing it as condition, so no change there.
> 
> another way I tried for my tests is to refactor configData string to replace "\n\n\n+" with "\n", before writing to a final file, and it did the what we want, but this may not be a best way.

we are not using jinia templating this is golang templating.

there is a way to trim using -  i.e. 

{{- -}} i belive see https://pkg.go.dev/text/template#hdr-Text_and_spaces for details.

SeanMooney:
i woudl perfer to not include removing of extra white space in this pr

templates/nova.conf
```diff
@@ -1,9 +1,10 @@
 [DEFAULT]
 # concurrent live migrations are more likely to fail and are slower
 # overall then serializing live migrations so set this to 1 explictly
-max_concurrent_live_migrations=1
```
gibizer, line 4: as the config option is dropped the related comment needs to be dropped too as otherwise the resulting config file will be confusing:
https://review.rdoproject.org/zuul/build/e50fccd89f48499787f27fda89cc4a65/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-cell1-compute-config.yaml-01-nova.conf#2-4

gibizer (changes requested):
I have one more nit to fix otherwise the generated config files looks good to me.
Also I'm OK to try to remove the excessive whitespaces from the generated config files in a separate commit.

templates/nova.conf
SeanMooney: we should not drop this.
It is the upstream default but I really want us to set this explicitly

gibizer: OK, lets keep it explicit.

auniyal61: reverted, max_concurrent_live_migrations done

auniyal61:
/retest
`Nova reconfiguration [BeforeEach] reconfigures memcached service and check` works locally

auniyal61:
/retest
`Samples when nova_v1beta1_nova-compute-fake.yaml sample is applied`  passed locally
```
Ran 11 of 323 Specs in 8.195 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 312 Skipped
```

mrkisaolamb:
I rebased patch to contain fix for unstable tests, this should stabilize reconfiguration tests like "Nova reconfiguration [BeforeEach] reconfigures memcached service and check"

mrkisaolamb:
recheck

auniyal61:
/retest 
`nova_v1beta1_nova-compute-fake.yaml sample ` test works locally

mrkisaolamb (approved):
Looks like CI hiccups on timeouts in functionall test ended. I'm ok with this changes but @gibizer -1 so it will be nice to have his approve

gibizer (approved):
looks good thanks.

auniyal61:
/retest pod pending timeout, seems deployment issue
`/logs/process-log.txt: no such file or directory
`

PR #825, opened 2024-07-18
Disable cpu power management by default

gibizer:
We have multiple issue with power management so we disable it again.

Related: https://issues.redhat.com/browse/OSPRH-8806
Related: https://issues.redhat.com/browse/OSPRH-8712

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -209,7 +209,10 @@ tx_queue_size=512
 swtpm_enabled=True
 volume_use_multipath=true
 live_migration_uri = qemu+ssh://nova@%s/system?keyfile=/var/lib/nova/.ssh/ssh-privatekey
-cpu_power_management=true
+# We can only re-enable it when the following Jiras are fixed:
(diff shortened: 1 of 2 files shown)


gibizer:
recheck we need to spend time stabilizing the local storage job as it fails a lot intermittently recently.

gibizer:
recheck

gibizer:
recheck

gibizer:
now that #828 merge we have a chance to get a green CI here.

gibizer:
/unhold

We agreed over Slack that due to the topology issue causing NUMA resource accounting issues even if we fix the governor issue quickly we cannot go out with power mgmt enabled by default.

SeanMooney (approved):
approving this based on the proximity to ga

we will move the default to on to bugfix 1 or later once the upstream bugs are resolved and backported

we will also document in a know issue releasnote how this fucntioantly can be used safely without the bugfix via the govoner policy.

we do not want to default to the governer policy because it has an inherent upgrade impact.

we can enable this by default in a later release.

PR #827, opened 2024-07-19
[zuul]Use crc 2.39 that has ocp 4.16

gibizer:

PR #828, opened 2024-07-21
[zuul]Pin local storage job to 4.15

gibizer:
It seems the base job is moved to ocp 4.16 and since then the job fails
to deploy a stable crc. Pinning back to 4.15 while we separately working
on making 4.16 based job green.

Trying 4.16 in https://github.com/openstack-k8s-operators/nova-operator/pull/827


mrkisaolamb (approved):
lgtm

PR #833, opened 2024-07-24
[DNM]Test cross repo build

gibizer:
Depends-On: https://review.opendev.org/c/openstack/nova/+/924823

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -13,7 +13,7 @@
     parent: cifmw-podified-multinode-edpm-base-crc
     abstract: true
     attempts: 1
-    dependencies: ["openstack-k8s-operators-content-provider"]
+    dependencies: ["openstack-meta-content-provider"]
     required-projects:
       - github.com/openstack-k8s-operators/ci-framework
       - github.com/openstack-k8s-operators/install_yamls
@@ -27,7 +27,7 @@
 - job:
     name: nova-operator-kuttl
     parent:  nova-operator-base
-    dependencies: ["openstack-k8s-operators-content-provider"]
+    dependencies: ["openstack-meta-content-provider"]
     roles:
       - zuul: github.com/openstack-k8s-operators/ci-framework
     description: |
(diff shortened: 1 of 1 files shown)


gibizer:
Testing with an intentionally broken openstack/nova change to see if our tests really running with a nova-compute built from that upstream change

SeanMooney:
im glad you tried this. 

the comptues are indeed not failing in init_host
but unfortunately your using [openstack-k8s-operators-content-provider](https://review.rdoproject.org/zuul/stream/b0101965a9af41388f1e802f71ba84af?logfile=console.log)

not the meta provider so that one is not building new nova images only operator images.

gibizer:
> im glad you tried this.
> 
> the comptues are indeed not failing in init_host but unfortunately your using [openstack-k8s-operators-content-provider](https://review.rdoproject.org/zuul/stream/b0101965a9af41388f1e802f71ba84af?logfile=console.log)
> 
> not the meta provider so that one is not building new nova images only operator images.

Ahh you are right :/ Let me fix it :)

gibizer:
Looked at the results and the computes are failing with :
```
Jul 24 08:14:53 np0004809099 nova_compute[153711]: 2024-07-24 12:14:53.909 2 ERROR oslo_service.service     raise ValueError("!!! test")

```
as expected due to the depends-on patch.

I will do one more round breaking an service running on the ocp side to see that also takes effect.

gibizer:
recheck testing with PS2 of the openstack/nova patch that will kill the scheduler.

gibizer:
recheck both jobs failed on rabbit connection issue instead of failing tempest tests due to the scheduler breaking patch we depending on.

gibizer:
OK now the nova-scheduler failed due to the patch we are depending on:
```
2024-07-29 11:31:27.435 1 ERROR oslo_messaging.rpc.server [None req-b84eb033-3718-4209-bd7a-e229c33d3b1b efcd56eb2a0a4fb185b41b9f80ae5ac5 e2adc3c6aaa84528b7b5bee012dbafaf - - default default] Exception during message handling: ValueError: !!! test
```

bottom line depending on openstack/nova patches actually works if we are using the meta content provider. \o/

PR #834, opened 2024-07-24
Handle simulating conductor readiness

mrkisaolamb:
Looks like during some async operations we
can simulate readines too fast. To avoid this issue we
should handle this in eventually statement


gibizer (commented):
Thanks @mrkisaolamb for making the job stable . Still we need to go back, understand and refactor how we simulate readyness as this is now all over the place and I'm afraid we will keep chasing similar issues as soon as a new change will be needed in the controllers.

PR #846, opened 2024-08-19
Add deleting cells automation

mrkisaolamb:
close: [OSPRH-105](https://issues.redhat.com//browse/OSPRH-105)


mrkisaolamb:
Still require more testing and adding more func/kuttle tests

templates/nova-manage/bin/delete_cell.sh
```diff
@@ -0,0 +1,29 @@
...
+# NOTE(gibi): nova-manage should be enhanced upstream to get rid of this
+# uglyness
+# Note the "|" around the CELL_NAME, that is needed as a single line from
+# nova-manage cell_v2 cell_list can match to multiple cells if the cell name
+# is part of the line, e.g. as the user name of the DB URL
+cell_uuid=$(nova-manage cell_v2 list_cells | tr ' ' '|' | tr --squeeze-repeats '|' | grep -e "^|$CELL_NAME|" | cut -d '|' -f 3)
+
+if [ -z "${cell_uuid}" ]; then
+    nova-manage cell_v2 delete_cell --cell_uuid "${cell_uuid}"
+fi
```
SeanMooney, line 29: i guess we need to do this because we don't include the cell uuid in our statutes and use the name 

i think that's ok in General.

templates/nova-manage/config/cell-delete-config.json
```diff
@@ -0,0 +1,36 @@
+{
+    "command": "/bin/ensure_cell_mapping.sh",
```
gibizer, line 2: I guess your intention was to call delete_cell.sh from the delete job.

templates/nova-manage/bin/delete_cell.sh
gibizer: Yeah we either needs to bubble up the celll uuid to the Nova Status, which is uggly as it would need log parsing.
Or we need to improve nova-manage to work with cell names as well. 

Until any of them this is acceptable to me.

gibizer (changes requested):
I tried to use this locally but it fails with:
```
 ❯ oc logs nova-cell1-cell-delete-tw8ts
+ sudo -E kolla_set_configs
sudo: unable to send audit message: Operation not permitted
INFO:__main__:Loading config file at /var/lib/kolla/config_files/config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Copying service configuration files
INFO:__main__:Deleting /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/nova-blank.conf to /etc/nova/nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/01-nova.conf to /etc/nova/nova.conf.d/01-nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf.d/01-nova.conf
INFO:__main__:Copying /var/lib/openstack/bin/delete_cell.sh to /bin/delete_cell.sh
INFO:__main__:Setting permission for /bin/delete_cell.sh
INFO:__main__:Copying /var/lib/openstack/config/my.cnf to /etc/my.cnf
INFO:__main__:Setting permission for /etc/my.cnf
INFO:__main__:Writing out command to execute
++ cat /run_command
+ CMD=/bin/ensure_cell_mapping.sh
+ ARGS=
+ sudo kolla_copy_cacerts
sudo: unable to send audit message: Operation not permitted
+ [[ ! -n '' ]]
+ . kolla_extend_start
+ echo 'Running command: '\''/bin/ensure_cell_mapping.sh'\'''
Running command: '/bin/ensure_cell_mapping.sh'
+ umask 0022
+ exec /bin/ensure_cell_mapping.sh
/usr/local/bin/kolla_start: line 22: /bin/ensure_cell_mapping.sh: No such file or directory
```

mrkisaolamb:
recheck

gibizer:
Tried this again locally by deleting cell1.

I see two issues:
* the cell_mapping is not deleted 
```
❯ oc logs nova-cell1-cell-delete-t8ddb
+ sudo -E kolla_set_configs
sudo: unable to send audit message: Operation not permitted
INFO:__main__:Loading config file at /var/lib/kolla/config_files/config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Copying service configuration files
INFO:__main__:Deleting /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/nova-blank.conf to /etc/nova/nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf
INFO:__main__:Copying /var/lib/openstack/config/01-nova.conf to /etc/nova/nova.conf.d/01-nova.conf
INFO:__main__:Setting permission for /etc/nova/nova.conf.d/01-nova.conf
INFO:__main__:Copying /var/lib/openstack/bin/delete_cell.sh to /bin/delete_cell.sh
INFO:__main__:Setting permission for /bin/delete_cell.sh
INFO:__main__:Copying /var/lib/openstack/config/my.cnf to /etc/my.cnf
INFO:__main__:Setting permission for /etc/my.cnf
INFO:__main__:Writing out command to execute
++ cat /run_command
+ CMD=/bin/delete_cell.sh
+ ARGS=
+ sudo kolla_copy_cacerts
sudo: unable to send audit message: Operation not permitted
+ [[ ! -n '' ]]
+ . kolla_extend_start
+ echo 'Running command: '\''/bin/delete_cell.sh'\'''
Running command: '/bin/delete_cell.sh'
+ umask 0022
+ exec /bin/delete_cell.sh
+ export CELL_NAME=cell1
+ CELL_NAME=cell1
++ nova-manage cell_v2 list_cells
++ tr ' ' '|'
++ tr --squeeze-repeats '|'
++ cut -d '|' -f 3
++ grep -e '^|cell1|'
+ cell_uuid=30e83114-9261-4476-bfa4-e1301953e85f
+ '[' -z 30e83114-9261-4476-bfa4-e1301953e85f ']'
 ```

```
❯ oc rsh nova-cell0-conductor-0 nova-manage cell_v2 list_cells
+-------+--------------------------------------+-----------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------+----------+
|  Name |                 UUID                 |                                      Transport URL                                      |                                             Database Connection                                             | Disabled |
+-------+--------------------------------------+-----------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------+----------+
| cell0 | 00000000-0000-0000-0000-000000000000 |                                         rabbit:                                         |    mysql+pymysql://****:****@openstack.openstack.svc/nova_cell0?read_default_file=/etc/my.cnf    |  False   |
| cell1 | 30e83114-9261-4476-bfa4-e1301953e85f | rabbit://****:****@rabbitmq-cell1.openstack.svc:5671/?ssl=1 | mysql+pymysql://****:****@openstack-cell1.openstack.svc/nova_cell1?read_default_file=/etc/my.cnf |  False   |
+-------+--------------------------------------+-----------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------+----------+
```
* MariaDBAccount and MariaDBDatabase is not deleted
```
❯ oc get MariaDBAccount | grep nova
nova-api     18h
nova-cell0   18h
nova-cell1   18h
❯ oc get MariaDBDatabase | grep nova
nova-api     18h
nova-cell0   18h
nova-cell1   18h
 
❯ oc get MariaDBAccount/nova-cell1 -o yaml
apiVersion: mariadb.openstack.org/v1beta1
kind: MariaDBAccount
metadata:
  creationTimestamp: "2024-09-02T15:04:44Z"
  finalizers:
  - openstack.org/nova
  - openstack.org/mariadbaccount
  generation: 1
  labels:
    mariaDBDatabaseName: nova-cell1
  name: nova-cell1
  namespace: openstack
 
 
❯ oc get MariaDBDatabase/nova-cell1 -o yaml
apiVersion: mariadb.openstack.org/v1beta1
kind: MariaDBDatabase
metadata:
  creationTimestamp: "2024-09-02T15:04:44Z"
  finalizers:
  - openstack.org/nova
  - openstack.org/mariadbdatabase
  - openstack.org/mariadbaccount-nova-cell1
  generation: 1
  labels:
    dbName: openstack-cell1
  name: nova-cell1
  namespace: openstack
```

File: templates/nova-manage/bin/delete_cell.sh [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+export CELL_NAME=${CELL_NAME:?"Please specify a CELL_NAME variable."}
+
+# NOTE(gibi): nova-manage should be enhanced upstream to get rid of this
+# uglyness
+# Note the "|" around the CELL_NAME, that is needed as a single line from
+# nova-manage cell_v2 cell_list can match to multiple cells if the cell name
+# is part of the line, e.g. as the user name of the DB URL
+cell_uuid=$(nova-manage cell_v2 list_cells | tr ' ' '|' | tr --squeeze-repeats '|' | grep -e "^|$CELL_NAME|" | cut -d '|' -f 3)
+
+if [ -z "${cell_uuid}" ]; then
```
gibizer, line 27: -z it true if the variable is empty, I think this is why the cell mapping is not deleted.

controllers/nova_controller.go
```diff
@@ -591,10 +591,94 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	job := job.NewJob(
+		jobDef, cell.Name+"-cell-delete",
+		instance.Spec.PreserveJobs, r.RequeueTimeout,
+		instance.Status.RegisteredCells[cell.Name])
+
+	_, err = job.DoJob(ctx, h)
+	if err != nil {
+		return err
+	}
+
```
gibizer, line 677: We need to also delete the TransportURL, MariaDBAccount and the MariaDBDatabase CRs related to the deleted cell.

gibizer: We also have leftover secrets:
```
nova-cell1                                             Opaque                    2      18h
nova-cell1-db-secret                                   Opaque                    1      18h
nova-cell1-manage-config-data                          Opaque                    6      18h
nova-cell1-manage-scripts                              Opaque                    3      18h
```

These secrets owned by Nova CR this is why they are not garbage collected.

gibizer (changes requested):
I found more leftovers. See inline.

gibizer:
Also we need to call cleanServiceFromNovaDb in the reconcileDelete of the NovaConductor to ensure that the nova service is also deleted form the nova api DB.

templates/nova-manage/bin/delete_cell.sh
mrkisaolamb: done

controllers/nova_controller.go
mrkisaolamb: done, also switch manage secret to cell0

mrkisaolamb:
Still need to work with cert deleting for cell services, I will update this case in next comments

test/kuttl/test-suites/default/cell-delete-tests/02-assert.yaml
```diff
@@ -0,0 +1,126 @@
...
+    ready: true
+    started: true
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - script: |
+      set -euxo pipefail
+      MESSAGE=$(oc exec -n $NAMESPACE openstackclient -- openstack flavor create my-flavor 2>&1) || true
```
gibizer, line 125: this seems to be a copy paste .

mrkisaolamb: done and also 01 assert should be correct now

templates/nova-manage/bin/delete_cell.sh
```diff
@@ -0,0 +1,27 @@
...
+
+export CELL_NAME=${CELL_NAME:?"Please specify a CELL_NAME variable."}
+
+# NOTE(gibi): nova-manage should be enhanced upstream to get rid of this
+# uglyness
+# Note the "|" around the CELL_NAME, that is needed as a single line from
+# nova-manage cell_v2 cell_list can match to multiple cells if the cell name
+# is part of the line, e.g. as the user name of the DB URL
+cell_uuid=$(nova-manage cell_v2 list_cells | tr ' ' '|' | tr --squeeze-repeats '|' | grep -e "^|$CELL_NAME|" | cut -d '|' -f 3)
+
```
auniyal61, line 26: `nova-manage cell_v2 list_cells | grep $CELL_NAME | awk '{print $4}'`

gibizer:
Still the same tempest failure. So I guess we need to wait a bit more to get the new image in the pipe.

https://softwarefactory-project.io/zuul/t/rdoproject.org/build/8f92ed3d2834422ba21eaa731e89282f/log/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tempest-tests-n4gd2/logs/tempest-tests-tests-runner.log#225

gibizer:
recheck we expect that the new tcib image is now built by the periodic job

gibizer (commented):
commit 3 in the history is a mix of many things not just kuttl tests. Could you move the controller and envtest pieces to their respective commits?

.zuul.yaml
```diff
@@ -72,6 +72,54 @@
                 ip: 172.18.0.5
               tenant:
                 ip: 172.19.0.5
+- job:
+    name: nova-operator-kuttl-cell-delete
```
gibizer, line 76: Why do we need a separate kuttl job for this scenario? I feel if this scenario cannot be run alongside the other scenarios then that is bug and moving it to a separate job just papering over the issue

gibizer (changes requested):

mrkisaolamb:
Done, I untangled commit 3. The reason I created a separate job was that I didn't want to complicate our playbooks for the nova-kuttle job too much. However, I can try to make it less invasive and make one job with all tests.

templates/nova-manage/bin/delete_cell.sh
gibizer: As this hack is now copied from cell creation to cell deletion too I filed a Jira to eventually improve this upstream https://issues.redhat.com/browse/OSPRH-10284

controllers/novacell_controller.go
```diff
@@ -283,6 +289,44 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+		}
+	}
+
+	dbName, accountName := novaapi.ServiceName+"-"+instance.Spec.CellName, instance.Spec.CellDatabaseAccount
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, h, dbName, accountName, instance.ObjectMeta.Namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+	if !k8s_errors.IsNotFound(err) {
+		if err := db.DeleteFinalizer(ctx, h); err != nil {
```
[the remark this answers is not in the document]
gibizer: MariaDBDatabse and Account only has nova's finalizer

controllers/novacell_controller.go
```diff
@@ -283,6 +289,44 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+	err = r.ensureConductorDeleted(ctx, instance)
+	if err != nil {
+		return err
+	}
+	for computeName := range instance.Status.NovaComputesStatus {
+		err = r.ensureNovaComputeDeleted(ctx, instance, computeName)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
gibizer: I just tested it locally with main. If I remove the cell1 from OpenStackControlPlane and then I simulate by hand that nova-operator deletes the NovaCell/nova-cell1 CR, then eventually the NovaNovncProxy and Conductor CRs is deleted automatically by the k8s garbage collector. So now I'm sure we don't need these explicit deletes.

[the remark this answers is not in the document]
gibizer: The explicit delete of the NovaCell CR in the nova-controller is still needed as the Nova CR is not deleted during cell deletion.

rabi:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

test/functional/base_test.go
```diff
@@ -214,6 +214,14 @@ func NovaConductorConditionGetter(name types.NamespacedName) condition.Condition
 	return instance.Status.Conditions
 }
 
+func AssertConductorDoesNotExist(name types.NamespacedName) {
```
gibizer, line 217: I guess this is unused now

test/functional/nova_reconfiguration_test.go
```diff
@@ -168,6 +170,37 @@ var _ = Describe("Nova reconfiguration", func() {
...
+				g.Expect(newJobInputHash).NotTo(BeNil())
+			}, timeout, interval).Should(Succeed())
+			th.AssertSecretDoesNotExist(cell1.InternalCellSecretName)
+
+			Eventually(func(g Gomega) {
+				instance := &rabbitmqv1.TransportURL{}
+				err := k8sClient.Get(ctx, cell1.TransportURLName, instance)
+				g.Expect(k8s_errors.IsNotFound(err)).To(BeTrue())
+			}, timeout, interval).Should(Succeed())
+		})
```
gibizer, line 202: you can also assert that the hash is removed from the registeredcells from the Nova.Status.

mrkisaolamb: hash is asserted above

mrkisaolamb: it is asserted https://github.com/openstack-k8s-operators/nova-operator/pull/846/files#diff-5b5adf7c31e2c615eaf3d65ba102b7e6b3af0c8f7222e83402e731a27bb86a8dR185

test/kuttl/test-suites/default/tests/02-assert.yaml
```diff
@@ -0,0 +1,125 @@
...
+  - name: nova-kuttl-metadata-metadata
+    ready: true
+    started: true
+
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+namespaced: true
+commands:
+  - command: echo oc get -n $NAMESPACE pod/nova-kuttl-cell1-novncproxy-0 --ignore-not-found
```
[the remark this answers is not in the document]
mrkisaolamb: Done, now it's working inside kuttle job

mrkisaolamb:
recheck

mrkisaolamb:
Fail in func test look like something random, also created improvement for db account deletion: https://issues.redhat.com/browse/OSPRH-10406

test/functional/nova_reconfiguration_test.go
gibizer: true. sorry.

test/functional/base_test.go
mrkisaolamb: done, thanks!

controllers/nova_controller.go
```diff
@@ -591,10 +591,212 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		inputHash)
+
+	_, err = job.DoJob(ctx, h)
+	if err != nil {
+		return err
+	}
+
+	// Delete secrets
+	dbSecret := fmt.Sprintf("%s-db-secret", cell.Name)
+	err = secret.DeleteSecretsWithName(ctx, h, dbSecret, instance.Namespace)
```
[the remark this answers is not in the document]
gibizer: Now my local testing shows that this is probably a mistake and we should only remove the finalizer and let the mariadb-operator do the secret deletion when MariadbAccount is deleted.

controllers/nova_controller.go
```diff
@@ -591,10 +591,206 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	}
+
+	// also do a delete for "unused" MariaDBAccounts, associated with
+	// this MariaDBDatabase.
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(
+		ctx, h, name, accountName, namespace,
+	)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
```
gibizer, line 685: we can call this but it won't delete the current account as it is still considered used as we are passing the name of it.

However we do need to delete the MariaDBAccount otherwise we cannot delete the MariaDBDatabase.

controllers/nova_controller.go
```diff
@@ -591,10 +591,206 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	} else if err == nil && controllerutil.RemoveFinalizer(mariaDBDatabase, h.GetFinalizer()) {
+		err := h.GetClient().Update(ctx, mariaDBDatabase)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return err
+		}
+		util.LogForObject(h, fmt.Sprintf("Removed finalizer %s from MariaDBDatabase %s", h.GetFinalizer(), mariaDBDatabase.Spec.Name), mariaDBDatabase)
+	}
+
```
gibizer, line 697: We need to delete the MariaDBDatabase as Nova created it in the first place for the cell.

controllers/nova_controller.go
```diff
@@ -591,10 +591,206 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if err != nil {
+		return err
+	}
+
+	secretName := getNovaCellCRName(instance.Name, cellName)
+	err = secret.DeleteSecretsWithName(ctx, h, secretName, instance.Namespace)
+	if err != nil {
+		return err
+	}
+
```
gibizer, line 771: we need to delete the cell1 manage script and config secrets (that was used for the db sync during cell creation)

gibizer (changes requested):

gibizer (dismissed):

gibizer:
I will prepare an extra commit fixing my comments as @mrkisaolamb is on a longer PTO and we need to progress. But first the knowledge I gathered while troubleshooting this.
* the MariaDBAccount handling is complex and therefore it was a source of lot of confusion. There are in flight commit changing it but we need to review those carefully based on the below learnings.
* the MariaDBAccount is either passed in to Nova or created by Nova as a fallback by calling mariadb-operator api functions. However the two case is not really separated, so during delete it is hard to tell if nova needs to delete the Account CR as it created it in the first place or should not delete it as it was passed in and the user created it.
* The DB secret connected to the MariaDBAccont CR has the same fallback, if it is not exists the nova creates it as part of the mariadb-operator/api calls. But it can also passed in. Again during deletion, if MariaDBAccount is deleted the mariadb-operator cannot decide if the connected Secret needs to be deleted as it was autogenerated, or needs to be kept as it is user provided. The maridb-operator simply keeps the secret today.
* In reality today the account is never passed in to Nova so both the Account and the connected secret is generated by nova via the mariadb-operator/apis function. Therefore during cell deletion both the Account and the secret needs to be cleaned up. I will push a local solution for it, but it needs to be moved to mariadb-operator/apis eventually.

Additional things that we need to document:
* nova will delete the MariaDBDatabase/nova-cellx CR during cell deletion but it does not mean the underlying database is removed from the disk of the Galera instance. It can be considered a data safety feature to avoid real data loss in a unintentional cell deletion. It also means if cellx is recreated later with the same Galera instance then cell DB data is simply resurrected. This might not what the user want in an intentional cell delete and recreate. So we need to document that in such case the DB manually needs to be dropped via SQL command on the galerea pod.
* As there might be host mappings in the nova_api database for the deleted cells we also need to document that a cell can only be deleted if all the computes in that cell is scaled-in first to trigger host_mapping cleanup. We can potentially implement a feature where the NodeSet using the nova-operator generated nova-cell1-compute secret puts finalizers on the secret and nova-operator can check that and refuse to delete if there are finalizers left on the secret while the cell is being deleted.

I also confirm that after the cell deletion there are leaked cert secrets for the cell:
```
cert-nova-novncproxy-cell1-public-route              kubernetes.io/tls         3      6m7s
cert-nova-novncproxy-cell1-public-svc                kubernetes.io/tls         3      6m8s
cert-nova-novncproxy-cell1-vencrypt                  kubernetes.io/tls         3      6m6s
```
These are created and owned by the openstack-operator so we need a PR there to clean these up during the cell deletion.

gibizer:
Some local test results with cell delete and recreate to show that the DB is not recreated but resurrected from disk https://paste.opendev.org/show/bGe8LZZjM9Onak38mffe/

pkg/nova/celldelete.go
```diff
@@ -0,0 +1,86 @@
...
+							VolumeMounts: volumeMounts,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	return job
+}
```
SeanMooney, line 86: i got to about here before meeting. ill skim the rest quickly but i think ill wait until this is updated to look again.

i may or may not have time to test this next week. but mostly this all seams reasonable

templates/nova-manage/config/cell-delete-config.json
```diff
@@ -0,0 +1,36 @@
...
+            "owner": "nova",
+            "perm": "0600"
+        },
+        {
+            "source": "/var/lib/openstack/config/02-nova-override.conf",
+            "dest": "/etc/nova/nova.conf.d/02-nova-override.conf",
+            "owner": "nova",
+            "perm": "0600",
+            "optional": true
+        },
```
SeanMooney, line 22: as an aside is there ever really a reason to allow user overrides when creating static jobs like this.

in theory nothing they can set should affect anything.


we can do this for consistency with the other jobs as i don't know if i thought about that before but for delete thei is probably not needed.

SeanMooney (commented):
i have not looked at the testing but more just did a shallow pass to refamilarise myself a bit with this code

controllers/nova_controller.go
```diff
@@ -591,10 +591,253 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	} else if err == nil {
+		if databaseAccount.Spec.Secret != "" {
+			dbSecret, _, err := secret.GetSecret(ctx, h, databaseAccount.Spec.Secret, namespace)
+			if err != nil && !k8s_errors.IsNotFound(err) {
+				return err
+			}
+
+			if err == nil && controllerutil.RemoveFinalizer(dbSecret, h.GetFinalizer()) {
```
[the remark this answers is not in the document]
zzzeek: it's just a code pattern that's been copied along for some time.   I've been annoyed at Go's silly insistence on "try/except is bad" since the language was introduced but here we are, wasting time on how to structure conditionals around parsing error codes.

whatever we want to change this to, I can adjust in the originating PR on the mariadb API side

controllers/nova_controller.go
```diff
@@ -591,10 +591,253 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.Namespace),
+	}
+	if err := r.Client.List(ctx, novaCellList, listOpts...); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	for _, cr := range novaCellList.Items {
+		_, ok := instance.Spec.CellTemplates[cr.Spec.CellName]
+		if !ok {
```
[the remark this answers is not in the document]
gibizer: I'm in the situation where I need to make progress so while I agree that it would be a meaningful change I will not do it now. When Kamil is back he can do a follow up to refactor this.

controllers/nova_controller.go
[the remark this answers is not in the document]
gibizer: @SeanMooney 

> This feel weird to me
> 
> ```
> if err != nil && !k8s_errors.IsNotFound(err) {
> 				return err
> 			}
> ```
> 
> so after this "err is nil" or it holds an error that is not "IsNotFound"
> 
> but we then use dbSecret even if err hold an error meaning dbSecret might not be safe to use? dbSecret could be nil right or likely would be if a different error happened.
> 
> i feel like if we are ever checking `err == nil` that might be a code smell or possible bug.

The logic we need:
* if GetSecret found the Secret then remove the finalizer from it
* if GetSecret not found the Secret then we know that the finalizer is already removed, so just move forward but do not try to remove it again as the Secret does not exists.
* if GetSecret encountered an error (other than not found handled above) the propagate the error up.

> but we then use dbSecret even if err hold an error meaning dbSecret might not be safe to use? dbSecret could be nil right or likely would be if a different error happened.

We don't use the dbSecret if err contains any kind of error, this is why we have err == nil check.

templates/nova-manage/config/cell-delete-config.json
gibizer: yeah that is probably not necessary in the cell handling jobs. As it is already exists in other cell handling jobs I will not change it here but I will add a task to the Jira to do a follow up to clean it up everywhere.

gibizer (commented):
I will try to replace the vendored DeleteDatabaseAndAccountFinalizers with the direct call to mariadb-database/api and test that the clenaup still works.

controllers/nova_controller.go
```diff
@@ -591,10 +591,253 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		util.LogForObject(h, "Deleted MariaDBAccount", account)
+	}
+
+	// NOTE(gibi): This is also controversial as we don't know if the account
+	// secret was created by us by falling back to create the MariaDBAccount.
+	// Still MariaDBAccount cleanup does not delete the secret so we have to.
+	err = secret.DeleteSecretsWithName(ctx, h, account.Spec.Secret, namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
```
gibizer, line 712: this is a divergence from https://github.com/openstack-k8s-operators/mariadb-operator/pull/268 so I need to move that out of DeleteDatabaseAndAccountFinalizers before I replace it with the mariadb-operator version

controllers/nova_controller.go
```diff
@@ -591,10 +591,253 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+					"Removed finalizer %s from MariaDBDatabase %s", h.GetFinalizer(), mariaDBDatabase.Spec.Name),
+				mariaDBDatabase)
+		}
+
+		err := h.GetClient().Delete(ctx, mariaDBDatabase)
+		if err != nil {
+			return err
+		}
+		util.LogForObject(h, "Deleted MariaDBDatabase", mariaDBDatabase)
+	}
```
gibizer, line 735: This also needs to be moved out

controllers/nova_controller.go
```diff
@@ -591,10 +591,253 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	fullCellName := types.NamespacedName{
+		Name:      getNovaCellCRName(instance.Name, cellName),
+		Namespace: instance.GetNamespace(),
+	}
+
+	err := r.Client.Get(ctx, fullCellName, cell)
+	if k8s_errors.IsNotFound(err) {
+		// Nothing to do as it does not exists
+		return nil
+	}
```
gibizer, line 762: this feels like a bug. It is possible that a previous reconcile deleted the Cell but failed to clean up the other bits we clean up below. So we cannot return here, we need to skip the cell delete but we still need to try to delete rest (transport, db, db account, etc)

controllers/nova_controller.go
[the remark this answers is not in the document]
SeanMooney: yep thats fin ei think in the long run we may want to avoid failing fast here and make as much forward progress as we can but i agree its out of scope of FR1 so ill resolve this thread.

controllers/nova_controller.go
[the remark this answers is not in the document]
SeanMooney: i guess we are relying on short cuirit evaluation of the `&&` operator

```
err == nil && controllerutil.RemoveFinalizer(dbSecret, h.GetFinalizer())
```

my main concern with this pattern is we do not log the error or return it when err != nil in this case.

the err will only be non nil if its a not found error.
so we may want to log that for debug reasons but i was mostly concerned about potentially not handling the err properly.

given it can only be not found and this is a delete code path i think that is fine.

gibizer:
* rebased on main
* removed the vendored change and calling https://github.com/openstack-k8s-operators/mariadb-operator/pull/268 directly
* fixed potential bugs when cleanup needs to be retried
* refactored a bit.

controllers/nova_controller.go
gibizer: Actually we cannot delete DB and Account without having the NovaCell CR as the name of the account is only stored there and in the cellTemplate but the cellTemplate is deleted by the user.

controllers/nova_controller.go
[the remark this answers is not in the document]
gibizer: this is now hidden in the mariadb-operator :)

gibizer (commented):
Opened trackers for the remaining items

gibizer:
the mariadb-operator dependency now landed, so for me this is ready to go.

zzzeek:
I'm assuming the nova cells use case is unusual enough that the routines here are still needing to deal with MariaDBAccount and MariaDBDatabase explicitly, rather than using the new DeleteDatabaseAndAccountFinalizers routine, is that correct?

gibizer:
> I'm assuming the nova cells use case is unusual enough that the routines here are still needing to deal with MariaDBAccount and MariaDBDatabase explicitly, rather than using the new DeleteDatabaseAndAccountFinalizers routine, is that correct?

Yes. We do call DeleteDatabaseAndAccountFinalizers but we also need to explicitly delete the MariaDBDatabase CR afterwards as it is owned by the Nova CR but the Nova CR is not going away, just the NovaCell CR, so we cannot rely on the k8s garbage collector to delete the MariaDBDatabase.

SeanMooney (approved):

PR #854, opened 2024-09-02
adjust job template to account for master vs main

SeanMooney:
this change makes the opendev job template override
the checkout of all required porject to main to account
for the fact that the job will run on master.


SeanMooney:
i am testing this with https://review.opendev.org/c/openstack/nova/+/927762

we currently have a isue with the cloud provider backing zuul so we need to wait to retrigger this. once its back we can trigger this from gerrit and confirm it works end to end
if that runes properly we can proceed with merging this.

raukadah:
LGTM!
I can see jobs are appearing in RDO Zuul openstack-experimental pipeline.

PR #862, opened 2024-09-17
use DeleteDatabaseAndAccountFinalizers

zzzeek:
This switches the use of GetDatabaseAndAccount +
database.DeleteFinalizers() to use a new function in mariadb DeleteDatabaseAndAccountFinalizers, which looks up each individual component used for database access individually and removes finalizers from all of those that were found, regardless of other elements not
existing.    All controllers should be changed to use this method to
ensure that early-created MariaDBAccount objects are removed from a
control plane where MariaDBDatabase objects did not get created before
the control plane were deleted, as well as similar scenarios.

relies upon https://github.com/openstack-k8s-operators/mariadb-operator/pull/268/

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/openstack-k8s-operators/lib-common/modules/common v0.4.1-0.20240926101719-8fc1c3da53f7
 	github.com/openstack-k8s-operators/lib-common/modules/openstack v0.4.1-0.20240926101719-8fc1c3da53f7
 	github.com/openstack-k8s-operators/lib-common/modules/test v0.4.1-0.20240926101719-8fc1c3da53f7
-	github.com/openstack-k8s-operators/mariadb-operator/api v0.4.1-0.20240927143624-61d230f582d6
+	github.com/openstack-k8s-operators/mariadb-operator/api v0.4.1-0.20241008102252-6f4b7a67029f
 	github.com/openstack-k8s-operators/nova-operator/api v0.0.0-20221209164002-f9e6b9363961
 	go.uber.org/zap v1.27.0
 	gopkg.in/yaml.v3 v3.0.1
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -1269,14 +1269,10 @@ func (r *NovaReconciler) ensureDBDeletion(
 	// iterate over novaDbs and remove finalizers
 	for _, novaDb := range novaDbs {
 		dbName, accountName := novaDb[0], novaDb[1]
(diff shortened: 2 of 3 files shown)


auniyal61:
this is a [DNM] patch. as its targeting a unmerged patch/build

because the depends-on mariadb patch did not merged yet [here ](https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/api/v1beta1/mariadbdatabase_funcs.go)

okay its defined [here ](https://github.com/openstack-k8s-operators/mariadb-operator/pull/268/files#diff-7c08d2fcf1f0026acea18c6a06272c6a21e31af4bbebb8b6bebcaf72e515b21e)


though `DeleteDatabaseAndAccountFinalizers` additionaly  delete `"unused" MariaDBAccounts`
while `DeleteFinalizer` only for asked account.
but that is a part of mariadb-operator..

the change seems to be working fine here. but it is a [DNM] patch.

zzzeek:
> though `DeleteDatabaseAndAccountFinalizers` additionaly delete `"unused" MariaDBAccounts`, which seems like something do be done on restart/start, instead of individual req. while `DeleteFinalizer` only for asked account. but that is a part of mariadb-operator..

this is not new functionality in the mariadb operator, it was there previously, and it's to delete any other MariaDBAccounts that may have existed previously regarding the particular MariaDBDatabase that the nova operator here wants to ensure is gone.    when an account rotation occurs, a new MariaDBAccount is created, and finalizers will be removed from the old MariaDBAccount, however it does not automatically delete the old MariaDBAccount.    this is not any kind of change right now.

go.mod
```diff
@@ -85,6 +85,8 @@ require (
 
 replace github.com/openstack-k8s-operators/nova-operator/api => ./api
 
+replace github.com/openstack-k8s-operators/mariadb-operator/api => github.com/zzzeek/mariadb-operator/api v0.3.1-0.20240917193141-8598d28fb998 //allow-merging
```
gibizer, line 88: don't use allow-merging, let the github check fail, this way we won't merge the PR before the dependency really lands.

gibizer (approved):
Looks good to me. Lets land the mariadb-operator PR and then we can land this.

go.mod
SeanMooney: i added hold until that is fixed feel free to drop it when its rebased

gibizer:
need a rebase and the go.mod replace line can be removed now as the mariadb-operator PR is landed.h

gibizer (commented):
looks good.

SeanMooney (approved):
the depency is now merged and im ok with this change so approving

PR #863, opened 2024-09-18
Escape password for tempest conf

mrkisaolamb:
In https://github.com/openstack-k8s-operators/tcib/commit/9258ba0ed856d5fdbf44a9cd25a49d51e4490ea7 eval was introduced and password for cirros start
returning an error

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -112,9 +112,9 @@
             service_available.swift false
             service_available.cinder false
             validation.image_ssh_user cirros
-            validation.image_ssh_password cubswin:)
+            validation.image_ssh_password cubswin:\)
             validation.image_alt_ssh_user cirros
-            validation.image_alt_ssh_password cubswin:)
+            validation.image_alt_ssh_password cubswin:\)
             validation.allowed_network_downtime 10
       cifmw_test_operator_tempest_include_list: |
           tempest.api.compute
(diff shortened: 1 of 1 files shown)


SeanMooney:
they do know that eval is generally considered a security risk and that probably should be reverted right...

.zuul.yaml
```diff
@@ -112,9 +112,9 @@
             service_available.swift false
             service_available.cinder false
             validation.image_ssh_user cirros
-            validation.image_ssh_password cubswin:)
+            validation.image_ssh_password cubswin:\)
```
SeanMooney, line 115: i think this can be validation.image_ssh_password 'cubswin:)'


if we read the commit they are saying
```
Use 'eval' to expand the TEMPESTCONF_OVERRIDES variable prior to
using it to run discover-tempest-config. This prevents embedded
quotes within the variable from being mangled.

Without the 'eval' embedded quoted strings like "Some String" get
mangled into '"Some' 'String"' on the resulting command line.
```

that would imply we should just be able to quote it to prevent having to escape anyting

mrkisaolamb: I tried eval locally with "cubswin:)" and unfortunatly return error

ASBishop: See https://github.com/openstack-k8s-operators/tcib/pull/212

gibizer:
I prefer https://github.com/openstack-k8s-operators/tcib/pull/212 as that is a global solution

mrkisaolamb:
agree, closing

PR #868, opened 2024-10-01
ensure nova-operator is always required

SeanMooney:
This change ensures that if we trigger a job without a chagne to the
nova operator repo via upstream nova that we include it as a
required project so that the meta content provider will include
the nova operator image.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -234,12 +234,20 @@
       jobs:
         - openstack-meta-content-provider:
             override-checkout: main
+            # when run from nova we need to ensure we list the nova-operator
+            # so that it will be built by the meta content provider
+            required-projects:
+              - github.com/openstack-k8s-operators/nova-operator
             vars:
               cifmw_bop_openstack_release: master
               cifmw_bop_dlrn_baseurl: "https://trunk.rdoproject.org/centos9-master"
               cifmw_repo_setup_branch: master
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -234,12 +234,20 @@
...
             vars:
               cifmw_bop_openstack_release: master
               cifmw_bop_dlrn_baseurl: "https://trunk.rdoproject.org/centos9-master"
               cifmw_repo_setup_branch: master
         - nova-operator-tempest-multinode: &job_vars
             override-checkout: main
+            # when run from nova we need to ensure we list the nova-operator
+            # so that it will be built by the meta content provider
+            required-projects:
+              - github.com/openstack-k8s-operators/nova-operator
```
SeanMooney, line 250: for context when we trigger a job form nova with a depends on it appears to work
but if we don't have a depend on against this repo and trigger it it fails in a way that
makes me think its because we don't actually build the correct images.

this is an attempt to fix that

the problem is i don't think we can test that without merging this.

the other way i could do this is to move the require projects to the definition of the jobs however that would require me to inherit form openstack-meta-content-provider
and add a nova-meta-content-provider with the required-project to make that work

so while i prefer to avoid creating job variants in templates i think this is the cleanest way to do this today

gibizer (approved):
looks safe to merge and iterate on if needed.

PR #871, opened 2024-10-07
Adds Goal statemet for nova-operator

auniyal61:
This change adds a goal statement for nova-operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,12 @@
 
 A golang operator for openstack nova lifecycle management
 
+## nova-operator Goal
+
+The goal of nova-operator is to manage custom resource that define a nova control plane (CRD).
+Nova operator continuously monitors the state of Nova CR and takes actions to ensure that the desired state is applied and reflected in nova pods. These pods run actual nova services, nova-operators ensure they are deployed, scaled, and configured correctly.
+**Note:** Nova CR is created by openstack-operator as per user intial deployment.
+
 ## Description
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,12 @@
 
 A golang operator for openstack nova lifecycle management
 
+## nova-operator Goal
+
+The goal of nova-operator is to manage custom resource that define a nova control plane (CRD).
+Nova operator continuously monitors the state of Nova CR and takes actions to ensure that the desired state is applied and reflected in nova pods. These pods run actual nova services, nova-operators ensure they are deployed, scaled, and configured correctly.
+**Note:** Nova CR is defined by openstack-operator as per user intial deployment.
+
```
SeanMooney, line 10: is this being added to other operators?

also 

**Note:** Nova CR is defined by openstack-operator as per user intial deployment.

that is not exactly correct.

we should not use the term defined, we should say created or instancaiated.

the definition of the CRD is part of nova-operator and defined in this sentence could be misinterpreted as implying that the definition fo the CR is part of OpenStack operator

the OpenStack operator mearly hydrates the embded nova template in the openstackContopleplane cr to create a nova cr that is not really a defintion.

auniyal61: I have checked with Cinder, Neutron and Glance operators now, its not added there.
I was trying to understand the goal, hence discussed with Gibi and added here.

Updated change.
thanks

SeanMooney (approved):

PR #872, opened 2024-10-08
Remove reference to the dataplane-operator

bshephar:
The dataplane-operator no longer exists since it has been merged with the openstack-operator. This change amends the terminology to change references to the dataplane-operator to dataplane controller in an effort to avoid user confusion.


gibizer (approved):
Thanks!

PR #880, opened 2024-10-16
Automate OLM cleanup for running local operator w/ webhooks

abays:
When using make `run-with-webhook`, local versions of the operator and its webhooks are added to the cluster. If the operator was previously installed via OLM, then there might be lingering webhooks from that installation. We've previously been assuming that the user would manually remove them, but we should just get rid of them ourselves since those OLM webhooks need to be deleted anyhow for the local webhooks to function unimpeded.  We can also automatically scale down the operator's OLM deployment in such a scenario.

### DIFF
--- a/hack/configure_local_webhook.sh
+++ b/hack/configure_local_webhook.sh
@@ -482,3 +482,11 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+# Scale-down operator deployment replicas to zero and remove OLM webhooks
+CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/nova-operator.openstack-operators -o name)"
+
+if [ -n "${CSV_NAME}" ]; then
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
+fi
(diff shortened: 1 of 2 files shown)


hack/configure_local_webhook.sh
```diff
@@ -482,3 +482,11 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+# Scale-down operator deployment replicas to zero and remove OLM webhooks
+CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/nova-operator.openstack-operators -o name)"
+
+if [ -n "${CSV_NAME}" ]; then
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
+    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
```
gibizer, line 491: how this relates to the webhook removal in https://github.com/openstack-k8s-operators/nova-operator/blob/main/hack/clean_local_webhook.sh Do we need both?

abays: `clean_local_webhook.sh` removes the local webhooks that are created via `configure_local_webhook.sh` (i.e. via `make run-with-webhook`).  `configure_local_webhook.sh` _removes_ the OLM-installed webhooks, if any, so that the local webhooks it is creating are the only ones used in the cluster.  It also down-scales the OLM-installed Nova operator controller manager pod so that the local instance is the only reconcile loop.

abays: Previously we were relying on the user to manually perform L490 and L491 before running `make run-with-webhook`.

gibizer: Ahh OK. Thanks for the clarification.

gibizer (approved):

PR #884, opened 2024-10-25
[rabbit] enable quorum queues

SeanMooney:
This change simplfies the oslo_messaging_rabbit config generations
by enabling rabbit_quorum_queue and hardcoding heartbeat_in_pthread
as false.

Closes: OSPRH-9736
Related: OSPRH-10790


bogdando:
let's test it with downstream uni job that runs minor updates

templates/nova.conf
```diff
@@ -55,19 +55,10 @@ key={{.SSLCertificateKeyFile}}
 lock_path = /var/lib/nova/tmp
 
 [oslo_messaging_rabbit]
-amqp_durable_queues=false
```
SeanMooney, line 58: rabbit notes that durable queue are required for quorm queues
https://www.rabbitmq.com/docs/quorum-queues#feature-comparison
but the oslo messaging option is ignored if quorum queues are enabled
https://github.com/openstack/oslo.messaging/blob/master/oslo_messaging/_drivers/amqp.py#L33-L37
so its more correct to not set this

bogdando: oh, good point, this wont' work w/o durable queues, and that requires separated vhosts for osp services, right?

bogdando: The last time I tested a rabbit cluster with quorum queues with my [Jepsen fork](https://github.com/bogdando/jepsen/commit/de0887d3990180fdb395909093b8f6fe6bdb4afa), it was not possible to do w/o setting durable queues, FYI.

SeanMooney: so i think reading the commnet in oslo that oslo will enable durable queue if you request quorma cues and ignore the amqp_druable_queues value

this failed with 

amqp.exceptions.PreconditionFailed: Exchange.declare: (406) PRECONDITION_FAILED - inequivalent arg 'durable' for exchange 'nova' in vhost '/': received 'true' but current is 'false'

and i think that was because all the other operators are using false.

ill try expclity setting ti to true but i think this just wont work without vhost

bogdando: > this just wont work without vhost

yes, as other services would conflict on creating things there with different persistence flags

bogdando: technically, we might want to test the altogether switch of the openstack services to that option in the downstream testproject

SeanMooney: Technically you should not be doing any dev downstream and we should be testing changing this everywhere upstream. 

we could just make nova use a separate rabbit for this testing but in reality this is better to test manually in a local deployment then try and do it in the ci.

with that said i wont really have time to work on this so what ever way works for you.

PR #887, opened 2024-10-30
Improve automated support for local webhooks

abays:
Reworks `make run-with-webhook` target to trap `SIGINT` and `SIGTERM` so as to call `clean_local_webhook.sh` in such a scenario.  This way, a user can run the operator locally and automatically have the local webhooks removed when they finish running the operator via `ctrl+c`.

Also creates a backup of any existing OLM CSV for the operator while preparing to run the operator locally with webhooks, _if_ the current CSV has a deployment `replicas` count greater than zero and/or there are `webhookdefinitions` present.  That backup can then be easily re-applied to restore OLM CSV defaults via `oc patch` by the user at their convenience.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -376,19 +376,19 @@ update-nova-csv:
 	fi
 
 # Used for webhook testing
-# The configure_local_webhooks.sh script below will remove any OLM webhooks
+# The configure_local_webhook.sh script below will remove any OLM webhooks
 # for the operator and also scale its deployment replicas down to 0 so that
 # the operator can run locally.
-# Make sure to cleanup the webhook configuration for local testing by running
-# ./hack/clean_local_webhook.sh before deplying with OLM again.
+# We will attempt to catch SIGINT/SIGTERM and clean up the local webhooks,
+# but it may be necessary to manually run ./hack/clean_local_webhook.sh
+# before deploying with OLM again for other untrappable signals.
 SKIP_CERT ?=false
 .PHONY: run-with-webhook
-run-with-webhook: export METRICS_PORT?=24600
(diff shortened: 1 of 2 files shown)


gibizer (approved):

SeanMooney:
as noteed on the placement version this would break my workflow if i was to work on the operator again.

i always rant the operator once via the make target, stopped it with contol-c and then ran it from vscode
so cleaning up the webhook config on ctrl-c is something is would never want to happen

Makefile
```diff
@@ -376,19 +376,19 @@ update-nova-csv:
...
-# ./hack/clean_local_webhook.sh before deplying with OLM again.
+# We will attempt to catch SIGINT/SIGTERM and clean up the local webhooks,
+# but it may be necessary to manually run ./hack/clean_local_webhook.sh
+# before deploying with OLM again for other untrappable signals.
 SKIP_CERT ?=false
 .PHONY: run-with-webhook
-run-with-webhook: export METRICS_PORT?=24600
-run-with-webhook: export HEALTH_PORT?=24601
-run-with-webhook: manifests generate fmt vet update-nova-csv ## Run a controller from your host.
+run-with-webhook: export METRICS_PORT?=8080
```
SeanMooney, line 387: these are using non default ports so that i could run the placement operator and nova operate locally without conflict.

SeanMooney: by that i mean using 2 vscode windows to run both of them form vscode.

if we don't just ever this i think we should turn of the cleanup by default and ever the port back to the non conflicting ones.

abays: I mistakenly copy-pasted over the unique defaults here.  I will submit a PR to restore them.

PR #888, opened 2024-10-30
Default heal_instance_info_cache_interval=-1

gibizer:
The supported networking backends expected to work without the periodic
healing of the cache in nova. So turning the periodic healing off can be
done to gain performance.

Resolves: [OSPRH-10744](https://issues.redhat.com//browse/OSPRH-10744)

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -50,6 +50,7 @@ cert={{.SSLCertificateFile}}
 key={{.SSLCertificateKeyFile}}
 {{end}}
 {{end}}
+heal_instance_info_cache_interval=-1
 
 [oslo_concurrency]
 lock_path = /var/lib/nova/tmp


gibizer:
recheck
Seems to be working OK.

gibizer:
recheck so far so good
no heal period is visible now in the nova-compute logs.

gibizer:
recheck unrelated
```
2024-11-18 14:07:12.745706 | controller |   msg: 'Failed to download metadata for repo ''delorean-component-barbican'': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried'
```

gibizer:
recheck content provider

gibizer:
lets try a rebase as it seems to be a compile error due to a recent change on main/openstack-operator.

mrkisaolamb (approved):
lgtm

gibizer:
/unhold
We have +1 both form network and compute perspective so I will let this merge.

SeanMooney (changes requested):
i apparently my request to not do this downstream only in the installer was not widely known

so just for psoperity when the jira that motivated this was filed i asked that we not diverge between the upstream behaviour of nova and the behavior of the installer.


ill propose a patch to nova to make this change there and we can see what upstream thinks of that.

bogdando:
based on inputs provided by @SeanMooney we can unfroze this one after https://review.opendev.org/c/openstack/nova/+/939453 is merged upstream

bogdando:
So, https://review.opendev.org/c/openstack/nova/+/939453 was merged, we can restore the merging path for this one

gibizer:
I suggest to backport https://review.opendev.org/c/openstack/nova/+/939453 to nova 18.0 instead.

PR #894, opened 2024-11-15
Ensure nodeSelector logic is consistent for all operators

olliewalsh:
nodeSelectors are not currently applied consistently across all operators.

Some do not support nodeSelectors at all - will be implemented in this series of pull requests.
Some do not apply them to every pod (jobs/cronjobs esp) - will be resolved in this series of pull requests.
Some do not apply a node selector update correctly, only when not initially set.
Some support nodeSelectors but do not inherit the default nodeSelector from the OpenstackControlPlane CR. 

Jira: [OSPRH-10734](https://issues.redhat.com//browse/OSPRH-10734)

### DIFF
--- a/pkg/nova/celldelete.go
+++ b/pkg/nova/celldelete.go
@@ -82,7 +82,7 @@ func CellDeleteJob(
 		},
 	}
 
-	if cell.Spec.NodeSelector != nil && len(*cell.Spec.NodeSelector) > 0 {
+	if cell.Spec.NodeSelector != nil {
 		job.Spec.Template.Spec.NodeSelector = *cell.Spec.NodeSelector
 	}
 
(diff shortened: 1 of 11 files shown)


stuggi (commented):
looks good to me

pkg/nova/celldelete.go
```diff
@@ -82,5 +82,9 @@ func CellDeleteJob(
 		},
 	}
 
+	if cell.Spec.NodeSelector != nil && len(*cell.Spec.NodeSelector) > 0 {
```
gibizer, line 85: Do we need the length check? Is there some limitation that we cannot set an empty map as a node selector?

gibizer: I would rather not add that condition if it is not needed.

gibizer (approved):
Thanks. Looks good to me.

PR #901, opened 2024-11-27
Delete next cell irrespective of last deletion

auniyal61:
Right now when multiple cells gets deleted if any one the cell deletion fails, the control exits with error msg.
This change stores the errors in a list and let next cells deleted.

Closes #[OSPRH-10550](https://issues.redhat.com//browse/OSPRH-10550)

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	batchv1 "k8s.io/api/batch/v1"
@@ -600,6 +602,16 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
+	sortNovaCellListByName := func(cellList *novav1.NovaCellList) {
+		sort.SliceStable(cellList.Items, func(i, j int) bool {
+			return cellList.Items[i].Name < cellList.Items[j].Name
+		})
+	}
+
+	sortNovaCellListByName(novaCellList)
+
+	var deleteErrs []error
+
 	for _, cr := range novaCellList.Items {
 		_, ok := instance.Spec.CellTemplates[cr.Spec.CellName]
 		if !ok {
@@ -607,12 +619,18 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 				cr.Spec.CellName, apiTransportURL,
 				secret, apiDB, cellDBs[novav1.Cell0Name].Database.GetDatabaseHostname(), cells[novav1.Cell0Name])
 			if err != nil {
-				return ctrl.Result{}, err
-			}
-			Log.Info("Cell deleted", "cell", cr.Spec.CellName)
-			delete(instance.Status.RegisteredCells, cr.Name)
-		}
+				deleteErrs = append(deleteErrs, fmt.Errorf("Cell '%s' deletion failed, because: %w", cr.Spec.CellName, err))
 
+			} else {
+				Log.Info("Cell deleted", "cell", cr.Spec.CellName)
+				delete(instance.Status.RegisteredCells, cr.Name)
+			}
+		}
+	}
+
+	if len(deleteErrs) > 0 {
+		delErrs := errors.Join(deleteErrs...)
+		return ctrl.Result{}, delErrs
 	}
 
 	Log.Info("Successfully reconciled")
--- a/test/functional/nova_multicell_test.go
+++ b/test/functional/nova_multicell_test.go
@@ -960,3 +960,196 @@ var _ = Describe("Nova multi cell", func() {
 		})
 	})
 })
+
+func CreateNovaWith4CellsAndEnsureReady(novaNames NovaNames) {
+	cell0 := novaNames.Cells["cell0"]
+	cell1 := novaNames.Cells["cell1"]
+	cell2 := novaNames.Cells["cell2"]
+	cell3 := novaNames.Cells["cell3"]
+
+	DeferCleanup(k8sClient.Delete, ctx, CreateNovaSecret(novaNames.NovaName.Namespace, SecretName))
+	DeferCleanup(k8sClient.Delete, ctx, CreateNovaMessageBusSecret(cell0))
+	DeferCleanup(k8sClient.Delete, ctx, CreateNovaMessageBusSecret(cell1))
+	DeferCleanup(k8sClient.Delete, ctx, CreateNovaMessageBusSecret(cell2))
+	DeferCleanup(k8sClient.Delete, ctx, CreateNovaMessageBusSecret(cell3))
+
+	serviceSpec := corev1.ServiceSpec{Ports: []corev1.ServicePort{{Port: 3306}}}
+	DeferCleanup(
+		mariadb.DeleteDBService,
+		mariadb.CreateDBService(novaNames.APIMariaDBDatabaseName.Namespace, novaNames.APIMariaDBDatabaseName.Name, serviceSpec))
+	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell0.MariaDBDatabaseName.Namespace, cell0.MariaDBDatabaseName.Name, serviceSpec))
+	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell1.MariaDBDatabaseName.Namespace, cell1.MariaDBDatabaseName.Name, serviceSpec))
+	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell2.MariaDBDatabaseName.Namespace, cell2.MariaDBDatabaseName.Name, serviceSpec))
+	DeferCleanup(mariadb.DeleteDBService, mariadb.CreateDBService(cell3.MariaDBDatabaseName.Namespace, cell3.MariaDBDatabaseName.Name, serviceSpec))
+
+	apiMariaDBAccount, apiMariaDBSecret := mariadb.CreateMariaDBAccountAndSecret(
+		novaNames.APIMariaDBDatabaseAccount, mariadbv1.MariaDBAccountSpec{})
+	DeferCleanup(k8sClient.Delete, ctx, apiMariaDBAccount)
+	DeferCleanup(k8sClient.Delete, ctx, apiMariaDBSecret)
+
+	cell0Account, cell0Secret := mariadb.CreateMariaDBAccountAndSecret(
+		cell0.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+	DeferCleanup(k8sClient.Delete, ctx, cell0Account)
+	DeferCleanup(k8sClient.Delete, ctx, cell0Secret)
+
+	cell1Account, cell1Secret := mariadb.CreateMariaDBAccountAndSecret(
+		cell1.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+	DeferCleanup(th.DeleteInstance, cell1Account)
+	DeferCleanup(
+		th.DeleteSecret,
+		types.NamespacedName{Name: cell1Secret.Name, Namespace: cell1Secret.Namespace})
+
+	cell2Account, cell2Secret := mariadb.CreateMariaDBAccountAndSecret(
+		cell2.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+	// DeferCleanup(k8sClient.Delete, ctx, cell2Account)
+	// DeferCleanup(k8sClient.Delete, ctx, cell2Secret)
+	logger.Info("Not Creating defercleanup for cell2 ...", " -- ", cell2Secret)
+
+	cell3Account, cell3Secret := mariadb.CreateMariaDBAccountAndSecret(
+		cell3.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+	// DeferCleanup(k8sClient.Delete, ctx, cell3Account)
+	// DeferCleanup(k8sClient.Delete, ctx, cell3Secret)
+	logger.Info("Not Creating defercleanup for cell3 ...", " -- ", cell3Secret)
+
+	spec := GetDefaultNovaSpec()
+	cell0Template := GetDefaultNovaCellTemplate()
+	cell0Template["cellDatabaseInstance"] = cell0.MariaDBDatabaseName.Name
+	cell0Template["cellDatabaseAccount"] = cell0Account.Name
+
+	cell1Template := GetDefaultNovaCellTemplate()
+	cell1Template["cellDatabaseInstance"] = cell1.MariaDBDatabaseName.Name
+	cell1Template["cellDatabaseAccount"] = cell1Account.Name
+	cell1Template["cellMessageBusInstance"] = cell1.TransportURLName.Name
+	cell1Template["novaComputeTemplates"] = map[string]interface{}{
+		ironicComputeName: GetDefaultNovaComputeTemplate(),
+	}
+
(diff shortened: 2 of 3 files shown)


auniyal61:
The jira report says, we need to do this for cellCreation as well, but I think its already taken care of at here https://github.com/openstack-k8s-operators/nova-operator/blob/main/controllers/nova_controller.go#L328-L356

mrkisaolamb (changes requested):
Please add some func tests for cell deletion to assert if deletion is working and returning errors for all cells. Also it will be nice to assert what are conditions of nova instance

test/functional/nova_multicell_test.go
```diff
@@ -51,14 +51,191 @@ var _ = Describe("Nova multi cell", func() {
...
+			// NovaCellNotExists(cell3.CellCRName)
+			GetNovaCell(cell2.CellCRName)
+
+			// cell2 deletion should have failed
+			Eventually(func(g Gomega) {
+				mappingJob := th.GetJob(cell2.CellDeleteJobName)
+				newJobInputHash := GetEnvVarValue(
+					mappingJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
+				g.Expect(newJobInputHash).NotTo(BeEmpty())
+			}, timeout, interval).Should(Succeed())
```
gibizer, line 225: I'm not sure how this proves that the cell2 delete failed. It checks for the cell2 delete job to exists
.

test/functional/nova_multicell_test.go
```diff
@@ -51,14 +51,191 @@ var _ = Describe("Nova multi cell", func() {
...
+
+			// manually delete DB for cell2, to reproduce the error in cell2 deletion
+			Expect(k8sClient.Delete(ctx, mariadb.GetMariaDBDatabase(cell2.MariaDBDatabaseName))).To(Succeed())
+
+			// delete cell2 and then cell3,
+			// the overall call should succeed
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				delete(nova.Spec.CellTemplates, "cell2")
+				delete(nova.Spec.CellTemplates, "cell3")
```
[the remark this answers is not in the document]
gibizer: Maybe the way to make it testable is to sort the cells be cell name when iterated in the controller. We do that for cell creation anyhow in https://github.com/openstack-k8s-operators/nova-operator/pull/903 to avoid the unstable ordering to cause unstable condition messages.

controllers/nova_controller.go
```diff
@@ -600,18 +600,26 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
+	var deleteErrs []error
+
 	for _, cr := range novaCellList.Items {
```
gibizer, line 605: I would sort this list by cell name for two reasons:
* https://github.com/openstack-k8s-operators/nova-operator/pull/903 does it for cell creation already to make the error messages stable even if the order of the list of cells in the api is not stable.
* this makes testing possible as we can rely on the ordering to prove that even if cell2 failed to delete cell3 deletion is tried.

gibizer (commented):
Deletion logic looks good with a small request about ordering. I have couple of comments on the testing side.

gibizer:
Running the test locally shows that cell2 is also successfully deleted so the current test does not recreate the necessary scenario to prove that failure in one cell does not prevent the deletion of the other.

test/functional/nova_multicell_test.go
```diff
@@ -51,14 +51,183 @@ var _ = Describe("Nova multi cell", func() {
...
+			th.SimulateStatefulSetReplicaReady(cell3.ConductorStatefulSetName)
+			th.SimulateJobSuccess(cell3.CellMappingJobName)
+
+		})
+
+		It("deletes cell3 and verifies error for cell2 because its DB deleted already", func() {
+
+			// manually delete DB for cell2, to reproduce the error in cell2 deletion
+			Expect(k8sClient.Delete(ctx, mariadb.GetMariaDBDatabase(cell2.MariaDBDatabaseName))).To(Succeed())
+
```
gibizer, line 201: I have added a bit of check here:
```golang
                       nova := GetNova(novaNames.NovaName)
                       Expect(nova.Status.RegisteredCells).To(HaveKey(cell3.CellCRName.Name), "cell3 is not in the RegisteredCells", nova.Status.RegisteredCells)
```

This shows that the starting point of the test is probably incomplete as cell3 is not in RegisteredCells. So I suggest to investigate why cell3 is not there yet.

gibizer: Maybe this is a source of the instability we saw during our call when first we saw that the controller loggedd it deleted cell3 then later we was not able to reproduce that.

auniyal61: `[FAILED] cell3 is not in the RegisteredCells%!(EXTRA map[string]string=map[])`

got same for cell0 and cell1 too.

test/functional/nova_multicell_test.go
auniyal61: removed

test/functional/nova_multicell_test.go
auniyal61: in updated change cell0, cell1, cell2 are listed.

controllers/nova_controller.go
auniyal61: novaCellList is already sorted. so deleteErrs will be sorted automatically.

test/functional/nova_multicell_test.go
```diff
@@ -952,3 +952,198 @@ var _ = Describe("Nova multi cell", func() {
...
+		It("delete cell1 and cell2, verify for cell1", func() {
+
+			nova := GetNova(novaNames.NovaName)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell0.CellCRName.Name), "cell0 is not in the RegisteredCells", nova.Status.RegisteredCells)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name), "cell1 is not in the RegisteredCells", nova.Status.RegisteredCells)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell2.CellCRName.Name), "cell2 is not in the RegisteredCells", nova.Status.RegisteredCells)
+
+			// manually delete DB for cell2, to reproduce the error in cell2 deletion
+			Expect(k8sClient.Delete(ctx, mariadb.GetMariaDBDatabase(cell1.MariaDBDatabaseName))).To(Succeed())
```
auniyal61, line 1126: this is not reproducing cell-deletion error

controllers/nova_controller.go
```diff
@@ -600,18 +600,26 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+				Log.Info("Cell deleted", "cell", cr.Spec.CellName)
+				delete(instance.Status.RegisteredCells, cr.Name)
 			}
-			Log.Info("Cell deleted", "cell", cr.Spec.CellName)
-			delete(instance.Status.RegisteredCells, cr.Name)
 		}
+	}
+
+	if len(deleteErrs) > 0 {
+		return ctrl.Result{}, fmt.Errorf("errors: %v", deleteErrs)
```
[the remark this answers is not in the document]
auniyal61: updated, thanks

gibizer:
@auniyal61 so you pinged me on slack about the DeferCleanup issue. 
> so I am deleting a cell, but cell-resource (like secret ) is defined in setup (Before each or another funciton).
now once tests is finished deferecleanup runs and look for secret, but as we already deleted cell, its not present in DB, so defercleanup complains,
how to resolve this

The DeferCleanup doc says:
https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup
> DeferCleanup can be passed a function that takes no arguments and returns no value. You can also pass a function that returns values. DeferCleanup ignores all these return value except for the last. If the last return value is a non-nil error - a common go pattern - DeferCleanup will fail the spec.

So I guess the situation is that 
1. we create cell3 during the BeforeEach() section and put a DeferCleanup() call to the all the pieces we created
2. then delete the cell during the normal It() section
3. then the test runner executes the deferred cleanups where the client.Delete call returns a non nil error as the last return value  because the resource is not found and that causes the test cleanup to fail as defined by the doc.

This means we have multiple options:
* a) we wrap the client.Delete call and ignore the NotFound error. So cleanup still tries to delete the resources but if they do not exists then the cleanup still succeeds.
* b) we don't create the cell3 in the shared CreateNovaWith3CellsAndEnsureReady() helper. But create it just for our test case (in our BeforeEach or in the It ) and we don't put DeferCleanup for the pieces of cell3 as we expect that the our test will delete them anyhow.

I would do option b) as it is more explicit about why we do thing differently.

(I need to get back to your other comments above a bit later)

auniyal61:
> So I guess the situation is that
> 
> 1. we create cell3 during the BeforeEach() section and put a DeferCleanup() call to the all the pieces we created
> 2. then delete the cell during the normal It() section
> 3. then the test runner executes the deferred cleanups where the client.Delete call returns a non nil error as the last return value  because the resource is not found and that causes the test cleanup to fail as defined by the doc.
>
Yes
 
> This means we have multiple options:
> 
> * a) we wrap the client.Delete call and ignore the NotFound error. So cleanup still tries to delete the resources but if they do not exists then the cleanup still succeeds.

> * b) we don't create the cell3 in the shared CreateNovaWith3CellsAndEnsureReady() helper. But create it just for our test case (in our BeforeEach or in the It ) and we don't put DeferCleanup for the pieces of cell3 as we expect that the our test will delete them anyhow.
> 
> I would do option b) as it is more explicit about why we do thing differently.

we have to create new-cell in CreateNovaWith3CellsAndEnsureReady, because that's where `CR-spec` is defined and saved, we can't pass and update spec later in our `Describe("Nova multi cell deletion")` 

the oher way would be to redefine all cells again, which is just a duplicate code of CreateNovaWith3CellsAndEnsureReady.

test/functional/nova_multicell_test.go
```diff
@@ -952,3 +952,196 @@ var _ = Describe("Nova multi cell", func() {
...
+			}, timeout, interval).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell2.CellCRName.Name))
+				g.Expect(nova.Status.RegisteredCells).NotTo(HaveKey(cell3.CellCRName.Name))
+			}, timeout, interval).Should(Succeed())
+
+			GetNovaCell(cell2.CellCRName)
+			NovaCellNotExists(cell3.CellCRName)
```
auniyal61, line 1141: yet to read delete-err msg

test/functional/nova_multicell_test.go
auniyal61: went with updating secret to empty string.

test/functional/nova_multicell_test.go
```diff
@@ -952,3 +952,210 @@ var _ = Describe("Nova multi cell", func() {
...
+		It("delete cell2 and cell3, verify for cell3", func() {
+
+			nova := GetNova(novaNames.NovaName)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell0.CellCRName.Name), "cell0 is not in the RegisteredCells", nova.Status.RegisteredCells)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name), "cell1 is not in the RegisteredCells", nova.Status.RegisteredCells)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell2.CellCRName.Name), "cell2 is not in the RegisteredCells", nova.Status.RegisteredCells)
+			Expect(nova.Status.RegisteredCells).To(HaveKey(cell3.CellCRName.Name), "cell3 is not in the RegisteredCells", nova.Status.RegisteredCells)
+
+			// manually delete DB for cell3, to reproduce the error in cell3 deletion
+			// Expect(k8sClient.Delete(ctx, mariadb.GetMariaDBAccount(cell3.MariaDBDatabaseName))).To(Succeed())
```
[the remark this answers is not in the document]
auniyal61: thanks Gibi, updated.

test/functional/nova_multicell_test.go
auniyal61: Verifying cell2 exists is enough here.
this will update controlplane status after CR update

test/functional/nova_multicell_test.go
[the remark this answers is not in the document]
auniyal61: yes novaCellList is already sorted and cell2 always comes before cell3 in reconciler.

controllers/nova_controller.go
mrkisaolamb: I think what @gibi meant is that novaCellList can be retrieved from the database in any order. So, if we want to sort it by name, we need to do something like this:
`// SortNovaCellListByName sorts the NovaCellList by name
func SortNovaCellListByName(cellList *novav1.NovaCellList) {
    sort.SliceStable(cellList.Items, func(i, j int) bool {
        return cellList.Items[i].Name < cellList.Items[j].Name
    })
}`

Similar to what we did with orderedCellNames, where we ensured cell0 was placed first.

auniyal61: added sortNovaCellListByName

test/functional/nova_multicell_test.go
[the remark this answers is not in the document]
auniyal61: sorted

controllers/nova_controller.go
```diff
@@ -600,19 +602,35 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
+	sortNovaCellListByName := func(cellList *novav1.NovaCellList) {
```
mrkisaolamb, line 605: it will be nice to move it to common and write some test

gibizer (approved):
I have couple of follow up requests but the current code and test looks good to me so lets land this and do the refactors in a separate PR

controllers/nova_controller.go
```diff
@@ -600,19 +602,35 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+				Log.Info("Cell deleted", "cell", cr.Spec.CellName)
+				delete(instance.Status.RegisteredCells, cr.Name)
 			}
-			Log.Info("Cell deleted", "cell", cr.Spec.CellName)
-			delete(instance.Status.RegisteredCells, cr.Name)
 		}
+	}
 
+	if len(deleteErrs) > 0 {
+		delErrs := errors.Join(deleteErrs...)
```
mrkisaolamb, line 632: It will be good to land this changes before https://github.com/openstack-k8s-operators/nova-operator/pull/917 then we will don't need to change logic in this pr but directly in 917

mrkisaolamb (approved):

controllers/nova_controller.go
auniyal61: done

controllers/nova_controller.go
```diff
@@ -600,19 +602,35 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
+	sortNovaCellListByName := func(cellList *novav1.NovaCellList) {
+		sort.SliceStable(cellList.Items, func(i, j int) bool {
+			return cellList.Items[i].Name < cellList.Items[j].Name
+		})
+	}
+
+	sortNovaCellListByName(novaCellList)
```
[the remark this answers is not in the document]
auniyal61: follow-up patch https://github.com/openstack-k8s-operators/nova-operator/pull/940

controllers/nova_controller.go
auniyal61: done #940

PR #902, opened 2024-11-28
Use cifmw kuttl job as parent job for reusability

cescgina:
In the depends-on PR, a new job definition is introduced that runs kuttl
tests following the pattern used in nova-operator. This is done mainly
to reuse the job definition for watcher-operator, which has just
introduced kuttl testing following the same pattern.

This change modifies the nova job to use the same parent.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2571


cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

.zuul.yaml
```diff
@@ -26,22 +26,20 @@
 
 - job:
     name: nova-operator-kuttl
-    parent:  nova-operator-base
```
SeanMooney, line 29: this is the only job using nova-operator-base now so we can likely remove that job definition now too. that's not critical however as it just an abstract job definition just wanted to mention that.

cescgina: true, now that I got confirmation that the kuttl job works ok, I've pushed a new commit to remove the unused files

SeanMooney:
check-rdo the depency is now merged so lets retests with the final version

SeanMooney (approved):
im happy that this is functional and that the job logging is still working as intended

https://logserver.rdoproject.org/02/902/59d806e4202ea7ee893d86dd2f1fa2303b19c3e2/github-check/nova-operator-kuttl/8d4d7ca/controller/controller/kuttl-test-run.out

https://softwarefactory-project.io/zuul/t/rdoproject.org/build/8d4d7ca2d6cb46918e798b8786a80753/log/job-output.txt#4487-4494

PR #903, opened 2024-11-28
Sort cell-names in nova_controller

ratailor:
This change sorts the cell-names and also make sure that the cell0 is the first element in the orderedCellNames slice.

Closes: #[OSPRH-9980](https://issues.redhat.com//browse/OSPRH-9980)

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


SeanMooney (changes requested):

ratailor:
> How does the code on main leads to unstable condition messages? Could you please add that description to the PR's message?

Done

controllers/nova_controller.go
```diff
@@ -305,16 +306,20 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
-	orderedCellNames := []string{novav1.Cell0Name}
+	orderedCellNames := []string{}
 	for cellName := range instance.Spec.CellTemplates {
 		if cellName != novav1.Cell0Name {
 			orderedCellNames = append(orderedCellNames, cellName)
 		}
 	}
 
+	// We need to sort the list of cellNames to iterate on to avoid
+	// unnecessary reconcile loop runs, as the map iteration order is undefined.
```
gibizer, line 317: This is not for avoiding reconcile loops. It is for avoiding that different reconcile loops sees a different cell order while the cell creation is in the same state but generate a different condition message due to the different order.

gibizer:
I still would like to get a commit message / PR description that describes why we are doing this instead of what the patch does.

mrkisaolamb:
IMO it will be good to have some functional tests

bogdando (commented):
any updates?

mrkisaolamb:
original ticket has been closed, and this PR is now outdated

PR #904, opened 2024-11-29
Configure dogpile.cache to deal with memcached pods failures

lmiccini:
Whenever one of the mecached pods disappears, because of a rolling restart during a minor update or as result of a failure, APIs can take a long time to detect that the pod went away and keep trying to reconnect.

From a quick round of tests we saw downtimes up to ~150s.

By enabling the retry_client and limiting the number of retries the behavior seems much more acceptable.

Similarly, when TLS is not in use, we may want to set a lower value for memcache_dead_retry so to eventually reconnect to a new pod (having the same dns name but different ip) much faster.

Related: [OSPRH-11935](https://issues.redhat.com//browse/OSPRH-11935)

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -173,9 +173,13 @@ enabled = True
 {{if .MemcachedTLS}}
 backend = dogpile.cache.pymemcache
 memcache_servers={{ .MemcachedServers }}
+enable_retry_client = true
+retry_attempts = 2
+retry_delay = 0
 {{else}}
 backend = dogpile.cache.memcached
 memcache_servers={{ .MemcachedServersWithInet }}
+memcache_dead_retry = 10
 {{end}}
 tls_enabled={{ .MemcachedTLS }}
 {{else}}


templates/nova.conf
```diff
@@ -172,8 +172,12 @@ enabled = True
 # on contoler we prefer to use memcache when its deployed
 {{if .MemcachedTLS}}
 backend = dogpile.cache.pymemcache
+enable_retry_client = true
+retry_attempts = 2
+retry_delay = 0
```
auniyal61, line 177: the issue is reproduced with server list as well, it does take lot of time, even after several sec of memcached pod start.
but there was no visible change after updating suggested change in confs for me.

lmiccini: Thanks Amit. You need to patch the keystone config as well, otherwise you'll still have keystone waiting for the memcache pod to come back (see [https://github.com/openstack-k8s-operators/keystone-operator/pull/511](https://github.com/openstack-k8s-operators/keystone-operator/pull/511)) .

SeanMooney: the other thing to consider is we use memcache to cache the server metadata which his expensive to compute. restaring memcache like this effectively clears the cache so new requests will miss the cache and need to hit the db directly, that's fine but it could cause things to timeout.

this si mostly migrated because we are using config drive by default and there for instance boot has no direct dependency on the matadata API but it is just something to be aware of.

any tempest tests that tried to hit the metadata API for an instance that was created before the minor update will take slightly longer the normal to rescive the response beasue its not coming form the cache.

mrkisaolamb (approved):
lgtm

PR #916, opened 2025-01-10
Improve error handling

mrkisaolamb:
To follow best practices for error and exception handling in Go and to prevent regressions,
err113 and gosec checks in golang-ci-linter are now enabled.
Errors now use API errors from k8s.io/apimachinery and predefined messages for status updates.

Resolve: https://issues.redhat.com/browse/OSPRH-9046
Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/602


controllers/novaapi_controller.go
```diff
@@ -588,13 +588,12 @@ func (r *NovaAPIReconciler) ensureDeployment(
 	if networkReady {
 		instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
 	} else {
-		err := fmt.Errorf("not all pods have interfaces with ips as configured in NetworkAttachments: %s", instance.Spec.NetworkAttachments)
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.NetworkAttachmentsReadyCondition,
 			condition.ErrorReason,
 			condition.SeverityWarning,
-			condition.NetworkAttachmentsReadyErrorMessage,
```
gibizer, line 596: This is used in common.go ensureNetworkAttachments() as well, but you did not changed it there. Is it because there the message using an error coming from a call instead of defined locally?

api/v1beta1/conditions.go
```diff
@@ -56,6 +56,9 @@ const (
 
 // Common Messages used by API objects.
 const (
+	//NovaNetworkAttachmentsReadyInitMessage
+	NovaNetworkAttachmentsReadyErrorMessage = "NetworkAttachments error occurred not all pods have interfaces with ips as configured in NetworkAttachments: %s"
+
```
gibizer, line 61: Will this eventually move to lib-common to be used across the operators? E.g. in https://github.com/openstack-k8s-operators/placement-operator/blob/5c39b092c37feddb1d94539b759ee8cebab371e2/controllers/placementapi_controller.go#L1159-L1160

controllers/common.go
```diff
@@ -225,7 +223,10 @@ func ensureSecret(
 	for _, field := range expectedFields {
 		val, ok := secret.Data[field]
 		if !ok {
-			err := fmt.Errorf("field '%s' not found in secret/%s", field, secretName.Name)
+			// Usage in the code
```
gibizer, line 226: I don't get this comment

controllers/common.go
```diff
@@ -225,7 +223,10 @@ func ensureSecret(
...
+			err := k8s_errors.NewBadRequest(fmt.Sprintf("field '%s' not found in secret/%s", field, secretName.Name))
+			// ensureSecret - ensures that the Secret object exists and the expected fields
+			// are in the Secret. It returns a hash of the values of the expected fields.
```
gibizer, line 229: This belongs to the function definition.

controllers/common.go
```diff
@@ -579,7 +580,7 @@ func getNovaClient(
 			return nil, err
 		}
 		if (ctrlResult != ctrl.Result{}) {
-			return nil, fmt.Errorf("the CABundleSecret %s not found", auth.GetCABundleSecretName())
+			return nil, fmt.Errorf("the CABundleSecret %s not found: %w", auth.GetCABundleSecretName(), err)
```
gibizer, line 583: as far as I see err will be always `nil` here as if not then the previous condition is true and the function returns.

controllers/common.go
```diff
@@ -225,7 +223,10 @@ func ensureSecret(
 	for _, field := range expectedFields {
 		val, ok := secret.Data[field]
 		if !ok {
-			err := fmt.Errorf("field '%s' not found in secret/%s", field, secretName.Name)
+			// Usage in the code
+			err := k8s_errors.NewBadRequest(fmt.Sprintf("field '%s' not found in secret/%s", field, secretName.Name))
```
gibizer, line 227: Do we call ensureSecret from the validation webhook? As far as I remember we don't. So it is a bit strange to me to use HTTP response codes as errors here. I.e. the NewBadRequest creates a StatusError where the description says:
> StatusError is an error intended for consumption by a REST API server; it can also be reconstructed by clients from a REST response. Public to allow easy type switches.


https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#StatusError

controllers/common.go
```diff
@@ -641,7 +642,7 @@ func ensureMemcached(
 				condition.RequestedReason,
 				condition.SeverityInfo,
 				condition.MemcachedReadyWaitingMessage))
-			return nil, fmt.Errorf("memcached %s not found", memcachedName)
+			return nil, fmt.Errorf("%w: memcached %s not found", err, memcachedName)
```
gibizer, line 645: yepp this is a good improvement. Thanks.

controllers/nova_controller.go
```diff
@@ -1631,8 +1631,8 @@ func (r *NovaReconciler) ensureMQ(
 
 	url, ok := secret.Data[TransportURLSelector]
 	if !ok {
-		return "", nova.MQFailed, fmt.Errorf(
-			"the TransportURL secret %s does not have 'transport_url' field", transportURL.Status.SecretName)
+		return "", nova.MQFailed, k8s_errors.NewBadRequest(fmt.Sprintf(
```
gibizer, line 1634: ditto REST errors feels strange to me inside the reconciler loop

mrkisaolamb: done

controllers/novaapi_controller.go
mrkisaolamb: yes exactly that was the reasone

mrkisaolamb (commented):
Thanks for review!

api/v1beta1/conditions.go
stuggi: seems we could probably just update https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/conditions.go#L269 ?

api/v1beta1/conditions.go
```diff
@@ -172,4 +175,7 @@ const (
 
 	//CellHostDiscoverErrorMessage
 	CellHostDiscoverErrorMessage = "CellHostDiscover error occurred %s"
+
+	//InputReadyErrorMessage
+	InputReadyErrorMessage = "Input data error occurred %s"
```
stuggi, line 180: is this used somewhere?

controllers/errors.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"errors"
+)
+
+var (
+	ErrResourceIsNotReady = errors.New("Resource is not ready")
+	ErrInvalidStatus      = errors.New("invalid status")
+	ErrCannotUpdateObject = errors.New("cannot update object")
+	ErrFieldNotFound      = errors.New("field not found in Secret")
```
stuggi, line 11: maybe also good to have those generic errors in lib-common

stuggi (commented):
lgtm

PR #917, opened 2025-01-15
Delete NovaCell cr only when deletion job pass

mrkisaolamb:
resolve: https://issues.redhat.com/browse/OSPRH-12582

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -732,3 +732,9 @@ func ensureMemcached(
 
 	return memcached, err
 }
+
+func sortNovaCellListByName(cellList *novav1.NovaCellList) {
+	sort.SliceStable(cellList.Items, func(i, j int) bool {
+		return cellList.Items[i].Name < cellList.Items[j].Name
+	})
+}
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sort"
 	"strings"
 
 	batchv1 "k8s.io/api/batch/v1"
@@ -602,12 +601,6 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
-	sortNovaCellListByName := func(cellList *novav1.NovaCellList) {
-		sort.SliceStable(cellList.Items, func(i, j int) bool {
-			return cellList.Items[i].Name < cellList.Items[j].Name
-		})
-	}
-
 	sortNovaCellListByName(novaCellList)
 
 	var deleteErrs []error
@@ -615,13 +608,13 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	for _, cr := range novaCellList.Items {
 		_, ok := instance.Spec.CellTemplates[cr.Spec.CellName]
 		if !ok {
-			err := r.ensureCellDeleted(ctx, h, instance,
+			result, err := r.ensureCellDeleted(ctx, h, instance,
 				cr.Spec.CellName, apiTransportURL,
 				secret, apiDB, cellDBs[novav1.Cell0Name].Database.GetDatabaseHostname(), cells[novav1.Cell0Name])
 			if err != nil {
 				deleteErrs = append(deleteErrs, fmt.Errorf("Cell '%s' deletion failed, because: %w", cr.Spec.CellName, err))
-
-			} else {
+			}
+			if result == nova.CellDeleteComplete {
 				Log.Info("Cell deleted", "cell", cr.Spec.CellName)
 				delete(instance.Status.RegisteredCells, cr.Name)
 			}
@@ -687,7 +680,7 @@ func (r *NovaReconciler) ensureCellDeleted(
 	apiDB *mariadbv1.Database,
 	APIDatabaseHostname string,
 	cell0 *novav1.NovaCell,
-) error {
+) (nova.CellDeploymentStatus, error) {
 	Log := r.GetLogger(ctx)
 	cell := &novav1.NovaCell{}
 	fullCellName := types.NamespacedName{
@@ -700,17 +693,17 @@ func (r *NovaReconciler) ensureCellDeleted(
 		// We cannot do further cleanup of the MariaDBDatabase and
 		// MariaDBAccount as their name is only available in the NovaCell CR
 		// since the cell definition is removed from the Nova CR already.
-		return nil
+		return nova.CellDeleteComplete, nil
 	}
 	if err != nil {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 	// If it is not created by us, we don't touch it
 	if !OwnedBy(cell, instance) {
 		Log.Info("Cell isn't defined in the Nova, but there is a  "+
 			"Cell CR not owned by us. Not deleting it.",
 			"cell", cell)
-		return nil
+		return nova.CellDeleteComplete, nil
 	}
 
 	dbName, accountName := novaapi.ServiceName+"-"+cell.Spec.CellName, cell.Spec.CellDatabaseAccount
@@ -718,11 +711,11 @@ func (r *NovaReconciler) ensureCellDeleted(
 	configHash, scriptName, configName, err := r.ensureNovaManageJobSecret(ctx, h, instance,
 		cell0, topLevelSecret, APIDatabaseHostname, apiTransportURL, apiDB)
 	if err != nil {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 	inputHash, err := util.HashOfInputHashes(configHash)
 	if err != nil {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 
 	labels := map[string]string{
@@ -734,24 +727,30 @@ func (r *NovaReconciler) ensureCellDeleted(
 		instance.Spec.PreserveJobs, r.RequeueTimeout,
 		inputHash)
 
-	_, err = job.DoJob(ctx, h)
+	result, err := job.DoJob(ctx, h)
 	if err != nil {
-		return err
+		return nova.CellDeleteFailed, err
+	}
+
+	if (result != ctrl.Result{}) {
+		// Job is still running. We can simply return as we will be reconciled
+		// when the Job status changes
+		return nova.CellDeleteInProgress, nil
 	}
 
 	secretName := getNovaCellCRName(instance.Name, cellName)
 	err = secret.DeleteSecretsWithName(ctx, h, secretName, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 	configSecret, scriptSecret := r.getNovaManageJobSecretNames(cell)
 	err = secret.DeleteSecretsWithName(ctx, h, configSecret, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 	err = secret.DeleteSecretsWithName(ctx, h, scriptSecret, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
-		return err
+		return nova.CellDeleteFailed, err
 	}
 
 	// Delete transportURL cr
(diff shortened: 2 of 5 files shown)


controllers/nova_controller.go
```diff
@@ -603,14 +603,16 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
 				secret, apiDB, cellDBs[novav1.Cell0Name].Database.GetDatabaseHostname(), cells[novav1.Cell0Name])
 			if err != nil {
 				return ctrl.Result{}, err
 			}
-			Log.Info("Cell deleted", "cell", cr.Spec.CellName)
-			delete(instance.Status.RegisteredCells, cr.Name)
+			if result == nova.CellDeleteComplete {
+				Log.Info("Cell deleted", "cell", cr.Spec.CellName)
+				delete(instance.Status.RegisteredCells, cr.Name)
+			}
```
gibizer, line 615: so if the result is not CellDeleteComplete then we return success from the reconciler below and allowing the NovaCell object to be deleted, but we keep the nova.Status.RegisteredCells entry. However when the job finally finishes and a new reconciler starts on nova controller the novaCellList above will not containt the NovaCell CR any more so the nova controller will not cleanup the RegisteredCells entry.

controllers/nova_controller.go
```diff
@@ -718,22 +720,22 @@ func (r *NovaReconciler) ensureCellDeleted(
 
 	_, err = job.DoJob(ctx, h)
 	if err != nil {
-		return err
+		return nova.CellDeleteFailed, err
```
gibizer, line 723: I think we need to check the result of the DoJob to see if the deletion job is still running and return to the caller if so.

pkg/nova/common.go
```diff
@@ -96,6 +96,12 @@ const (
 	// CellComputeDiscoveryReady indicates that all NovaComputes in cell reached the Ready status and
 	// the hosts in the cell have been successfully discovered
 	CellComputeDiscoveryReady CellDeploymentStatus = iota
+	// CellDeleteInProgress indicates that the NovaCell deletion is in progress
+	CellDeleteInProgress CellDeploymentStatus = iota
```
gibizer, line 100: You are right we need in progress indication. You however missed to use this status.

gibizer (changes requested):

controllers/nova_controller.go
mrkisaolamb: Thanks! I added additional loop to check if registered cell job was sucess to ensure that we have correct status

test/functional/nova_reconfiguration_test.go
```diff
@@ -182,18 +182,28 @@ var _ = Describe("Nova reconfiguration", func() {
...
 			}, timeout, interval).Should(Succeed())
+
+			// Check that the cell status is not deleted
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name))
+			}, timeout, interval).Should(Succeed())
+
+			// Simulate the cell delete job success
+			th.SimulateJobSuccess(cell1.CellDeleteJobName)
```
auniyal61, line 199: why to simulate, isn't this already verified in L189 ?

test/functional/nova_reconfiguration_test.go
```diff
@@ -182,18 +182,28 @@ var _ = Describe("Nova reconfiguration", func() {
...
 				newJobInputHash := GetEnvVarValue(
 					mappingJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
 				g.Expect(newJobInputHash).NotTo(BeNil())
 			}, timeout, interval).Should(Succeed())
+
+			// Check that the cell status is not deleted
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name))
+			}, timeout, interval).Should(Succeed())
```
auniyal61, line 196: is this a side-effect/bug for now ?
we removed it from CR at L180, still we are verifying thats its present in CR.
I have seen this, that after even delete job is  completed cell exists in CR (because of finalizer I think!)

controllers/nova_controller.go
```diff
@@ -603,18 +603,42 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		if !ok {
+			job, err := job.GetJobWithName(ctx, h, crName+"-cell-delete", instance.Namespace)
+			if err != nil {
+				if !k8s_errors.IsNotFound(err) {
+					return ctrl.Result{}, err
+				}
+			}
+			if job != nil {
+				if job.Status.Succeeded == 1 {
+					Log.Info("Cell deleted", "cell", crName)
```
auniyal61, line 632: should this say "cell job is deleted, cell deletion completed"
we are alredy saying cell deleted at L613

test/functional/nova_reconfiguration_test.go
mrkisaolamb: We verified that object exist not if it suceed

controllers/nova_controller.go
mrkisaolamb: added result check for job

controllers/nova_controller.go
mrkisaolamb: I changed logic of checks so this loop is no need anymore

test/functional/nova_reconfiguration_test.go
mrkisaolamb: We are checking here if we still have registered cell because  job is still runing so we shoudn't remove cell1 from status. Only when job sucess reconciler code should procede further

controllers/common.go
```diff
@@ -732,3 +732,9 @@ func ensureMemcached(
 
 	return memcached, err
 }
+
+func sortNovaCellListByName(cellList *novav1.NovaCellList) {
+	sort.SliceStable(cellList.Items, func(i, j int) bool {
+		return cellList.Items[i].Name < cellList.Items[j].Name
+	})
+}
```
auniyal61, line 740: sure, though I do have this move added here with a test specific to sorting 
 https://github.com/openstack-k8s-operators/nova-operator/pull/940/files

test/functional/nova_multicell_test.go
```diff
@@ -1139,6 +1139,8 @@ var _ = Describe("Nova multi cell deletion", func() {
 				g.Expect(k8sClient.Update(ctx, nova)).To(Succeed())
 			}, timeout, interval).Should(Succeed())
 
+			th.SimulateJobSuccess(cell2.CellDeleteJobName)
+			th.SimulateJobSuccess(cell3.CellDeleteJobName)
```
auniyal61, line 1143: +1

controllers/common.go
mrkisaolamb: Yeah I did this like 5 day ago after rebase https://github.com/openstack-k8s-operators/nova-operator/compare/99a98e8e08ca4e49c4569aac778a58c423649c81..2b3cf8c0f7981a1d1d9418052411109164ef6b30 . This PR is going for quite long time so I didn't want to wait on any other pr

test/functional/nova_reconfiguration_test.go
```diff
@@ -183,18 +183,28 @@ var _ = Describe("Nova reconfiguration", func() {
...
 				mappingJob := th.GetJob(cell1.CellDeleteJobName)
 				newJobInputHash := GetEnvVarValue(
 					mappingJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
 				g.Expect(newJobInputHash).NotTo(BeNil())
 			}, timeout, interval).Should(Succeed())
+
+			// Check that the cell status is not deleted
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name))
```
auniyal61, line 196: I think (or unless I am forgotten), `nova.Status.RegisteredCells` is nova CR?
and here it checks if cell1 is present in CR template or not 

if yes, it will be removed from template before deletion reconcile or job will start right.

So we may not need to move this.

auniyal61 (commented):
over all lgtm
just an inline question regarding check in CR template.

controllers/common.go
auniyal61: sure, as this my merge first, lets use this.

controllers/nova_controller.go
```diff
@@ -700,29 +693,29 @@ func (r *NovaReconciler) ensureCellDeleted(
 		// We cannot do further cleanup of the MariaDBDatabase and
 		// MariaDBAccount as their name is only available in the NovaCell CR
 		// since the cell definition is removed from the Nova CR already.
-		return nil
+		return nova.CellDeleteFailed, nil
```
gibizer, line 696: I guess if in the past we returned nil here that meant we considered the deletion as success, even though we could not delete the underlying stuff. I think it would be better signalling success here in the result too as this state is not something we will ever recover from as the NovaCell CR is gone.

controllers/nova_controller.go
```diff
@@ -700,29 +693,29 @@ func (r *NovaReconciler) ensureCellDeleted(
...
-		return err
+		return nova.CellDeleteFailed, err
 	}
 	// If it is not created by us, we don't touch it
 	if !OwnedBy(cell, instance) {
 		Log.Info("Cell isn't defined in the Nova, but there is a  "+
 			"Cell CR not owned by us. Not deleting it.",
 			"cell", cell)
-		return nil
+		return nova.CellDeleteFailed, nil
```
gibizer, line 706: ditto I think this is not an Failed case. nova-operator saw a NovaCell that is not in the cellTemplates but also it is not owned by Nova so we just don't touch it.

test/functional/nova_reconfiguration_test.go
```diff
@@ -183,18 +183,28 @@ var _ = Describe("Nova reconfiguration", func() {
...
-			NovaCellNotExists(cell1.CellCRName)
 			Eventually(func(g Gomega) {
 				mappingJob := th.GetJob(cell1.CellDeleteJobName)
 				newJobInputHash := GetEnvVarValue(
 					mappingJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
 				g.Expect(newJobInputHash).NotTo(BeNil())
 			}, timeout, interval).Should(Succeed())
+
+			// Check that the cell status is not deleted
+			Eventually(func(g Gomega) {
```
gibizer, line 194: this should be Consistently as the state asserted here should be true until later we simulate the job success

gibizer (changes requested):
couple small requests but overall the code looks good

controllers/nova_controller.go
mrkisaolamb: done

controllers/nova_controller.go
mrkisaolamb: done

test/functional/nova_reconfiguration_test.go
mrkisaolamb: done

mrkisaolamb:
FYI @gibizer this will solve only 'Delete NovaCell cr only when deletion job pass' for second issue with readines of nova I will prepare separate patch to not mix it

gibizer (approved):
Looks good to me. 

I tested it locally with the following reproduction:

* deploy 18.0 with 2 EDPM node to cell1
* delete cell1 from the OpenStackControlPlane CR
* observe that the cell deletion job fails as the nova-manage command rejectes the cell deletion as computes exists in the cell
* observe that NovaCell CR is not delete
* observe that cell1 is not removed from the Nova.Status.RegisteredCells
* delete the two EDPM nodes from cell1
* wait until the cell deletion job is retried and succeeds
* observer that NovaCell CR is deleted
* obsever that cell1 is removed from the Nova.Status.RegisteredCells field.

Thanks @mrkisaolamb!

mrkisaolamb:
recheck

test/functional/nova_reconfiguration_test.go
```diff
@@ -183,18 +183,28 @@ var _ = Describe("Nova reconfiguration", func() {
...
 				newJobInputHash := GetEnvVarValue(
 					mappingJob.Spec.Template.Spec.Containers[0].Env, "INPUT_HASH", "")
 				g.Expect(newJobInputHash).NotTo(BeNil())
 			}, timeout, interval).Should(Succeed())
+
+			// Check that the cell status is not deleted
+			Consistently(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name))
+			}, timeout, interval).Should(Succeed())
```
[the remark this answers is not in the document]
mrkisaolamb: Thanks will fix with other patch touching statuses

test/functional/nova_reconfiguration_test.go
auniyal61: missed `NotTo(Have)` --> `To(Have)`,
later at L200 deletion job succeeds and then

at L203 confirmed that removed from CR
at L206 confirmed that removed from k8s

auniyal61:
recheck subnet creation failed for deployment

mrkisaolamb:
recheck

gibizer (approved):
It was just a rebase, still looks god

auniyal61 (approved):
lgtm

auniyal61:
recheck 
rdoproject.org/github-check nova-operator kuttl failed on `make openstack_wait` during deployment

mrkisaolamb:
recheck

gibizer (approved):

PR #920, opened 2025-01-23
add oslo_limit config

SeanMooney:
This change add supprot for configuring the [oslo_limit] section.
oslo.limit is used by the optional unifed limits quota driver.

This change relys on the automic endpoint discovery intoduced in
oslo.limit 2.6.0

Related: OSPRH-9518


SeanMooney:
adding a hold to this because it currently wont work.
we can proceed with this if we either backprot 

https://github.com/openstack/oslo.limit/commit/9575a24796e019fd66f1bb1a5ef0bcbfc167351a

or i can port the endpoint lookup code form the glance operator

https://github.com/openstack-k8s-operators/glance-operator/blob/72c189b41a6f71b4ea3e349b9a4fc8c100dd45f9/controllers/glanceapi_controller.go#L1401-L1429

added in 
https://github.com/openstack-k8s-operators/glance-operator/commit/f08e23a357bfa818afa06b195cad40cea6cce85a

SeanMooney:
the kuttl test failures seam unrelated

case.go:398: Nova.nova.openstack.org "nova-kuttl" is invalid: [spec.apiContainerImageURL: Required value, spec.computeContainerImageURL: Required value, spec.conductorContainerImageURL: Required value, spec.metadataContainerImageURL: Required value, spec.novncproxyContainerImageURL: Required value, spec.schedulerContainerImageURL: Required value]

SeanMooney:
rdo-check

SeanMooney:
check-rdo

gibizer:
I will wait for @jamepark4 testing this together with the unified limit backports on nova 18.0 before approving it. I will try to help @jamepark4 getting nova-operator build including this.

templates/nova.conf
```diff
@@ -326,6 +326,18 @@ project_name = service
 username = {{ .nova_keystone_user }}
 password = {{ .nova_keystone_password }}
 
+[oslo_limit]
+system_scope = all
+endpoint_interface = internal
+endpoint_service_type = compute
+auth_url = {{ .keystone_internal_url }}
+auth_type = password
+project_domain_name = {{ .default_project_domain }}
```
jamepark4, line 335: In my testing if both project_name/project_domain_name and system_scope are set it results in the keystone error: `keystoneauth1.exceptions.auth.AuthorizationFailure: Authentication cannot be scoped to multiple targets. Pick one of: project, domain, trust, system or unscoped`.  Removing project_domain_name/project_name and leaving system_scope was all that I needed.

melwitt: +1
* We need `system_scope = all` so that the nova service user can read project limits from other projects
* Keystone auth rules do not allow multiple scopes at a time -- so we cannot include `project_name` and `project_domain_name`

melwitt:
> I will wait for @jamepark4 testing this together with the unified limit backports on nova 18.0 before approving it. I will try to help @jamepark4 getting nova-operator build including this.

Before approving we also need to know whether we are backporting the oslo.limit discovery necessary for the current proposed code or not. Because if we are not backporting, then this needs to be updated to pull the endpoint_id in the same manner as glance-operator, as @SeanMooney mentioned earlier.

SeanMooney:
so backporting the oslo.limit feature will not be enough because we also need to change nova.

i would suggest the following path forward.

1.) proceed with the pr and document the need to set the endpoint_id after the initial deployment as a known_issue.
2.a) in a separate pr prot the endpoint discovery logic form glance operator.
2.b) update upstream nova to not need the endpoint id and depend on the new oslo.limit functionality.
3.) decided if we want to backport the nova/oslo.limits change to make the edpoint_id not required.

personally i think 1 is the MVP and we should aim for 2.a in fr 2 or fr 3 dependignin time.
2.b and 3 are nice to have but we need to do upstream work in nova which wont happen until 2025.2 so we should wait until that is done to decide how we will use this going forward.

the operator is generating a config that will work with the oslo.limit backprot if its present so it should not need future modifications outside of perhaps removing the endpoint_discovery code if we implemented 2.a.

templates/nova.conf
```diff
@@ -326,6 +326,16 @@ project_name = service
 username = {{ .nova_keystone_user }}
 password = {{ .nova_keystone_password }}
 
+[oslo_limit]
+system_scope = all
+endpoint_interface = internal
+endpoint_service_type = compute
```
SeanMooney, line 332: this line will ensure that if we have the oslo.limit feature to auto discover the endpoint we would need to populate it.

melwitt: I assume we don't need it for our deployments but just wanted to note there is one more config option that we _could_ set if needed: `endpoint_region_name`.

melwitt:
> so backporting the oslo.limit feature will not be enough because we also need to change nova.

Right that is true -- please ignore my earlier comment.

> i would suggest the following path forward.
> 
> 1.) proceed with the pr and document the need to set the endpoint_id after the initial deployment as a known_issue. 2.a) in a separate pr prot the endpoint discovery logic form glance operator. 2.b) update upstream nova to not need the endpoint id and depend on the new oslo.limit functionality. 3.) decided if we want to backport the nova/oslo.limits change to make the edpoint_id not required.
> 
> personally i think 1 is the MVP and we should aim for 2.a in fr 2 or fr 3 dependignin time. 2.b and 3 are nice to have but we need to do upstream work in nova which wont happen until 2025.2 so we should wait until that is done to decide how we will use this going forward.
> 
> the operator is generating a config that will work with the oslo.limit backprot if its present so it should not need future modifications outside of perhaps removing the endpoint_discovery code if we implemented 2.a.

I agree with this plan.

melwitt (commented):
Changes look good.

gibizer:
> so backporting the oslo.limit feature will not be enough because we also need to change nova.
> 
> i would suggest the following path forward.
> 
> 1.) proceed with the pr and document the need to set the endpoint_id after the initial deployment as a known_issue. 2.a) in a separate pr prot the endpoint discovery logic form glance operator. 2.b) update upstream nova to not need the endpoint id and depend on the new oslo.limit functionality. 3.) decided if we want to backport the nova/oslo.limits change to make the edpoint_id not required.
> 
> personally i think 1 is the MVP and we should aim for 2.a in fr 2 or fr 3 dependignin time. 2.b and 3 are nice to have but we need to do upstream work in nova which wont happen until 2025.2 so we should wait until that is done to decide how we will use this going forward.
> 

I agree. I'm OK to land this PR for FR2 with a known issue documenting how to get and configure the endpoint ID via customeServiceConfig.

templates/nova.conf
SeanMooney: we do set region_name = {{ .openstack_region_name }} in the other config section so i can add it

we do not support multiple regions today so the region is always a hardcoded value in reality there will only be one created.

that might change in the future so it does not hurt to add this now

gibizer:
I've built https://github.com/openstack-k8s-operators/nova-operator/commit/2eb7b85af0cf2f217ea62161fa647d002658bd94
into the latest openstack-operator: quay.io/gibizer/openstack-operator-index:v0.0.490

jamepark4:
Testing downstream with this incorporated.  No issues were found and everything was configured as needed to validate limits were operating as expected.

gibizer (approved):
Based on @jamepark4's feedback I'm good with this.

SeanMooney:
rdo-check

SeanMooney:
check-rdo

PR #921, opened 2025-02-05
Enable defaulting of nova in kuttle job

mrkisaolamb:
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2735
Depends-On: https://github.com/openstack-k8s-operators/nova-operator/pull/927
close: https://issues.redhat.com/browse/OSPRH-14126

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -9,12 +9,20 @@
       This job deploy a basic "Compute Starter Kit" topology
       https://www.openstack.org/software/sample-configs#compute-starter-kit
       that is the minium set of openstack services required to boot a vm.
+    post-run:
+      - ci/nova-operator-base/playbooks/collect-logs.yaml
     vars:
       collection_namespace_override: "nova-kuttl-default"
       operator_name: nova-operator
+      nova_repo: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/nova-operator"
       zuul_log_collection: true
-    post-run:
-      - ci/nova-operator-base/playbooks/collect-logs.yaml
+      nova_hook: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/nova-operator'].src_dir }}/ci/nova-operator-kuttl/deploy_webhooks.yaml"
+      pre_kuttl_from_operator:
+        - name: 70 Create nova webhooks
+          type: playbook
+          source: "{{ nova_hook }}"
+          extra_vars:
+            nova_catalog_image: "{{ content_provider_registry_ip }}:5001/openstack-k8s-operators/nova-operator-index:{{ zuul.patchset }}"
     extra-vars:
       crc_ci_bootstrap_networking:
         networks:
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ DEFAULT_IMG ?= quay.io/openstack-k8s-operators/nova-operator:latest
 IMG ?= $(DEFAULT_IMG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.29
+CATALOG_IMAGE ?= quay.io/openstack-k8s-operators/nova-operator-index:latest
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -428,3 +429,11 @@ PHONY: crd-schema-check
 crd-schema-check: manifests
 	INSTALL_DIR=$(LOCALBIN) CRD_SCHEMA_CHECKER_VERSION=$(CRD_SCHEMA_CHECKER_VERSION) hack/build-crd-schema-checker.sh
 	INSTALL_DIR=$(LOCALBIN) BASE_REF="$${PULL_BASE_SHA:-$(BRANCH)}" hack/crd-schema-checker.sh
+
+
+.PHONY: run_with_olm
+run_with_olm: export CATALOG_IMG=${CATALOG_IMAGE}
+run_with_olm: ## Install nova operator via olm
+	bash ci/olm.sh
+	oc apply -f ci/olm.yaml
+	timeout 300s bash -c "while ! (oc get csv -n openstack-operators -l operators.coreos.com/nova-operator.openstack-operators -o jsonpath='{.items[*].status.phase}' | grep Succeeded); do sleep 1; done"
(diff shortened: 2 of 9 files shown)


mrkisaolamb:
recheck

mrkisaolamb:
recheck

ci/olm.yaml
```diff
@@ -0,0 +1,34 @@
+---
```
bogdando, line 1: no need to duplicate this in olm.sh

.zuul.yaml
```diff
@@ -224,4 +227,12 @@
             vars:
               cifmw_repo_setup_branch: master
         - nova-operator-tempest-multinode-ceph: *job_vars
+        - nova-operator-kuttl:
+            override-checkout: main
```
cescgina, line 231: @mrkisaolamb can you move the `override-checkout` parameter to line 187, that is the job that is actually run in the PR. According to the [zuul docs](https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.override-checkout) this might solve the problem in this case. also the required-projects and vars are not needed here

mrkisaolamb:
recheck

test/kuttl/test-suites/default/cell-tests/01-assert.yaml
```diff
@@ -138,12 +138,10 @@ metadata:
   name: nova-kuttl-api-0
 status:
   containerStatuses:
-  - image: quay.io/podified-antelope-centos9/openstack-nova-api:current-podified
```
SeanMooney, line 141: ah you removed this becuase it may come form the content provider i assume

ya that makes sense to remove

SeanMooney (commented):
/lgtm

i agree with these updates we should merge this once the depnecy is merged

SeanMooney:
we need to do one more rdo un with rdo-check too because the current one has #927 included

.zuul.yaml
```diff
@@ -181,7 +189,8 @@
     github-check:
       jobs:
         - openstack-meta-content-provider
-        - nova-operator-kuttl
+        - nova-operator-kuttl:
+            override-checkout: main
```
gibizer, line 193: what does this mean? Will this prevent testing the content of the nova-operator PR the job is triggered on?

Makefile
```diff
@@ -428,3 +429,11 @@ PHONY: crd-schema-check
 crd-schema-check: manifests
 	INSTALL_DIR=$(LOCALBIN) CRD_SCHEMA_CHECKER_VERSION=$(CRD_SCHEMA_CHECKER_VERSION) hack/build-crd-schema-checker.sh
 	INSTALL_DIR=$(LOCALBIN) BASE_REF="$${PULL_BASE_SHA:-$(BRANCH)}" hack/crd-schema-checker.sh
+
+
+.PHONY: nova
```
gibizer, line 434: I would use a bit more descriptive name. We have a `run` target that runs the operator from the git repo. So this would be something like `run_with_olm`

ci/nova-operator-kuttl/deploy_webhooks.yaml
```diff
@@ -0,0 +1,17 @@
+- name: Create custom service
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  gather_facts: false
+  tasks:
+    - name: Scale down deployment
+      ansible.builtin.shell: |
+        oc patch csv -n openstack-operators openstack-operator.v0.0.1 --type json -p="[{"op": "replace", "path": "/spec/install/spec/deployments/0/spec/replicas", "value": "0"}]"
```
gibizer, line 10: I'm wondering if we can always rely on the csv name being v0.0.1

ci/nova-operator-kuttl/deploy_webhooks.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - name: Install Nova Operator
+      cifmw.general.ci_script:
+        output_dir: "{{ cifmw_basedir }}/artifacts"
+        chdir: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/nova-operator"
+        script: make nova
+        extra_args:
+          CATALOG_IMAGE: "{{ nova_catalog_image | default('quay.io/openstack-k8s-operators/nova-operator-index:latest') }}"
```
gibizer, line 17: I guess nova_catalog_name is here to ensure that the nova-operator that is built from this PR will be used for the deployment. Who sets this variable?

gibizer, line 17: The default here seems to be a duplication as CATALOG_IMAGE is already defaulted in the makefile.

.zuul.yaml
bogdando: IIUC https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.override-checkout right, that picks the main branches for **other** involved projects

ci/nova-operator-kuttl/deploy_webhooks.yaml
bogdando: That is a good question, but CSV versioning questions are in flux as yet. Let's copy that other projects do for the time being.

ci/nova-operator-kuttl/deploy_webhooks.yaml
mrkisaolamb: done

gibizer:
Kuttl run looks green https://softwarefactory-project.io/zuul/t/rdoproject.org/build/453f64801ce749e49ef658a399731fa2/log/controller/controller/kuttl-test-run.out . We need to wait for the dependencies to land.

SeanMooney:
in #928 i have taken this pr and added a patch to drop the edpm josb and enable debug on the pipeline i also have removed #921

that should give us more info about why its not working without that.

SeanMooney:
im not 100% sure but i think the actually issue is how we configure implied branches
so we could trigger jobs on opendev master
i belive will fix the issue for fr1 and main for now
https://github.com/openstack-k8s-operators/nova-operator/pull/929

and we can then alter how we make the teampels work withopen dev on main
there are other ways to make that towrk i think but i belive that the cause of the current issue with fr1 job content running on main

we are declaring on fr1 that the job definition fr1 is actually for main

https://zuul-ci.org/docs/zuul/latest/config/pragma.html#attr-pragma.implied-branches

https://imgur.com/a/cjtmZL6

zuul should not have 2 job definition with the same name of rht same branch like that
the fr2 variant should have fr2 as the branch 

I'm 99% sure that because of

- pragma:
    implied-branch-matchers: True
    implied-branches:
      - main
      - master

mrkisaolamb:
recheck

mrkisaolamb:
recheck

SeanMooney (approved):
/lgtm  we agreed to adress the underlying issue in a follow up.

ill remove hold when the dependcies are merged

PR #922, opened 2025-02-10
Correct defaulting of timeout

mrkisaolamb:
Add missing defaulting logic called by the OpenStack operator.

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -178,6 +178,7 @@ func SetupDefaults() {
 		APIContainerImageURL:       util.GetEnvVar("RELATED_IMAGE_NOVA_API_IMAGE_URL_DEFAULT", NovaAPIContainerImage),
 		SchedulerContainerImageURL: util.GetEnvVar("RELATED_IMAGE_NOVA_SCHEDULER_IMAGE_URL_DEFAULT", NovaSchedulerContainerImage),
 		NovaCellDefaults:           novaCellDefaults,
+		APITimeout:                 60,
 	}
 
 	SetupNovaDefaults(novaDefaults)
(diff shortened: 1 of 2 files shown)


api/v1beta1/nova_webhook.go
```diff
@@ -92,6 +93,10 @@ func (spec *NovaSpecCore) Default() {
 		spec.MetadataServiceTemplate.Enabled = ptr.To(true)
 	}
 
+	if spec.APITimeout == 0 {
```
auniyal61, line 96: why we need this ?
in CR bydefaut its 60, do you mean if operator/user set it 0 ?

api/v1beta1/common_types.go
```diff
@@ -172,6 +172,7 @@ func SetupDefaults() {
 		APIContainerImageURL:       util.GetEnvVar("RELATED_IMAGE_NOVA_API_IMAGE_URL_DEFAULT", NovaAPIContainerImage),
 		SchedulerContainerImageURL: util.GetEnvVar("RELATED_IMAGE_NOVA_SCHEDULER_IMAGE_URL_DEFAULT", NovaSchedulerContainerImage),
 		NovaCellDefaults:           novaCellDefaults,
+		APITimeout:                 60,
```
auniyal61, line 175: can you please tell, in description, why this is added now, 
my question is because it already present in CR
do we get any bug

api/v1beta1/nova_webhook.go
mrkisaolamb: This is for use by openstack-operator

mrkisaolamb:
recheck

auniyal61 (approved):
lgtm

PR #926, opened 2025-02-19
Removes unused openstack_cacerta var

auniyal61:
Removes unused openstack_cacerta var from nova-api conf templateParameters

closes: [OSPRH-9741](https://issues.redhat.com//browse/OSPRH-9741)


mrkisaolamb (approved):
lgtm

PR #927, opened 2025-02-19
Use cifmw kuttl job as parent job for reusability

mrkisaolamb:


SeanMooney (changes requested):
zuul is not runnng the kuttl tests or other jobs it should be on the fr1 branch

so that needs to be fixed too

mrkisaolamb:
> zuul is not runnng the kuttl tests or other jobs it should be on the fr1 branch
> 
> so that needs to be fixed too

Can we fix this in some follow up?  I propose this only because of issue in https://github.com/openstack-k8s-operators/nova-operator/pull/921 . We are block now on main branch so IMO will be nice to unblock it first especialy that we have few days to land code before FR2 branch will be created

mrkisaolamb:
I create issue to handle not running kuttl jobs: https://issues.redhat.com/browse/OSPRH-14211

gibizer (commented):
This is undoing a bunch of things from @SeanMooney so I would like to defer to them for approval. 

In general I'm OK to use a central kuttl base job for easier maintainability if we don't loose too much functionality (i.e. we can still use our existing kuttl coverage)

SeanMooney:
@gibi ya this is just backportign the change to use the functionality port to the ci framework.
i don't know if those change were also proted to the fr1 branch or ci framework but we extracted the common code when we were spinning up the watcher operator.

understandigy why this is need is the real problem.

we should really fix that.

we can backport this but since zuul is not running kuttle on fr1 we don't know that this is correct.

SeanMooney (approved):
/lgtm

PR #931, opened 2025-02-27
Switch Topology defaultLabelSelector to lib-common GetLabelSelector

fmount:
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1312


gibizer (commented):
Is it a bugfix or just a refactor? If the former then some test coverage would be nce

mrkisaolamb:
I think we test it in some places but not all posible eg we test dor https://github.com/openstack-k8s-operators/nova-operator/blob/9c6c30c7718ef5d7f5d5abaed0b12b99ceb04c79/test/functional/nova_novncproxy_test.go#L259 but not for scheduler

mrkisaolamb:
I'm ok with this changes because IMO this is tested but lets wait for @gibizer rsponse

fmount:
recheck

PR #936, opened 2025-03-10
Complete webhook validation for topology references

fmount:
This patch is a follow up of [1] in the `validation webhook` area. It enhances the validation workflow and it aims to complete the pattern by distributing the `ValidateTopology` logic across the API instances. 
The key improvements include:
- `Distributed validation`: it provides a ValidateTopology function across all API instances, ensuring consistent validation throughout the system
- `Optimized validation calls`: it refines how validation is triggered and reduce code duplication by using the existing pattern implemented for Metadata service
- `Complete webhook implementation`: it adds a dedicated `webhook` functions for each sub component

This patch leverages the centralized topology validator introduced in infra-operator [2].
Testing for each component has been added with `envTest` validation.

Depends-On: https://github.com/openstack-k8s-operators/infra-operator/pull/356

[1] https://github.com/openstack-k8s-operators/nova-operator/pull/924
[2] https://github.com/openstack-k8s-operators/infra-operator/pull/356


mrkisaolamb (approved):

api/v1beta1/nova_webhook.go
```diff
@@ -227,19 +218,18 @@ func (r *NovaSpecCore) ValidateCellTemplates(basePath *field.Path, namespace str
...
-			if computeTemplate.TopologyRef != nil {
-				errors = append(
-					errors, computeTemplate.ValidateComputeTopology(
-					cellPath.Child("novaComputeTemplates").Key(computeName),
-					namespace))
-			}
+			errors = append(
+				errors, cell.NoVNCProxyServiceTemplate.ValidateTopology(
+					cellPath.Child("novaComputeTemplates").Child("topologyRef"), namespace)...,
+			)
```
gibizer, line 224: this feels to be out of place. This loops checks the (ironic) compute templates so we should not check cell level VNCProxy here.

gibizer (changes requested):

gibizer:
I was two quick pressing the button to publish my review, I'm still looking at this patch...

api/v1beta1/nova_webhook.go
fmount: right, thank you, I'm going to fix it in my next update.

api/v1beta1/novacompute_webhook.go
gibizer, line 1: this looks good.

gibizer (changes requested):
I finished reviewing the code change. I have to check the tests at a later time.

api/v1beta1/novacell_webhook.go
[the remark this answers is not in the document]
fmount: np at all, I think I've missed that because I didn't think about the possibility of having a Cell CR created as a standalone one. I'm going to update this part as well and make sure we validate the subCRs (conductor, novnc, meta and compute) as part of the cell validation.

api/v1beta1/nova_webhook.go
```diff
@@ -170,23 +169,15 @@ func (r *NovaSpecCore) ValidateCellTemplates(basePath *field.Path, namespace str
...
-				errors,
-				cell.NoVNCProxyServiceTemplate.ValidateNoVNCProxyTopology(
-					cellPath.Child("noVNCProxyServiceTemplate"),
-					namespace,
-			))
-		}
+		errors = append(errors,
+			cell.NoVNCProxyServiceTemplate.ValidateTopology(
+				cellPath.Child("noVNCProxyServiceTemplate").Child("topologyRef"),
+				namespace)...)
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/nova_webhook.go
fmount: Done

api/v1beta1/novacell_webhook.go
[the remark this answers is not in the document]
fmount: Ok, check L125-L146 to see if this is what we expect as a cell level validation.

api/v1beta1/nova_webhook.go
```diff
@@ -253,9 +243,26 @@ func (r *NovaSpecCore) ValidateAPIServiceTemplate(basePath *field.Path) field.Er
...
 	return errors
 }
 
+// ValidateSchedulerServiceTemplate -
+func (r *NovaSpecCore) ValidateSchedulerServiceTemplate(basePath *field.Path, namespace string) field.ErrorList {
+	errors := field.ErrorList{}
+	// validate the referenced TopologyRef
+	errors = append(errors,
+		r.SchedulerServiceTemplate.ValidateTopology(
+			basePath.Child("apiServiceTemplate").Child("topologyRef"),
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/nova_webhook.go
```diff
@@ -140,11 +140,10 @@ func (r *NovaSpecCore) ValidateCellTemplates(basePath *field.Path, namespace str
...
 			ValidateCellName(cellPath, name)...,
 		)
-		if cell.TopologyRef != nil {
-			if err := topologyv1.ValidateTopologyNamespace(cell.TopologyRef.Namespace, *cellPath, namespace); err != nil {
-				errors = append(errors, err)
-			}
-		}
+		// verify the topology namespace is valid
+		errors = append(errors, topologyv1.ValidateTopologyRef(
+			r.TopologyRef, *basePath.Child("topologyRef"), namespace)...)
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/nova_webhook.go
```diff
@@ -266,13 +273,25 @@ func (r *NovaSpec) ValidateCreate(basePath *field.Path, namespace string) field.
...
 func (r *NovaSpecCore) ValidateCreate(basePath *field.Path, namespace string) field.ErrorList {
 	errors := r.ValidateCellTemplates(basePath, namespace)
-	errors = append(errors, r.ValidateAPIServiceTemplate(basePath)...)
+	errors = append(errors, r.ValidateAPIServiceTemplate(basePath, namespace)...)
+	errors = append(errors, r.ValidateSchedulerServiceTemplate(basePath, namespace)...)
+
+	// validate TopologyRef override for top-level MetadataServiceTemplate
+	errors = append(errors,
+		r.MetadataServiceTemplate.ValidateTopology(
+			basePath.Child("metadataServiceTemplate").Child("topologyRef"),
```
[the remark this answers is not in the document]
fmount: Done (resolving the thread because github always helps me missing things, but feel free to reopen if you think I didn't solve this).

api/v1beta1/nova_webhook.go
```diff
@@ -170,23 +169,15 @@ func (r *NovaSpecCore) ValidateCellTemplates(basePath *field.Path, namespace str
...
-			errors = append(
-				errors,
-				cell.NoVNCProxyServiceTemplate.ValidateNoVNCProxyTopology(
-					cellPath.Child("noVNCProxyServiceTemplate"),
-					namespace,
-			))
-		}
+		errors = append(errors,
+			cell.NoVNCProxyServiceTemplate.ValidateTopology(
+				cellPath.Child("noVNCProxyServiceTemplate").Child("topologyRef"),
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/nova_webhook.go
```diff
@@ -300,13 +319,24 @@ func (r *NovaSpec) ValidateUpdate(old NovaSpec, basePath *field.Path, namespace
...
+	errors = append(errors, topologyv1.ValidateTopologyRef(
+		r.TopologyRef, *basePath.Child("topologyRef"), namespace)...)
+
+	errors = append(errors, r.ValidateAPIServiceTemplate(basePath, namespace)...)
+	errors = append(errors, r.ValidateSchedulerServiceTemplate(basePath, namespace)...)
+
+	// validate TopologyRef override for top-level MetadataServiceTemplate
+	errors = append(errors,
+		r.MetadataServiceTemplate.ValidateTopology(
+			basePath.Child("metadataServiceTemplate").Child("topologyRef"),
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/novametadata_webhook.go
```diff
@@ -86,11 +86,16 @@ var _ webhook.Validator = &NovaMetadata{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *NovaMetadata) ValidateCreate() (admission.Warnings, error) {
 	novametadatalog.Info("validate create", "name", r.Name)
+	errors := field.ErrorList{}
+	basePath := field.NewPath("spec")
 
-	errors := ValidateMetadataDefaultConfigOverwrite(
+	errors = ValidateMetadataDefaultConfigOverwrite(
 		field.NewPath("spec").Child("defaultConfigOverwrite"),
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/novacell_webhook.go
```diff
@@ -158,8 +159,13 @@ func (r *NovaCellSpec) ValidateCreate(basePath *field.Path) field.ErrorList {
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *NovaCell) ValidateCreate() (admission.Warnings, error) {
 	novacelllog.Info("validate create", "name", r.Name)
+	errors := field.ErrorList{}
+	basePath := field.NewPath("spec")
 
-	errors := r.Spec.ValidateCreate(field.NewPath("spec"))
+	errors = r.Spec.ValidateCreate(basePath)
+
+	errors = append(errors, topologyv1.ValidateTopologyRef(
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/novaconductor_webhook.go
```diff
@@ -85,9 +86,14 @@ var _ webhook.Validator = &NovaConductor{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *NovaConductor) ValidateCreate() (admission.Warnings, error) {
 	novaconductorlog.Info("validate create", "name", r.Name)
+	errors := field.ErrorList{}
+	basePath := field.NewPath("spec")
 
-	errors := r.Spec.DBPurge.Validate(
-		field.NewPath("spec").Child("dbPurge"))
+	errors = r.Spec.DBPurge.Validate(
```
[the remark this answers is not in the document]
fmount: Done

api/v1beta1/novaconductor_webhook.go
```diff
@@ -101,15 +107,21 @@ func (r *NovaConductor) ValidateCreate() (admission.Warnings, error) {
...
 	oldConductor, ok := old.(*NovaConductor)
 	if !ok || oldConductor == nil {
 		return nil, apierrors.NewInternalError(fmt.Errorf("unable to convert existing object"))
 	}
 
 	novaconductorlog.Info("validate update", "diff", cmp.Diff(oldConductor, r))
 
-	errors := r.Spec.DBPurge.Validate(
-		field.NewPath("spec").Child("dbPurge"))
+	errors = r.Spec.DBPurge.Validate(
```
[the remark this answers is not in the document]
fmount: Done

test/functional/validation_webhook_test.go
```diff
@@ -1095,49 +1146,35 @@ var _ = Describe("Nova validation", func() {
...
+			raw := map[string]interface{}{
+				"apiVersion": "nova.openstack.org/v1beta1",
+				"kind":       kind,
+				"metadata": map[string]interface{}{
+					"name":      name,
+					"namespace": novaNames.Namespace,
+				},
+				"spec": spec,
+			}
+			unstructuredObj := &unstructured.Unstructured{Object: raw}
```
auniyal61, line 1164: why to convert raw as  `Unstructured` and not Nova resources?

I see this is used in webhook tests mostly.

test/functional/validation_webhook_test.go
```diff
@@ -1095,49 +1146,35 @@ var _ = Describe("Nova validation", func() {
...
-			ContainSubstring(
-				"Invalid value: \"namespace\": Customizing namespace field is not supported"),
-		)
-	})
+	DescribeTable("rejects wrong topology for",
+		func(serviceNameFunc func() (map[string]interface{}, string, string)) {
+			expectedErrorMessage := "spec.topologyRef[namespace]: Invalid value: \"namespace\": Customizing namespace field is not supported"
+
+			spec, kind, name := serviceNameFunc()
+			spec["topologyRef"] = map[string]interface{}{"name": "foo", "namespace": "bar"}
```
auniyal61, line 1154: I know this is a negative scenerio, thats why foo-bar.

but what should be the correct topology name ?
`nova-global-topology` for topLevelEntry ?

from `novaNames.NovaTopologies`

fmount: The point of this test is that regardless the name (`foo` or an existing `nova-global-topology`), if the namespace doesn't match with the current one (in other words it tries to get a topology from a different namespace), it fails in advance. The `name`, in this case, is not relevant because it's not evaluated, hence `foo`. 
If you think it helps to clarify the goal of this test, I can add a comment and change the name from `foo` to an entry of `NovaTopologies`, but I thought `foo` was a good choice to implicitly state "I don't care what the name is".

auniyal61: ack, thanks for clearing, the validation by namespace part, now its clear from infra-patch too: https://github.com/openstack-k8s-operators/infra-operator/pull/356/files

no `foo` serves the purpose, that name isn't whats being checked, so no changes are required.

my intention was to just understand whether topology name value stored in `novaNames.NovaTopologies`.

test/functional/validation_webhook_test.go
fmount: sorry, I think I missed this. we use `Unstructured` object to simulate a CR that is applied by `controllerutil` through the `CreateOrPatch` [1]. It accepts a `client.Object` (that technically can be anything), but here the idea is to simulate a .yaml CR being applied as user input and see the webhook raising the exception. 

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil

SeanMooney: just in case its not clear unstructured.Unstructured{Object: raw} is very important for the webhook test because that is where the serialization/defaulting behavior happens.

we can use golang classes in some of the other test but any time we want to ensure we are emulating doing a `oc apply` with the relevant defaulting ectra we used unstructured 
if we are simulating stateful sets ectra and just want to create the resource with hardcode values we can often just use the golang classes because we are setting all the values that we care about for the given test.

auniyal61: ack thanks

gibizer: Yeah the golang serialization/deserialization between json/yaml and golang structs is complicated. If we stat with a golang struct here as input to the API then we cannot be explicit what we send to the API as our golang struct in the test code will be serialized to json/yaml based on a set of rules that are complicated and sometimes implicit. In the other hand If we use raw then the serialization to json/yaml is a lot simpler.

An example of complicated behavior is simulating that the user sent a json document with an optional field not specified at all. In the raw input we can simply leave out the field to simulate this. With a golang struct as input we cannot leave out a golang struct field, even if the test code is not setting a field in the struct, golang will initialize the field to the null value of the type of the field (i.e. "" for str, 0 for int, null for a pointer, etc). Then when that struct is serialized into json the struct field tags will determine if a filed of the struct is serialized to the json document or left out. If the struct field is defined with an  omitempty tag, and the value of the field is the null value of the type of the field, then the json serializer will leave out the whole field from the json document, otherwise it will add it.

api/v1beta1/novacell_webhook.go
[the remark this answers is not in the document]
gibizer: Looks good. Thanks.

test/functional/validation_webhook_test.go
```diff
@@ -1095,49 +1146,35 @@ var _ = Describe("Nova validation", func() {
...
-			ctx, k8sClient, unstructuredObj, func() error { return nil })
-		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(
-			ContainSubstring(
-				"Invalid value: \"namespace\": Customizing namespace field is not supported"),
-		)
-	})
+	DescribeTable("rejects wrong topology for",
+		func(serviceNameFunc func() (map[string]interface{}, string, string)) {
+			expectedErrorMessage := "spec.topologyRef[namespace]: Invalid value: \"namespace\": Customizing namespace field is not supported"
```
gibizer, line 1151: hm this is strange, it suggests that spec.topologyRef is a map keyed by a namespace. I thin the correct value here would be spec.topologyRef.namespace as this is struct field de-reference all the way down.
//later
Ahh sh*t we missed this in the original review in the infra-operator:
https://github.com/openstack-k8s-operators/infra-operator/blob/03af22a6be4cb66b360b164ddb2dbe2e9d478565/apis/topology/v1beta1/topology_types.go#L127
This should be 
```golang
topologyNamespace := basePath.Child("namespace")
```
As the .Key() means indexing into a map with a Key, but ToplogyRef is a struct not a map.

test/functional/validation_webhook_test.go
```diff
@@ -1095,49 +1146,35 @@ var _ = Describe("Nova validation", func() {
...
-		unstructuredObj := &unstructured.Unstructured{Object: raw}
-		_, err := controllerutil.CreateOrPatch(
-			ctx, k8sClient, unstructuredObj, func() error { return nil })
-		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(
-			ContainSubstring(
-				"Invalid value: \"namespace\": Customizing namespace field is not supported"),
-		)
-	})
+	DescribeTable("rejects wrong topology for",
```
gibizer, line 1149: Nice data driver test case! Thanks.

gibizer (commented):
We have a bug in infra-operator how we generate the path for the field we reporting error on. I can be convinced to land this as is and fix the test case once the infra-operator fix is landed and the version bumped into nova-operator.

gibizer (approved):
Content looks good. There is a need for an infra-operator fix but we will handle that as a separate PR.

test/functional/validation_webhook_test.go
fmount: Fix covered by https://github.com/openstack-k8s-operators/infra-operator/pull/360

PR #940, opened 2025-03-19
Moves sort-by-cell-name func to common place

auniyal61:
Adds tests for same

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -720,7 +720,7 @@ func ensureMemcached(
 	return memcached, err
 }
 
-func sortNovaCellListByName(cellList *novav1.NovaCellList) {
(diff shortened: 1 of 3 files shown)


mrkisaolamb (commented):
This is still valid patch because of additional test, but we need rebase it

controllers/common.go
```diff
@@ -720,7 +720,7 @@ func ensureMemcached(
 	return memcached, err
 }
 
-func sortNovaCellListByName(cellList *novav1.NovaCellList) {
+func SortNovaCellListByName(cellList *novav1.NovaCellList) {
```
auniyal61, line 723: capitalized it, so can be used in test as well.

mrkisaolamb (approved):
lgtm

PR #944, opened 2025-03-19
Bump infra-operator and fix webhook topology validation

fmount:
We recently fixed a bug in the `topologyRef` struct field access [1]. This patch is a follow up that proactively bumps `infra-operator` and fixes the validation webhook as well.

Jira: https://issues.redhat.com/browse/OSPRH-14627

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/360


mrkisaolamb (approved):
lgtm

PR #948, opened 2025-03-27
Add Notfications message bus instance for Nova CRD

bogdando:
Extend CRDs to support a RabbitMqCluster reference as AMQP instance for
notifications.

To provide a way for human user to define the notifications bus for
Nova CR via the OpenStackControlPlane CR's nova section, add a new
field to the NovaSpecCore as that is the struct used when the
OpenStackControlPlaneSpec struct is defined.

At NovaSpec level having just "" or a name of the bus is OK as either
we need a bus, or we don't need notifications at all.  However, for
empedded NovaSpecCore we might need the 3rd possibility to "inherit
from somewhere higher", which we represent as: *string field where null
means inherit, "" means disable notification, "foo" means send
notification to bus foo.

Note that overriding NotificationsBusInstance for cells is not
supported as that implies that consumers of notifications can maintain
simultanous AMQP connections via multiple transport_url endpoints,
which is not a thing in Openstack.

Assume that in the implementation the required information will be
transferred via the existing internal secret via notifications specific
transport_url key (as the transport_url key is already taken by the RPC
bus value). Thus we do not modify other Nova services' CRDs.

Note that the pointer type is redundant for NovaSpec, but since
we have already defined it as a such in NovaSpecCore, and
changing the types system is not planned (requires changes in openstack
operator), it should be used as is for also NovaSpec.
However there shall be only a real use of a nil value in
NovaSpecCore (while unused in NovaSpec).                                                                                                                                                                

Jira: [OSPRH-14486](https://issues.redhat.com//browse/OSPRH-14486)


bogdando:
Note, implementation to follow after we agree on this design

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -390,6 +390,15 @@ spec:
                         the Message Bus Service instance used by the nova services to
                         communicate in this cell. For cell0 it is unused.
                       type: string
+                    cellNotificationsMessageBusInstance:
```
SeanMooney, line 393: This is explict, but maybe we could be less verbose

NotificationsBusInstance ?

The fact its in the cell template implies it the cell one

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -390,6 +390,15 @@ spec:
...
+                      default: rabbitmq
```
SeanMooney, line 394: i would use notifications as the default name too to make it clear that this really should be a separate one just for notifications.

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1633,15 @@ spec:
                   NodeSelector here acts as a default value and can be overridden by service
                   specific NodeSelector Settings.
                 type: object
+              notificationsMessageBusInstance:
```
SeanMooney, line 1636: again here we can shorten it for the top level service to just

NotificationsBusInstance

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1633,15 @@ spec:
...
+                default: rabbitmq
```
SeanMooney, line 1637: and set this to the same default `notifications` to show its fine to have one rabbit for all services  but that it should not be the default one used for rpc.

if people want to collapse it more they can.

auniyal61 (dismissed):

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -1136,6 +1136,15 @@ spec:
                       description: NodeSelector to target subset of worker nodes running
                         cell.
                       type: object
+                    notificationsBusInstance:
+                      default: notifications
```
SeanMooney, line 1140: actully thinking abotu this again. this cannot have a default because it should not be enabled by default.

config/crd/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1633,15 @@ spec:
...
                 type: object
+              notificationsBusInstance:
+                default: notifications
+                description: |-
+                  NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+                  the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+                  Avoid colocating it with APIMessageBugInstance or CellMessageBugInstance used for PRC.
+                  Overriding it with NotificationsBusInstance for cells implies that consumers of notifications
+                  can maintain simultanous AMQP connections via multiple transport_url endpoints.
+                type: string
```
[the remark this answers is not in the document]
SeanMooney: yes you answered your own question

both the global service and per cell service may want to use different location and should not share the same so we need this in two places so that the gloable service like nova-api can send notification to a central rabbot and your edge sights can log to per cell rabits or similar if needed.

either you use 1 or multiple locations is determined by the topology of your cloud.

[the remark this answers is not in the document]
bogdando: Note that we should not implement per cell level overrides yet, but we should have this included into CRD design for future work

api/bases/nova.openstack.org_nova.yaml
bogdando: it is optional, and had a default similarly to other *BusInstances. Please clarify what exactly do you suggest?

config/crd/bases/nova.openstack.org_nova.yaml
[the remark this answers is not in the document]
auniyal61: ack thanks

api/bases/nova.openstack.org_nova.yaml
gibizer: I suggest to default the top level field to "", meaning notifications are not enabled. That would make the behavior backward compatible, i.e. no notification traffic suddenly appear after the update. With the current default value after an update the Nova CR will fail to reconcile as the RabbitMQCluster/notifications resource would not exists. Instead we need to make sure that after an update the deployment behaves exactly the same, so Nova CR is ready and no notifications are emitted.

gibizer: For the cell level override we even have a bit more complication. I suggest to default that to "" meaning **either** that notification configuration is inherited from the top level Nova.Spec or meaning that notification is disabled for this specific cell.

However with a single string variable we cannot express 3 different things:

1. notification config should be inherited from the top level
2. notification should be disabled for this cell
3. notification config should not be inherited from top but use the specific config provided in the value.

The "" string can either be used for 1. or for 2. but not for both at the same time. So we need to be careful which one we choose.
If we need both behavior then we need to use a `*string` where `nul` represents the "just inherit from top" case, and `""` represents the "disable notifictions for this cell" case.

gibizer (changes requested):

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
bogdando: nope, that is supposed to become the implementation for that CRD design change
Do you suggest add more changes to CRD specs?
I didn't plan extend any other CRDs if we could simply manage that on top level

api/bases/nova.openstack.org_nova.yaml
mrkisaolamb: This case sounds similar to what we have with metadata at the top or cell level. So the question is: do we want to cover all three cases that @gibizer mentioned?
If yes, having an additional enable flag could make it more readable and much easier to debug, especially when dealing with default values during json/yaml serialization and deserialization.

bogdando: I am going to follow KISS here, and suggest it as all or nothing configuration option for the time being. We could circle back for the later improvements, when services become ready to recieve notifications from multiple transport URLs concurrently. Until then, we keep it simple.

api/v1beta1/novacell_types.go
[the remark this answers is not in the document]
bogdando: We will follow the established approach https://github.com/openstack-k8s-operators/nova-operator/commit/8d24502461d288057ca4f51203c7974515d689a6#diff-44196e8d90c781ada5aa15eaa74847cce8f40a8fd002a73965451f5cfd8095cdR313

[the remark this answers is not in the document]
bogdando: Sean provided a better example to follow for how inheritance should flow form top level to cells and how overrides can be done at each level, which is topologyRefs. So we do the same here. @mrkisaolamb

api/bases/nova.openstack.org_nova.yaml
gibizer: > I am going to follow KISS here, and suggest it as all or nothing configuration option for the time being. We could circle back for the later improvements, when services become ready to recieve notifications from multiple transport URLs concurrently. Until then, we keep it simple.

Fine, with one addition. The restriction of all or noting, or the limitation to have a single notification transport URL per deployment could be achieved without the cellTemplate level override fields. So you can drop that in the name of KISS as well. 

In general I support all kind of iterative development that works via API backward compatibility rules. 
Adding the cellTemplate level fields later is backward compatible as we can add them later with a meaningful default. 
Changing default value of an existing field might be OK. Changing the type from string to *string might be OK as well as it extends the list of possible values. But in general adding the field later is the most backward compatible move.

> Note that the future CRD design could be extended with the enablement flag as well, without changing the meaning of the empty defaults handling.
> So that an empty cell level default would keep meaning override from top scope. But it would take an additional constraint then: "...until disabled explicitly"

It is a bit more complicated than that. Adding a standalone NotificationEnabled field is backward compatible indeed but very ugly API design. If we want an enabled flag then we should combine it with the message bus name into a struct. But doing that after the message bus field is added is a non backward compatible change. 

Btw, I have no problems with enabled flags over *string fields. So we can go that direction.

Now on the top level field, I just realized that so far we did not considered that the same 3 way logic might be needed if the openstack-operator wants to implement inheritance. E.g. one top level notification bus definition for the whole deployment, then each service operator inherits that, if the user does not override in the service specific template (SpecCore) in the OpenStackControlPlane CR. Then the same situation appears, the user might want to say:
1. inherit the top level bus value, so use `nil`  in the service specific SpecCore field
2. override for a specific bus in this service, so use `my-bus-name` in the service specific SpecCore field
3. disable notifications for the specific service even if it is enabled globally, i.e. use `""'` (or have an equivalent enabled field) in the service specific SpecCore field.

So while in the NovaSpec we are OK to have a string with a two way logic as it is filled after the inheritance decision is made by the openstack-operator, in the NovaSpecCore we would need the 3 way logic. 

Bottom line: 
* If you want to reduce the scope with KISS then I suggest to drop the cellTemplate level override field altogether and keep only the Nova top level field. 
* On the top level 
  * add the `string` field to NovaSpec
  * add a `*string` or a `struct` with enabled flag to NovaSpecCore
Then bring this up with the podified core for review regarding the inheritance behavior on openstack-operator level.

api/v1beta1/novacell_types.go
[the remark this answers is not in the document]
gibizer: So that are the two main ways you mentioned above:
1. pass the data down from the Nova CR to the NovaAPI CR (and other service CRs) via a new field in the NovaAPI CRD. Then this is a CRD change as well which should be part of this CRD design discussion / PR. The topologyRef feature did this.

2. use the internal secret that is already generated by the nova controller and contains the transport_url for the RPC bus and already passed down via the existing secret field in the NovaAPI (and other service CRDs). Then you don't need to touch the service CRDs, but at least mention in the commit message that the information will be transferred via the existing internal secret via some key (maybe `notification_transport_url` as the `transport_url` key is already taken by the  RPC bus value)

api/v1beta1/nova_types.go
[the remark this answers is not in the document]
gibizer: It depends. See the discussion in the novacell_types.go

api/bases/nova.openstack.org_nova.yaml
bogdando: > Now on the top level field, I just realized that so far we did not considered that the same 3 way logic might be needed if the openstack-operator wants to implement inheritance. E.g. one top level notification bus definition for the whole deployment, then each service operator inherits that, if the user does not override in the service specific template (SpecCore) in the OpenStackControlPlane CR

yes, handling rabbitmq instance for notification in OSCP CR spec is gray area yet, out of this design spec, we'll need another tracker for that

api/v1beta1/novacell_types.go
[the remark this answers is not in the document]
bogdando: Thanks for clarifying, the option 2 is preferred path to follow

api/v1beta1/nova_webhook.go
```diff
@@ -80,12 +80,12 @@ func (r *Nova) Default() {
...
 func (spec *NovaSpec) Default() {
 	spec.NovaImages.Default(novaDefaults)
-	spec.NovaSpecCore.Default()
+	spec.NovaSpecBase.Default()
 }
 
 // Default - set defaults for this Nova spec. Expected to be called from
 // the higher level meta operator.
-func (spec *NovaSpecCore) Default() {
+func (spec *NovaSpecBase) Default() {
```
gibizer, line 88: This is acceptable. The openstack-operator still has a NovaSpecCore instance during its default webhook run and using that to call this Default(). It works as NovaSpecCore embeds a NovaSpecBase. If ever we need to default the NotificationsBusInstance in NovaSpecCore then we can add a new function like:
```golang
func (spec *NovaSpecCore) Default() {
// add defaulting for spec.NotificationsBusInstance here
    NovaSpecBase.Default()
}
```
Today we have no such need so not having the function that just forwards the call to the embeded struct is not needed.

api/v1beta1/nova_webhook.go
```diff
@@ -121,7 +121,7 @@ func (spec *NovaSpecCore) Default() {
 
 var _ webhook.Validator = &Nova{}
 
-func (r *NovaSpecCore) ValidateCellTemplates(basePath *field.Path, namespace string) field.ErrorList {
+func (r *NovaSpecBase) ValidateCellTemplates(basePath *field.Path, namespace string) field.ErrorList {
```
gibizer, line 124: yepp the functionality is fully implementable in the new embedded struct.

api/v1beta1/nova_webhook.go
```diff
@@ -267,16 +267,15 @@ func (r *NovaSpecCore) ValidateSchedulerServiceTemplate(basePath *field.Path, na
 	return errors
 }
 
-
 // ValidateCreate validates the NovaSpec during the webhook invocation.
 func (r *NovaSpec) ValidateCreate(basePath *field.Path, namespace string) field.ErrorList {
-	return r.NovaSpecCore.ValidateCreate(basePath, namespace)
+	return r.NovaSpecBase.ValidateCreate(basePath, namespace)
 }
```
gibizer, line 273: similarly if we would need any additional validation on the NovaSpec.NotificationsBusInstance then we could do it here.

api/v1beta1/nova_webhook.go
```diff
@@ -267,16 +267,15 @@ func (r *NovaSpecCore) ValidateSchedulerServiceTemplate(basePath *field.Path, na
...
 
-// ValidateCreate validates the NovaSpecCore during the webhook invocation. It is
+// ValidateCreate validates the NovaSpecBase during the webhook invocation. It is
 // expected to be called by the validation webhook in the higher level meta
 // operator
-func (r *NovaSpecCore) ValidateCreate(basePath *field.Path, namespace string) field.ErrorList {
+func (r *NovaSpecBase) ValidateCreate(basePath *field.Path, namespace string) field.ErrorList {
```
gibizer, line 278: similarly openstack-operator can still call this function on its NovaSpecCore instance as it embeds a NovaSpecBase. If we ever need to validate the NovaSpecCore.NotificationsBusInstance then we can have a NovaSpecCore specific ValidateCreate that does it and then calls this function.

gibizer (commented):
@bogdando awesome progress. Thanks for trying the type split. The result looks good to me. My only remaining request is about a doc nit on the NovaSpec field. Otherwise I think this is ready to be implemented. 

I acknowledge that in this iteration we are intentionally not supporting the cellTemplate level override as we see less user value in that functionality. As we discussed before it can be added later when needed with a backward compatible way.

gibizer:
@stuggi @dprince This is an incoming CRD change for OpenStackControlPlane (via the NovaSpecCore struct).

From CRD size perspective it should not cause a big issue as it is a single string field repeated only twice (once in OpenStackControlPlane CR and once in Nova CR). Later we might add one additional pair of string fields for a cell level override.

Still I would like to ask about opinions regarding consistency. The new field is used to provide the name of a RabbitMqCluster CR to Nova CR to configure nova notifications to be emitted to that bus. (There are different users for such notifications: watcher, ceilometer, external tools). There is one place in the new installer where we already configure notifications via the CRDs, between keystone and barbican. The interface there is the following:
* Keystone is configured with a KeystoneAPI.Spec.RabbitMqClusterName to emit notification
  * https://github.com/openstack-k8s-operators/keystone-operator/blob/dbb51208bbbe4e096d0dd3fa7e6325515069f8e8/api/v1beta1/keystoneapi_types.go#L187
* Barbican is configured with BarbicanKeystoneListener.Spec.TransportURLSecret (a Secret with a tranport_url key) to consume notifications
  * https://github.com/openstack-k8s-operators/barbican-operator/blob/3a222ba31afcf83d7c2469f37dadbc2419afc9ec/api/v1beta1/barbicankeystonelistener_types.go#L51

Right now the nova-operator change follows the server side patter established by KeystoneAPI above as both uses a rabbit CR name as input. I have no information about the client side (watcher / ceilometer) CRD plans at the moment.

Our design goal here is two fold:
* have a way to provide message bus instance that is separate from the RPC bus to avoid overloading the common RPC bus instance.
* have a way to provide a single bus instance for all notification traffic so the notification consumer can easily access all the openstack notifications form a single rabbit. (In nova we have a separate RPC bus per cell)

To achieve the latter I would like to suggest a top level field in OpenStackControlPlaneSpec field that holds a RabbitMQCluster CR name. Then openstack-operator should implement an inheritance mechanism from the top level field to the Nova service specific field based on the following rules:
* if NovaCoreSpec.NotificationBusInstance == `nil` then NovaSpec.NotificationBusInstance <= OpenStackControlPlaneSpec.NotificationBusInstance 
* else NovaSpec.NotificationBusInstance <= NovaCoreSpec.NotificationBusInstance

Then apply a similar pattern to the KeystoneAPI.Spec.RabbitMqClusterName to keep things consistent. Also if other service operators will start allowing enabling notifications then they should follow the same pattern. (At least cinder, glance and manila could as far as I understand)

How do you feel about it?

api/v1beta1/nova_types.go
```diff
@@ -124,13 +125,43 @@ type NovaSpecCore struct {
...
+	NovaSpecBase `json:",inline"`
 
 	// +kubebuilder:validation:Required
 	NovaImages `json:",inline"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=""
+	// NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+	// the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+	// If undefined, the value will be inherited from OpenStackControlPlane.
```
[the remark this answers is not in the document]
bogdando: I understand that, however as I explained above https://github.com/openstack-k8s-operators/nova-operator/pull/948/files/455e1814e32a9ecf8d63a83e997c7f88c4a42aac#diff-054727a2210c83936f9dbf4593c84144547d408661099c15fa62f8ad3604a422R128 this leaks in to CRD doc, and we have to provide this important detail about inheritance in the CRD docs which are users facing.
I had to add this statement, otherwise the documentation looks realy confusing. It is absolutely not obvious in which ways one may end up with notifications driver configured (by inheritance) while nothing was provided in this Nova CR value for it

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1624,16 @@ spec:
                   NodeSelector here acts as a default value and can be overridden by service
                   specific NodeSelector Settings.
                 type: object
+              notificationsBusInstance:
+                default: ""
+                description: |-
+                  NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+                  the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+                  If undefined, the value will be inherited from OpenStackControlPlane.
```
bogdando, line 1632: @gibi this ^^

config/crd/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1624,16 @@ spec:
                   NodeSelector here acts as a default value and can be overridden by service
                   specific NodeSelector Settings.
                 type: object
+              notificationsBusInstance:
+                default: ""
+                description: |-
+                  NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+                  the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+                  If undefined, the value will be inherited from OpenStackControlPlane.
```
bogdando, line 1632: ...and this ^^ @gibizer

bogdando:
@gibizer ci failure looks real
apis/core/v1beta1/openstackcontrolplane_webhook.go:423:63: cannot use *old.Nova.Template (variable of type "github.com/openstack-k8s-operators/nova-operator/api/v1beta1".NovaSpecCore) as "github.com/openstack-k8s-operators/nova-operator/api/v1beta1".NovaSpecBase value in argument to r.Spec.Nova.Template.ValidateUpdate

stuggi:
> @stuggi initially, the proposal is to make it simple - shared notifcations bus, which could point to a dedicated, or already existing (like a key of `spec.rabbitmq.templates` in the same oscp CR) rabbitmq cluster CR name
> 
> > for notifications all components use the same instance, or could there be multiple different for notifications?
> 
> all components will use the same instance, if provided top scope in oscp CR, or override in its `spec.foo.template`
> 
> > each service then will get its own transport_url for this instance?
> 
> they will get the same transport_url, just like they already do for RPC instance

with service I meant nova, keystone, cinder and others which will use notifications. for RPC they request each their own transport_url for the references cluster name, or do we have to share a single one for all of them?

bogdando:
> > @stuggi initially, the proposal is to make it simple - shared notifcations bus, which could point to a dedicated, or already existing (like a key of `spec.rabbitmq.templates` in the same oscp CR) rabbitmq cluster CR name
> > > for notifications all components use the same instance, or could there be multiple different for notifications?
> > 
> > 
> > all components will use the same instance, if provided top scope in oscp CR, or override in its `spec.foo.template`
> > > each service then will get its own transport_url for this instance?
> > 
> > 
> > they will get the same transport_url, just like they already do for RPC instance
> 
> with service I meant nova, keystone, cinder and others which will use notifications. for RPC they request each their own transport_url for the references cluster name, or do we have to share a single one for all of them?

@stuggi  We will share the single one for all, for the beginning. I am planning to propoce oscp CRD extention to cover this scenario, with a possibility of extending it in the future. For example:
```
NotificationsBus:
  enabled: *bool
  RabbitMQClusterName: *string
  // and we can add later to support other AMQP backends than RabbitMQ
  transport_url: *secretRef
```

gibizer:
> > @stuggi @dprince This is an incoming CRD change for OpenStackControlPlane (via the NovaSpecCore struct).
> > From CRD size perspective it should not cause a big issue as it is a single string field repeated only twice (once in OpenStackControlPlane CR and once in Nova CR). Later we might add one additional pair of string fields for a cell level override.
> > Still I would like to ask about opinions regarding consistency. The new field is used to provide the name of a RabbitMqCluster CR to Nova CR to configure nova notifications to be emitted to that bus. (There are different users for such notifications: watcher, ceilometer, external tools). There is one place in the new installer where we already configure notifications via the CRDs, between keystone and barbican. The interface there is the following:
> > 
> > * Keystone is configured with a KeystoneAPI.Spec.RabbitMqClusterName to emit notification
> >   
> >   * https://github.com/openstack-k8s-operators/keystone-operator/blob/dbb51208bbbe4e096d0dd3fa7e6325515069f8e8/api/v1beta1/keystoneapi_types.go#L187
> > * Barbican is configured with BarbicanKeystoneListener.Spec.TransportURLSecret (a Secret with a tranport_url key) to consume notifications
> >   
> >   * https://github.com/openstack-k8s-operators/barbican-operator/blob/3a222ba31afcf83d7c2469f37dadbc2419afc9ec/api/v1beta1/barbicankeystonelistener_types.go#L51
> > 
> > Right now the nova-operator change follows the server side patter established by KeystoneAPI above as both uses a rabbit CR name as input. I have no information about the client side (watcher / ceilometer) CRD plans at the moment.
> 
> I think barbican follows the same externally to configure the rabbitMqClusterName https://github.com/openstack-k8s-operators/barbican-operator/blob/3a222ba31afcf83d7c2469f37dadbc2419afc9ec/api/v1beta1/common_types.go#L34
> 
> internally the transportURLSecret is passed to the components, like the barbicankeystonelistener.

Thanks. I missed that indirection. It is better then, as at every place the external interface is a RabbitMQCluster CR name.

> 
> We should keep consistency, especially of the same parameter is used across multiple components. right now nova uses `*BusInstance`, while other operators have `RabbitMqClusterName`.

That wasn't my point. But I see the difference. I can only agree to change the name in nova-operator if the existing fields are renamed too, but that is backward incompatible. I value the consistency within the nova-operator higher than the consistency across the operators so I suggest to follow nova's pattern in this PR. In general I also consider nova's name better as it does not encode the type of the CR. So if once the message bus will be based on Kafka (or other) solution then the nova's field can be used as is and only the TransportURL abstraction needs to be extended to accept and work with other than a RabbitMqCluster. Also when we have multiple message busses per service like in nova, having a name that tells what the bus is used for instead of telling how the bus is implemented has more value in my eyes.

> 
> > Our design goal here is two fold:
> > 
> > * have a way to provide message bus instance that is separate from the RPC bus to avoid overloading the common RPC bus instance.
> 
> do I get it right that each service then will get its own transport_url for this instance?b

You are correct, most of the services will get it. I guess metadata and novncproxy does not emit notifications so those can be skipped. The transport_url will be passed via the internal secret so no new field is expected on the specific CRDs, like on NovaAPI.


> > * have a way to provide a single bus instance for all notification traffic so the notification consumer can easily access all the openstack notifications form a single rabbit. (In nova we have a separate RPC bus per cell)
> 
> for notifications all components use the same instance, or could there be multiple different for notifications?

The CRD proposed here allows to pass any bus to Nova CR and then each nova service will use that bus to emit notifications on. Later we might add a per cellTemplate level override to support notifications emitted to different buses per cell. However today the use cases we know about ( ceilometer, watcher, external integration for NVMe cleanup) are all relaying on getting all the notifications across the whole cluster from a single bus as they cannot listen to multiple buses in parallel.


> > To achieve the latter I would like to suggest a top level field in OpenStackControlPlaneSpec field that holds a RabbitMQCluster CR name. Then openstack-operator should implement an inheritance mechanism from the top level field to the Nova service specific field based on the following rules:
> > 
> > * if NovaCoreSpec.NotificationBusInstance == `nil` then NovaSpec.NotificationBusInstance <= OpenStackControlPlaneSpec.NotificationBusInstance
> > * else NovaSpec.NotificationBusInstance <= NovaCoreSpec.NotificationBusInstance
> 
> are notifications enabled as soon the bus is configured for it, or are there other settings required. just thinking about to have it disabled and if we need to decide it based on the BusInstance?
> 

We can be flexible here if we want to as the [oslo_messaging_notification] config section allows to configure a transport_url while the driver is set to noop. Our current intention here is to use the value of the NotificationBusInstance to enable/disable the notifications. If the NotificationBusInstance is empty then nova-operator will set `driver=noop`, if NotificationBusInstance is not empty then nova-operator will set `driver=messagingv2` and `transport_url` based on what it get from the related TransportURL CR.

I think the OpenStackControlPlane level global NotificationBusInstance field should be defaulted to "" to keep backward compatibility during upgrade.

NovaSpecCore.NotificationBusInstance will be defaulted to `nil` to signal that by default Nova wants to inherit the global notification bus settings. So by default it will inherit the `""` making the notification sending disabled to keep backward compatibility during upgrade.

> > Then apply a similar pattern to the KeystoneAPI.Spec.RabbitMqClusterName to keep things consistent. Also if other service operators will start allowing enabling notifications then they should follow the same pattern. (At least cinder, glance and manila could as far as I understand)
> > How do you feel about it?
> 
> I think that sounds good and aligns with the concept we have for other global/local set parameters, e.g. the nodeselector and the topology. for them its like if global set and not local service override, pass it down. for this it would be like you said, if a global notification rabbit is set, use it, otherwise use the default rabbit, right?

by default when NovaSpecCore level field is `nil`:
* If global notification rabbit is set, then use it
* if global notification rabbit is set to `""`, then do not emit notifications

but allow to override on NovaSpecCore level with a value or with ""
* if NovaSpecCore level is set to "", then do not emit notifications regardless of the global rabbit value.
* if NovaSpecCore level is et to "foo", then emit notification to bus foo regardless of the global rabbit value.

gibizer:
> > > @stuggi initially, the proposal is to make it simple - shared notifcations bus, which could point to a dedicated, or already existing (like a key of `spec.rabbitmq.templates` in the same oscp CR) rabbitmq cluster CR name
> > > > for notifications all components use the same instance, or could there be multiple different for notifications?
> > > 
> > > 
> > > all components will use the same instance, if provided top scope in oscp CR, or override in its `spec.foo.template`
> > > > each service then will get its own transport_url for this instance?
> > > 
> > > 
> > > they will get the same transport_url, just like they already do for RPC instance
> > 
> > 
> > with service I meant nova, keystone, cinder and others which will use notifications. for RPC they request each their own transport_url for the references cluster name, or do we have to share a single one for all of them?
> 
> @stuggi We will share the single one for all, for the beginning. I am planning to propoce oscp CRD extention to cover this scenario, with a possibility of extending it in the future. For example:
> 
> ```
> NotificationsBus:
>   enabled: *bool
>   RabbitMQClusterName: *string
>   // and we can add later to support other AMQP backends than RabbitMQ
>   transport_url: *secretRef
> ```
> 
> We will start implementing it for Nova only, letting other producers/consumers to catch up based on the accepted design choice

@bogdando this is confusing. The transport_url is hidden behind a proper abstraction already in TransportURL CR. That take the name of a RabbitMQCluster and produces a Secret with a `transport_url` key in it. If we ever want to support other then Rabbit, then we need to extend the TranportURL CR to take other than a RabbitMQCluster name.

Also I suggest to follow the pattern in OpenStackControlPlane, what we propose here in the nova CRD so no `enabled` bool field, just a `string` field with two possible value:
* "" => notification is disabled on the global level
* "foo" => user the RabbitMQCluster/foo to emit notifications on it

stuggi:
> > We should keep consistency, especially of the same parameter is used across multiple components. right now nova uses `*BusInstance`, while other operators have `RabbitMqClusterName`.
> 
> That wasn't my point. But I see the difference. I can only agree to change the name in nova-operator if the existing fields are renamed too, but that is backward incompatible. I value the consistency within the nova-operator higher than the consistency across the operators so I suggest to follow nova's pattern in this PR. In general I also consider nova's name better as it does not encode the type of the CR. So if once the message bus will be based on Kafka (or other) solution then the nova's field can be used as is and only the TransportURL abstraction needs to be extended to accept and work with other than a RabbitMqCluster. Also when we have multiple message busses per service like in nova, having a name that tells what the bus is used for instead of telling how the bus is implemented has more value in my eyes.
> 

agreed, the current naming with its type is probably problematic as soon there will/can be something else in addition to rabbitmq.

> > > Our design goal here is two fold:
> > > 
> > > * have a way to provide message bus instance that is separate from the RPC bus to avoid overloading the common RPC bus instance.
> > 
> > 
> > do I get it right that each service then will get its own transport_url for this instance?b
> 
> You are correct, most of the services will get it. I guess metadata and novncproxy does not emit notifications so those can be skipped. The transport_url will be passed via the internal secret so no new field is expected on the specific CRDs, like on NovaAPI.
> 

sound good. I was not sure if the idea was to share a a single TransportURL across multiple different openstack services, which is different to what we have right now.


> > > * have a way to provide a single bus instance for all notification traffic so the notification consumer can easily access all the openstack notifications form a single rabbit. (In nova we have a separate RPC bus per cell)
> > 
> > 
> > for notifications all components use the same instance, or could there be multiple different for notifications?
> 
> The CRD proposed here allows to pass any bus to Nova CR and then each nova service will use that bus to emit notifications on. Later we might add a per cellTemplate level override to support notifications emitted to different buses per cell. However today the use cases we know about ( ceilometer, watcher, external integration for NVMe cleanup) are all relaying on getting all the notifications across the whole cluster from a single bus as they cannot listen to multiple buses in parallel.
> 

sounds good to me

> > > To achieve the latter I would like to suggest a top level field in OpenStackControlPlaneSpec field that holds a RabbitMQCluster CR name. Then openstack-operator should implement an inheritance mechanism from the top level field to the Nova service specific field based on the following rules:
> > > 
> > > * if NovaCoreSpec.NotificationBusInstance == `nil` then NovaSpec.NotificationBusInstance <= OpenStackControlPlaneSpec.NotificationBusInstance
> > > * else NovaSpec.NotificationBusInstance <= NovaCoreSpec.NotificationBusInstance
> > 
> > 
> > are notifications enabled as soon the bus is configured for it, or are there other settings required. just thinking about to have it disabled and if we need to decide it based on the BusInstance?
> 
> We can be flexible here if we want to as the [oslo_messaging_notification] config section allows to configure a transport_url while the driver is set to noop. Our current intention here is to use the value of the NotificationBusInstance to enable/disable the notifications. If the NotificationBusInstance is empty then nova-operator will set `driver=noop`, if NotificationBusInstance is not empty then nova-operator will set `driver=messagingv2` and `transport_url` based on what it get from the related TransportURL CR.
> 
> I think the OpenStackControlPlane level global NotificationBusInstance field should be defaulted to "" to keep backward compatibility during upgrade.
> 
> NovaSpecCore.NotificationBusInstance will be defaulted to `nil` to signal that by default Nova wants to inherit the global notification bus settings. So by default it will inherit the `""` making the notification sending disabled to keep backward compatibility during upgrade.
> 

ok, I can remember back in the days of OSP7 or 8 or so, there was an issue where notifications got enabled per default, but there was basically no consumer. the notifications queue were growing until the rabbit service died and impacted the controllers. I think we just need to be careful with enable it per default to not run into such a situation again.

> > > Then apply a similar pattern to the KeystoneAPI.Spec.RabbitMqClusterName to keep things consistent. Also if other service operators will start allowing enabling notifications then they should follow the same pattern. (At least cinder, glance and manila could as far as I understand)
> > > How do you feel about it?
> > 
> > 
> > I think that sounds good and aligns with the concept we have for other global/local set parameters, e.g. the nodeselector and the topology. for them its like if global set and not local service override, pass it down. for this it would be like you said, if a global notification rabbit is set, use it, otherwise use the default rabbit, right?
> 
> by default when NovaSpecCore level field is `nil`:
> 
> * If global notification rabbit is set, then use it
> * if global notification rabbit is set to `""`, then do not emit notifications
> 
> but allow to override on NovaSpecCore level with a value or with ""
> 
> * if NovaSpecCore level is set to "", then do not emit notifications regardless of the global rabbit value.
> * if NovaSpecCore level is et to "foo", then emit notification to bus foo regardless of the global rabbit value.

+1

api/v1beta1/nova_types.go
```diff
@@ -124,13 +125,41 @@ type NovaSpecCore struct {
...
+	// For particular Nova cells, notifications cannot be disabled, nor configured differently.
+	NotificationsBusInstance *string `json:"notificationsBusInstance,omitempty"`
+}
+
 // NovaSpec defines the desired state of Nova
 type NovaSpec struct {
+	// NOTE(bogdando): Anything that is only submitted by opentack-operator should be in NovaSpec but not in NovaSpecCore.
+
 	// +kubebuilder:validation:Required
 	NovaSpecCore `json:",inline"`
```
gibizer, line 150: This cannot be NovaSpecCore any more as NovaSpecCore has NotificationsBusInstance and NovaSpec also has NotificationsBusInstance so embedding NovaSpecCore into NovaSpec will create duplicate fields.

gibizer (changes requested):

api/v1beta1/nova_types.go
bogdando: What would you suggest? I didn't plan to make this CRD change requiring any code changes yet, especially in OSCP operator

bogdando: that overcomplicates design, abandoning the types system change is out of scope

gibizer (commented):
Do I understand it correctly that due to the build issues you rolled back the split type solution and went back to having a single new field in NovaSpecCore and keep that embeded into NovaSpec?

If that is the intention then I'm fine with it. The only drawback of that is on the NovaSpec level nil is not a possible value but the type will allow it (as it shared with the NovaSpecCore field where it is an allowed value). So we will need some extra nil checks in the nova-operator impl.

bogdando:
> Do I understand it correctly that due to the build issues you rolled back the split type solution and went back to having a single new field in NovaSpecCore and keep that embeded into NovaSpec?
> 
> If that is the intention then I'm fine with it. The only drawback of that is on the NovaSpec level nil is not a possible value but the type will allow it (as it shared with the NovaSpecCore field where it is an allowed value). So we will need some extra nil checks in the nova-operator impl.

not exactly build issues, but required changes in openstack operator - which goes beyond CRD design for Nova scope
I don't want this blocking the main goal - CRD design

gibizer (approved):
Looks OK to me.

dprince:
Perhaps it was in another commit but will the mechanism here still use the TransportURL from infra-operator to obtain the secret for the notification bus?

auniyal61:
> Perhaps it was in another commit but will the mechanism here still use the TransportURL from infra-operator to obtain the secret for the notification bus?

yes, human-operator will add rabbit in oscp CR, that rabbit will be recognized in Nova as notifications-broadcaster.
yes, transport_url part of this rabbit secret will be from infra-operator.

gibizer:
> Perhaps it was in another commit but will the mechanism here still use the TransportURL from infra-operator to obtain the secret for the notification bus?

Yes, it will.

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1686,6 +1686,11 @@ func (in *NovaSpecCore) DeepCopyInto(out *NovaSpecCore) {
 		*out = new(topologyv1beta1.TopoRef)
 		**out = **in
 	}
+	if in.NotificationsBusInstance != nil {
+		in, out := &in.NotificationsBusInstance, &out.NotificationsBusInstance
+		*out = new(string)
+		**out = **in
+	}
```
auniyal61, line 1693: I think this the place from where default value gona be set on NovaSpecCore.NotificationsBusInstance during ova reconcile at openstack-operator.
is it ?

bogdando: I am not sure, that's an autogenerated code. What exactly is your question here?

SeanMooney: We are not defaulting the values intentionally, and this code has nothing to do with defaulting. 

The "default" is nil, i.e. we declared it as a pointer to string in the NovaCoreSpec
which is unset by default.

on the sub CRs they are defined as strings that will default to the empty string.

but the code that creates those does not exist in this PR.
we will explicitly be setting them to the empty string when we create the cr.
if we wanted to default, that it would happen in the defaulting webhook for the sub CRDs.

SeanMooney: just to be clear 

`in, out := &in.NotificationsBusInstance, &out.NotificationsBusInstance`
this is creating 2 new pointer to NotificationsBusInstance on the input and out object

`*out = new(string)` is allcoating a new string instance on the heap an saving it address into the value of the out pointer.
then 
`**out = **in`
is copying the data from in.NotificationsBusInstance into the out.NotificationsBusInstance field.

this is just implementing deepcopy for i.e. where the data is effectively memcopied into new backing storage rather then just copying the pointers whci would be a shallow copy.

gibizer:
just rebased to resolve the merge conflict

api/v1beta1/novacell_types.go
```diff
@@ -17,10 +17,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
```
SeanMooney, line 23: hum, i guess this moved because of different go-fmt versions?

the end result is identical so not important either way.

SeanMooney (approved):
/lgtm

PR #949, opened 2025-03-27
Improve build and use of crd-schema-checker

bogdando:
When bin directory removed, crd schema checker go build triggers again, which takes extra time.

Save a copy of a built tool into the user home dir to use it afterwards Also do not clone the repo if it is already there.

Fix SHELL for GNU make to use /bin/bash, when building the tool. Otherwise, the make binary uses /bin/sh, and fails to set -o pipefile.

### DIFF
--- a/hack/build-crd-schema-checker.sh
+++ b/hack/build-crd-schema-checker.sh
@@ -4,12 +4,18 @@ set -euxo pipefail
 if [ -f "$INSTALL_DIR/crd-schema-checker" ]; then
     exit 0
 fi
+if [ -f ~/crd-schema-checker ]; then
+    cp ~/crd-schema-checker "$INSTALL_DIR/"
+    exit
+fi
 
 mkdir -p "$INSTALL_DIR/git-tmp"
 git clone https://github.com/openshift/crd-schema-checker.git \
-    -b "$CRD_SCHEMA_CHECKER_VERSION" "$INSTALL_DIR/git-tmp"
+    -b "$CRD_SCHEMA_CHECKER_VERSION" "$INSTALL_DIR/git-tmp" || true
 pushd "$INSTALL_DIR/git-tmp"
+grep -q 'SHELL=/bin/bash' Makefile || sed -i '1 i\SHELL=/bin/bash' Makefile
 GOWORK=off make
+cp crd-schema-checker ~/
 cp crd-schema-checker "$INSTALL_DIR/"
 popd
 rm -rf "$INSTALL_DIR/git-tmp"


hack/build-crd-schema-checker.sh
```diff
@@ -4,12 +4,18 @@ set -euxo pipefail
 if [ -f "$INSTALL_DIR/crd-schema-checker" ]; then
```
SeanMooney, line 4: we are already caching it in the build dir.

hack/build-crd-schema-checker.sh
```diff
@@ -4,12 +4,18 @@ set -euxo pipefail
...
     exit 0
 fi
+if [ -f ~/crd-schema-checker ]; then
```
SeanMooney, line 7: you should not be douchign anything in home or about the root of the repo.
if you want to do extra cachign you should do it within the repo folders.

SeanMooney (changes requested):

hack/build-crd-schema-checker.sh
```diff
@@ -4,12 +4,18 @@ set -euxo pipefail
...
+    cp ~/crd-schema-checker "$INSTALL_DIR/"
+    exit
+fi
 
 mkdir -p "$INSTALL_DIR/git-tmp"
 git clone https://github.com/openshift/crd-schema-checker.git \
-    -b "$CRD_SCHEMA_CHECKER_VERSION" "$INSTALL_DIR/git-tmp"
+    -b "$CRD_SCHEMA_CHECKER_VERSION" "$INSTALL_DIR/git-tmp" || true
 pushd "$INSTALL_DIR/git-tmp"
+grep -q 'SHELL=/bin/bash' Makefile || sed -i '1 i\SHELL=/bin/bash' Makefile
```
SeanMooney, line 16: i dont think we shoudl be modfiying the content of the git repo ether.

hack/build-crd-schema-checker.sh
bogdando: that bin dir is local and the standard troubleshooting for dev laptops (not CI systems) suggests removing this altogether. I simply found this suboptimal as go build consumes a lot of resources we could have saved otherwise

hack/build-crd-schema-checker.sh
bogdando: as I said, git clean -xdff removes it all, and that is the only known thing to fix failing make targets.
For my setup I had to run it each time before committing changes :/

hack/build-crd-schema-checker.sh
bogdando: it doesn't work with GNU make then... ofc we could suggest a fix upstream instead

SeanMooney: cant we just do  `SHELL=$(which bash) GOWORK=off make` below

bogdando: I tried that, Make doesn't take into account the SHELL env var as developers of it considered changing behavior of make based on the user shell where it runs as unsafe

PR #952, opened 2025-04-01
[DNM] Osprh 14486 impl

bogdando:
This is an initial pass for @auniyal61 to take over and continue on

> use the internal secret that is already generated by the nova controller and contains the transport_url for the RPC bus and already passed down via the existing secret field in the NovaAPI (and other service CRDs). Then we don't need to touch the service CRDs. The information will be transferred via the existing internal secret via notifications_transport_url key as the transport_url key is already taken by the RPC bus value.

Jira: [OSPRH-15392](https://issues.redhat.com//browse/OSPRH-15392)

### DIFF
--- a/controllers/novascheduler_controller.go
+++ b/controllers/novascheduler_controller.go
@@ -513,27 +513,29 @@ func (r *NovaSchedulerReconciler) generateConfigs(
 	}
 
 	templateParameters := map[string]interface{}{
-		"service_name":             "nova-scheduler",
-		"keystone_internal_url":    instance.Spec.KeystoneAuthURL,
-		"nova_keystone_user":       instance.Spec.ServiceUser,
-		"nova_keystone_password":   string(secret.Data[ServicePasswordSelector]),
-		"api_db_name":              NovaAPIDatabaseName,
-		"api_db_user":              apiDatabaseAccount.Spec.UserName,
-		"api_db_password":          string(apiDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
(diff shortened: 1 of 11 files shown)


controllers/novascheduler_controller.go
```diff
@@ -513,27 +513,29 @@ func (r *NovaSchedulerReconciler) generateConfigs(
...
+		"cell_db_name":                   NovaCell0DatabaseName,
+		"cell_db_user":                   cellDatabaseAccount.Spec.UserName,
+		"cell_db_password":               string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
+		"cell_db_address":                instance.Spec.Cell0DatabaseHostname,
+		"cell_db_port":                   3306,
+		"openstack_region_name":          "regionOne", // fixme
+		"default_project_domain":         "Default",   // fixme
+		"default_user_domain":            "Default",   // fixme
+		"transport_url":                  string(secret.Data[TransportURLSelector]),
+		"nova_cell_notify_transport_url": string(secret.Data[NotificationsTransportURLSelector]),
```
bogdando, line 534: and so on for remaining top level services

bogdando: notes after the base patch changes (had not rebased yet)

- we will need some extra nil checks in the nova-operator impl.
- we must emit versioned and non-versioned notifications, when enabled (@SeanMooney please correct me)

bogdando:
I am glad this helped with moving forward, closing it now

PR #954, opened 2025-04-03
Cell update status during delete

mrkisaolamb:
resolve: https://issues.redhat.com/browse/OSPRH-10981

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -52,6 +52,8 @@ const (
 	NovaComputeServiceConfigReady condition.Type = "NovaComputeServiceConfigReady"
 	// NovaAllControlPlaneComputesReadyCondition indicates that every defined NovaCompute is ready but undiscovered
 	NovaAllControlPlaneComputesReadyCondition condition.Type = "NovaAllControlPlaneComputesReady"
+	//NovaCellsDeletionCondition indicates that the NovaCells deletion is in progress
+	NovaCellsDeletionCondition condition.Type = "NovaCellsDeletion"
 )
 
 // Common Messages used by API objects.
@@ -172,4 +174,13 @@ const (
 
 	//CellHostDiscoverErrorMessage
 	CellHostDiscoverErrorMessage = "CellHostDiscover error occurred %s"
+
+	// NovaCellsDeletionConditionInitMessage
+	NovaCellsDeletionConditionInitMessage = "NovaCells deletion not started"
+
+	// NovaCellsDeletionMessage
+	NovaCellsDeletionMessage = "NovaCells deletion in progress: %s"
+
+	// NovaCellsDeletionConditionReadyMessage
+	NovaCellsDeletionConditionReadyMessage = "There is no more NovaCells to delete"
 )
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -38,6 +38,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"golang.org/x/exp/maps"
+
 	"github.com/go-logr/logr"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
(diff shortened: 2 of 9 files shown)


api/v1beta1/conditions.go
```diff
@@ -52,6 +52,8 @@ const (
 	NovaComputeServiceConfigReady condition.Type = "NovaComputeServiceConfigReady"
 	// NovaAllControlPlaneComputesReadyCondition indicates that every defined NovaCompute is ready but undiscovered
 	NovaAllControlPlaneComputesReadyCondition condition.Type = "NovaAllControlPlaneComputesReady"
+	//NovaCellsDeketionCondition indicates that the NovaCells deletion is in progress
```
gibizer, line 55: nit: NovaCellsDeketionCondition => NovaCellsDeletionCondition

api/v1beta1/conditions.go
```diff
@@ -172,4 +174,13 @@ const (
...
 	CellHostDiscoverErrorMessage = "CellHostDiscover error occurred %s"
+
+	// NovaCellsDeletionConditionInitMessage
+	NovaCellsDeletionConditionInitMessage = "NovaCells deletion not started"
+
+	// NovaCellsDeletionMessage
+	NovaCellsDeletionMessage = "NovaCells deletion in progress"
+
+	// NovaCellsDeletionConditionReadyMessage
+	NovaCellsDeletionConditionReadyMessage = "There is no more NovaCells to delete"
```
gibizer, line 185: +1 for making a good ready message that

api/v1beta1/conditions.go
```diff
@@ -172,4 +174,13 @@ const (
 
 	//CellHostDiscoverErrorMessage
 	CellHostDiscoverErrorMessage = "CellHostDiscover error occurred %s"
+
+	// NovaCellsDeletionConditionInitMessage
+	NovaCellsDeletionConditionInitMessage = "NovaCells deletion not started"
+
+	// NovaCellsDeletionMessage
+	NovaCellsDeletionMessage = "NovaCells deletion in progress"
```
gibizer, line 182: We probably need a bit more information in the condition message.
* there are cells that failed to delete. They will not succeed without external intervention.
* there are cells where the deletion is progressing and we just waiting for the job to finish.

controllers/nova_controller.go
```diff
@@ -604,10 +604,18 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	toDeletCells := map[string]string{}
 
 	for _, cr := range novaCellList.Items {
 		_, ok := instance.Spec.CellTemplates[cr.Spec.CellName]
 		if !ok {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				novav1.NovaCellsDeletionCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				novav1.NovaCellsDeletionMessage,
```
gibizer, line 616: I would like to include the list of cell names:
* where deletion failed.
* where the deletion is in progress as we are waiting for the job to finish

I.e. a bit similar to how we handle the status during creation of the cells. E.g. https://github.com/openstack-k8s-operators/nova-operator/blob/99e0f2e04f20ecaa128a4877ce076d762a8cc7d1/controllers/nova_controller.go#L340-L357

gibizer (changes requested):

test/functional/nova_reconfiguration_test.go
```diff
@@ -191,6 +191,12 @@ var _ = Describe("Nova reconfiguration", func() {
 				g.Expect(nova.Status.RegisteredCells).To(HaveKey(cell1.CellCRName.Name))
 			}, timeout, interval).Should(Succeed())
 
+			th.ExpectCondition(
+				novaNames.NovaName,
+				ConditionGetterFunc(NovaConditionGetter),
+				novav1.NovaCellsDeletionCondition,
+				corev1.ConditionFalse,
```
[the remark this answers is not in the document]
mrkisaolamb: done, also I updated test with additional condition checks and update second test with cell deletion to check behaviour with failing  deletion

controllers/nova_controller.go
```diff
@@ -617,15 +620,33 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+				delete(toDeletCells, cr.Spec.CellName)
 			}
 		}
+		if len(toDeletCells) > 0 {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				novav1.NovaCellsDeletionCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				novav1.NovaCellsDeletionMessage,
+				toDeletCells,
```
gibizer, line 632: convert this to a list please as the condition message is end user facing. So we should have a nice readable list instead of a confusing map[:] construct in the message.

gibizer (changes requested):

controllers/nova_controller.go
mrkisaolamb: thanks! done

gibizer (approved):
Looks good. Thanks

mrkisaolamb:
recheck

PR #956, opened 2025-04-10
Do not update status when reconciler exits due to panic

ratailor:
As of now, we use deffered call to always update the status
field when Reconcile call exits, which seems correct for
any other error except Reconciler exits due to panic to avoid
persisting incorrect data.

This change adds a call to recover function and try to handle
panic and log message about the same.

NOTE: Testing steps:
- check CR status (eg. oc get novaapi -o yaml -w)
- inject panic in novaapi controller code like below.
  (eg. before ensureDeployment func call):
  if *instance.Spec.Replicas == int32(2) {
    	panic("panic replicas set to 2 ")
    }
- start controller change in webhook using:
	SKIP_FIREWALL=true make run-with-webhook
- change novaapi CR replica to 2
  (using oc edit openstackcontrolplane)
- check that CR status is not changing.

Closes: [OSPRH-9734](https://issues.redhat.com//browse/OSPRH-9734)
Closes: [OSPRH-12981](https://issues.redhat.com//browse/OSPRH-12981)

### DIFF
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -150,6 +150,11 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happened during the current reconciliation.
 	defer func() {
+		// Don't update the status, if Reconciler Panics
+		if r := recover(); r != nil {
+			Log.Info(fmt.Sprintf("Panic during reconcile %v\n", r))
+			panic(r)
+		}
 		// update the Ready condition based on the sub conditions
 		if allSubConditionIsTrue(instance.Status) {
 			instance.Status.Conditions.MarkTrue(
--- a/controllers/novaapi_controller.go
+++ b/controllers/novaapi_controller.go
@@ -136,6 +136,11 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happened during the current reconciliation.
 	defer func() {
+		// Don't update the status, if Reconciler Panics
+		if r := recover(); r != nil {
+			Log.Info(fmt.Sprintf("Panic during reconcile %v\n", r))
+			panic(r)
+		}
 		// update the Ready condition based on the sub conditions
 		if allSubConditionIsTrue(instance.Status) {
 			instance.Status.Conditions.MarkTrue(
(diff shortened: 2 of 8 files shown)


gibizer (commented):
Testing this any automated env is probably not possible without active instrumentation (i.e. intentionally breaking the code). But I would like to ask you to do a manual verification of the behavior instead. Inject a panic to a specific place in the middle of the reconciler that you can trigger from outside (e.g. add a panic if one of the replicas is set to 2). Then create a build and deploy it locally and trigger the panic in your env and observer that the panic is handled as expected and the CR status is not updated.

ratailor:
DNM this now. manual testing  as suggested by @gibizer is remaining.

mrkisaolamb (approved):
lgtm

controllers/nova_controller.go
```diff
@@ -148,6 +148,11 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happened during the current reconciliation.
 	defer func() {
+		// Don't update the status, if Reconciler Panics
+		if r := recover(); r != nil {
+			Log.Error(err, "Panic during reconcile: %v\n", r)
```
gibizer, line 153: the `err` here is still from the outer scope and hence unrelated to the problem we are logging here.

gibizer (changes requested):

auniyal61:
Change looks  good to me.

can you please share the steps, on how to test to reproduce panic.

We do not have any release notes or docs to update with patch in operator side, so may be we can add that in commit msg.

controllers/nova_controller.go
ratailor: Ack. I have addressed this in updated version of PR.

controllers/nova_controller.go
```diff
@@ -149,6 +149,11 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	// Always update the instance status when exiting this function so we can
 	// persist any changes happened during the current reconciliation.
 	defer func() {
+		// Don't update the status, if Reconciler Panics
+		if r := recover(); r != nil {
+			Log.Error(err, "Panic during reconcile: %s", r)
+			return
```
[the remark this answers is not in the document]
ratailor: Done.

gibizer (approved):
Looks good. Thanks

abays:
This looks like something that we should replicate across all our operators, as we use the same `defer()` pattern in all of them.

gibizer:
> This looks like something that we should replicate across all our operators, as we use the same `defer()` pattern in all of them.

Yepp. I agree. I've already pinged @ratailor and the team on #rhos-nova about it.

PR #958, opened 2025-04-16
Adds notification bus 

auniyal61:
Initial commit - Adds tests for notification bus

how this should work is:

1 - human operator will update rabbitmq in openstackcontrolplane CR
     a) under rabbitmq: add new rabbitmq, ex: rabbitmq-broadcaster, something like
```
           rabbitmq-broadcaster:
               replicas: 1
```

2 - human operator will update Nova in  openstackcontrolplane CR
     a) under nova.template.spec:  register new rabbit 
                    (as we do for new cell but not inside celltemplate, as this one is top-level)
                    
```
     nova:
	  template:
               spec:
                  apiContainerImageURL: image_url
                  apiMessageBusInstance: rabbitmq
                  // new var
                  notificationsBusInstance: rabbitmq-broadcaster    
```

3 - now nova-operator will recognize notificationsBusInstance and retrieve transport_url of new rabbit (rabbitmq-broadcaster)

4 - nova-operator will set this transport_url in nova.conf

Jira: [OSPRH-15392](https://issues.redhat.com//browse/OSPRH-15392)

### DIFF
--- a/api/v1beta1/conditions.go
+++ b/api/v1beta1/conditions.go
@@ -54,6 +54,10 @@ const (
 	NovaAllControlPlaneComputesReadyCondition condition.Type = "NovaAllControlPlaneComputesReady"
 	//NovaCellsDeletionCondition indicates that the NovaCells deletion is in progress
 	NovaCellsDeletionCondition condition.Type = "NovaCellsDeletion"
+
+	// notifications
+	// NovaNotificationMQReadyCondition indicated that the top level notification message bus is ready
+	NovaNotificationMQReadyCondition condition.Type = "NovaNotificationMQReady"
 )
 
 // Common Messages used by API objects.
@@ -183,4 +187,16 @@ const (
 
 	// NovaCellsDeletionConditionReadyMessage
 	NovaCellsDeletionConditionReadyMessage = "There is no more NovaCells to delete"
+	// notifications
+	// NovaNotificationMQReadyInitMessage
+	NovaNotificationMQReadyInitMessage = "Notification message bus not started"
+
+	// NovaNotificationMQReadyErrorMessage
+	NovaNotificationMQReadyErrorMessage = "Notification message bus creation failed: %s"
+
+	// NovaNotificationMQReadyCreatingMessage
+	NovaNotificationMQReadyCreatingMessage = "Notification message bus creation ongoing"
+
+	// NovaNotificationMQReadyMessage
+	NovaNotificationMQReadyMessage = "Notification message bus created successfully"
 )
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -96,6 +96,10 @@ const (
 	// Secret for the cell message bus transport URL
 	TransportURLSelector = "transport_url"
 
+	// NotificationTransportURLSelector is the name of
+	// top level notification message bus transport URL
+	NotificationTransportURLSelector = "notification_transport_url"
+
 	// fields to index to reconcile when change
 	passwordSecretField        = ".spec.secret"
 	caBundleSecretNameField    = ".spec.tls.caBundleSecretName" // #nosec G101
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -391,6 +391,47 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, fmt.Errorf("%w from  for the API MQ: %d", util.ErrInvalidStatus, apiMQStatus)
 	}
 
+	// nova broadcaster rabbit
+	notificationBusName := ""
+	if instance.Spec.NotificationsBusInstance != nil {
+		notificationBusName = *instance.Spec.NotificationsBusInstance
+	}
+
+	var notificationTransportURL string
+	var notificationMQStatus nova.MessageBusStatus
+	var notificationMQError error
+
+	if notificationBusName != "" {
+		notificationTransportURL, notificationMQStatus, notificationMQError = r.ensureMQ(
+			ctx, h, instance, instance.Name+"-notification-transport", notificationBusName)
+
+		switch notificationMQStatus {
+		case nova.MQFailed:
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				novav1.NovaNotificationMQReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityError,
+				novav1.NovaNotificationMQReadyErrorMessage,
+				notificationMQError.Error(),
+			))
+		case nova.MQCreating:
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				novav1.NovaNotificationMQReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityError,
+				novav1.NovaNotificationMQReadyCreatingMessage,
+			))
+		case nova.MQCompleted:
+			instance.Status.Conditions.MarkTrue(
+				novav1.NovaNotificationMQReadyCondition, novav1.NovaNotificationMQReadyMessage)
+		default:
+			return ctrl.Result{}, fmt.Errorf("%w from  for the Notification MQ: %d",
+				util.ErrInvalidStatus, notificationMQStatus)
+		}
+	} else {
+		instance.Status.Conditions.Remove(novav1.NovaNotificationMQReadyCondition)
+	}
+
 	cellMQs := map[string]*nova.MessageBus{}
 	var failedMQs []string
 	var creatingMQs []string
@@ -481,7 +522,7 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		}
 		cell, status, err := r.ensureCell(
 			ctx, h, instance, cellName, cellTemplate,
-			cellDB.Database, apiDB, cellMQ.TransportURL,
+			cellDB.Database, apiDB, cellMQ.TransportURL, notificationTransportURL,
 			keystoneInternalAuthURL, secret,
 		)
 		cells[cellName] = cell
@@ -546,7 +587,11 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, nil
 	}
 
-	topLevelSecretName, err := r.ensureTopLevelSecret(ctx, h, instance, apiTransportURL, secret)
+	topLevelSecretName, err := r.ensureTopLevelSecret(
+		ctx, h, instance,
+		apiTransportURL,
+		notificationTransportURL,
+		secret)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -936,6 +981,11 @@ func (r *NovaReconciler) initConditions(
 			condition.InitReason,
 			condition.MemcachedReadyInitMessage,
 		),
+		condition.UnknownCondition(
+			novav1.NovaNotificationMQReadyCondition,
+			condition.InitReason,
+			novav1.NovaNotificationMQReadyInitMessage,
+		),
 	)
 	instance.Status.Conditions.Init(&cl)
 	return nil
@@ -1133,12 +1183,16 @@ func (r *NovaReconciler) ensureCell(
 	cellDB *mariadbv1.Database,
 	apiDB *mariadbv1.Database,
 	cellTransportURL string,
+	notificationTransportURL string,
 	keystoneAuthURL string,
 	secret corev1.Secret,
 ) (*novav1.NovaCell, nova.CellDeploymentStatus, error) {
 	Log := r.GetLogger(ctx)
 
-	cellSecretName, err := r.ensureCellSecret(ctx, h, instance, cellName, cellTemplate, cellTransportURL, secret)
+	cellSecretName, err := r.ensureCellSecret(
+		ctx, h, instance, cellName, cellTemplate,
+		cellTransportURL, notificationTransportURL,
+		secret)
 	if err != nil {
 		return nil, nova.CellDeploying, err
 	}
@@ -1684,6 +1738,7 @@ func (r *NovaReconciler) ensureMQ(
 	}
 
 	secretName := types.NamespacedName{Namespace: instance.Namespace, Name: transportURL.Status.SecretName}
+
 	secret := &corev1.Secret{}
 	err = h.GetClient().Get(ctx, secretName, secret)
 	if err != nil {
@@ -1698,7 +1753,6 @@ func (r *NovaReconciler) ensureMQ(
 		return "", nova.MQFailed, fmt.Errorf(
 			"%w: the TransportURL secret %s does not have 'transport_url' field", util.ErrFieldNotFound, transportURL.Status.SecretName)
 	}
-
 	return string(url), nova.MQCompleted, nil
 }
 
@@ -1902,13 +1956,15 @@ func (r *NovaReconciler) ensureCellSecret(
 	cellName string,
 	cellTemplate novav1.NovaCellTemplate,
 	cellTransportURL string,
+	notificationTransportURL string,
 	externalSecret corev1.Secret,
 ) (string, error) {
 	// NOTE(gibi): We can move other sensitive data to the internal Secret from
 	// the NovaCellSpec fields, possibly hostnames or usernames.
 	data := map[string]string{
-		ServicePasswordSelector: string(externalSecret.Data[instance.Spec.PasswordSelectors.Service]),
-		TransportURLSelector:    cellTransportURL,
+		ServicePasswordSelector:          string(externalSecret.Data[instance.Spec.PasswordSelectors.Service]),
+		TransportURLSelector:             cellTransportURL,
+		NotificationTransportURLSelector: notificationTransportURL,
 	}
 
 	// If metadata is enabled in the cell then the cell secret needs the
@@ -1952,14 +2008,16 @@ func (r *NovaReconciler) ensureTopLevelSecret(
 	h *helper.Helper,
 	instance *novav1.Nova,
 	apiTransportURL string,
+	notificationTransportURL string,
 	externalSecret corev1.Secret,
 ) (string, error) {
 	// NOTE(gibi): We can move other sensitive data to the internal Secret from
 	// the subCR fields, possibly hostnames or usernames.
 	data := map[string]string{
-		ServicePasswordSelector: string(externalSecret.Data[instance.Spec.PasswordSelectors.Service]),
-		MetadataSecretSelector:  string(externalSecret.Data[instance.Spec.PasswordSelectors.MetadataSecret]),
-		TransportURLSelector:    apiTransportURL,
+		ServicePasswordSelector:          string(externalSecret.Data[instance.Spec.PasswordSelectors.Service]),
+		MetadataSecretSelector:           string(externalSecret.Data[instance.Spec.PasswordSelectors.MetadataSecret]),
+		TransportURLSelector:             apiTransportURL,
+		NotificationTransportURLSelector: notificationTransportURL,
 	}
 
 	// NOTE(gibi): When we switch to immutable secrets then we need to include
--- a/controllers/novaapi_controller.go
+++ b/controllers/novaapi_controller.go
@@ -183,16 +183,19 @@ func (r *NovaAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	// detect if something is changed.
 	hashes := make(map[string]env.Setter)
 
-	secretHash, result, secret, err := ensureSecret(
-		ctx,
-		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+	requiredSecretFields := []string{
 		// TODO(gibi): add keystoneAuthURL here is that is also passed via
 		// the Secret. Also add DB and MQ user name here too if those are
 		// passed via the Secret
-		[]string{
-			ServicePasswordSelector,
-			TransportURLSelector,
-		},
+		ServicePasswordSelector,
+		TransportURLSelector,
+		NotificationTransportURLSelector,
+	}
+
+	secretHash, result, secret, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		requiredSecretFields,
 		h.GetClient(),
 		&instance.Status.Conditions,
 		r.RequeueTimeout,
@@ -473,28 +476,29 @@ func (r *NovaAPIReconciler) generateConfigs(
 		"keystone_internal_url": instance.Spec.KeystoneAuthURL,
 		// NOTE(gibi): As per the definition of www_authenticate_uri this
 		// always needs to point to the public keystone endpoint.
-		"www_authenticate_uri":     instance.Spec.KeystonePublicAuthURL,
-		"nova_keystone_user":       instance.Spec.ServiceUser,
-		"nova_keystone_password":   string(secret.Data[ServicePasswordSelector]),
-		"api_db_name":              NovaAPIDatabaseName,
-		"api_db_user":              apiDatabaseAccount.Spec.UserName,
-		"api_db_password":          string(apiDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
-		"api_db_address":           instance.Spec.APIDatabaseHostname,
-		"api_db_port":              3306,
-		"cell_db_name":             NovaCell0DatabaseName,
-		"cell_db_user":             cellDatabaseAccount.Spec.UserName,
-		"cell_db_password":         string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
-		"cell_db_address":          instance.Spec.Cell0DatabaseHostname,
-		"cell_db_port":             3306,
-		"openstack_region_name":    "regionOne", // fixme
-		"default_project_domain":   "Default",   // fixme
-		"default_user_domain":      "Default",   // fixme
-		"transport_url":            string(secret.Data[TransportURLSelector]),
-		"log_file":                 "/var/log/nova/nova-api.log",
-		"tls":                      false,
-		"MemcachedServers":         memcachedInstance.GetMemcachedServerListString(),
-		"MemcachedServersWithInet": memcachedInstance.GetMemcachedServerListWithInetString(),
-		"MemcachedTLS":             memcachedInstance.GetMemcachedTLSSupport(),
+		"www_authenticate_uri":       instance.Spec.KeystonePublicAuthURL,
+		"nova_keystone_user":         instance.Spec.ServiceUser,
+		"nova_keystone_password":     string(secret.Data[ServicePasswordSelector]),
+		"api_db_name":                NovaAPIDatabaseName,
+		"api_db_user":                apiDatabaseAccount.Spec.UserName,
+		"api_db_password":            string(apiDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
+		"api_db_address":             instance.Spec.APIDatabaseHostname,
+		"api_db_port":                3306,
+		"cell_db_name":               NovaCell0DatabaseName,
+		"cell_db_user":               cellDatabaseAccount.Spec.UserName,
+		"cell_db_password":           string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
+		"cell_db_address":            instance.Spec.Cell0DatabaseHostname,
+		"cell_db_port":               3306,
+		"openstack_region_name":      "regionOne", // fixme
+		"default_project_domain":     "Default",   // fixme
+		"default_user_domain":        "Default",   // fixme
+		"transport_url":              string(secret.Data[TransportURLSelector]),
+		"notification_transport_url": string(secret.Data[NotificationTransportURLSelector]),
+		"log_file":                   "/var/log/nova/nova-api.log",
+		"tls":                        false,
+		"MemcachedServers":           memcachedInstance.GetMemcachedServerListString(),
+		"MemcachedServersWithInet":   memcachedInstance.GetMemcachedServerListWithInetString(),
+		"MemcachedTLS":               memcachedInstance.GetMemcachedTLSSupport(),
 	}
 	// create httpd  vhost template parameters
 	httpdVhostConfig := map[string]interface{}{}
--- a/controllers/novacell_controller.go
+++ b/controllers/novacell_controller.go
@@ -144,14 +144,17 @@ func (r *NovaCellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		}
 	}()
 
+	requiredSecretFields := []string{
+		ServicePasswordSelector,
+		TransportURLSelector,
+		NotificationTransportURLSelector,
+	}
+
 	// For the compute config generation we need to read the input secrets
 	_, result, secret, err := ensureSecret(
 		ctx,
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
-		[]string{
-			ServicePasswordSelector,
-			TransportURLSelector,
-		},
+		requiredSecretFields,
 		h.GetClient(),
 		&instance.Status.Conditions,
 		r.RequeueTimeout,
@@ -767,15 +770,16 @@ func (r *NovaCellReconciler) generateComputeConfigs(
 	secret corev1.Secret, vncProxyURL *string,
 ) error {
 	templateParameters := map[string]interface{}{
-		"service_name":           "nova-compute",
-		"keystone_internal_url":  instance.Spec.KeystoneAuthURL,
-		"nova_keystone_user":     instance.Spec.ServiceUser,
-		"nova_keystone_password": string(secret.Data[ServicePasswordSelector]),
-		"openstack_region_name":  "regionOne", // fixme
-		"default_project_domain": "Default",   // fixme
-		"default_user_domain":    "Default",   // fixme
-		"compute_driver":         "libvirt.LibvirtDriver",
-		"transport_url":          string(secret.Data[TransportURLSelector]),
+		"service_name":               "nova-compute",
+		"keystone_internal_url":      instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":         instance.Spec.ServiceUser,
+		"nova_keystone_password":     string(secret.Data[ServicePasswordSelector]),
+		"openstack_region_name":      "regionOne", // fixme
+		"default_project_domain":     "Default",   // fixme
+		"default_user_domain":        "Default",   // fixme
+		"compute_driver":             "libvirt.LibvirtDriver",
+		"transport_url":              string(secret.Data[TransportURLSelector]),
+		"notification_transport_url": string(secret.Data[NotificationTransportURLSelector]),
 	}
 	// vnc is optional so we only need to configure it for the compute
 	// if the proxy service is deployed in the cell
--- a/controllers/novaconductor_controller.go
+++ b/controllers/novaconductor_controller.go
@@ -182,6 +182,7 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	requiredSecretFields := []string{
 		ServicePasswordSelector,
 		TransportURLSelector,
+		NotificationTransportURLSelector,
 	}
 
 	secretHash, result, secret, err := ensureSecret(
@@ -431,22 +432,23 @@ func (r *NovaConductorReconciler) generateConfigs(
 	cellDbSecret := cellDB.GetSecret()
 
 	templateParameters := map[string]interface{}{
-		"service_name":             "nova-conductor",
-		"keystone_internal_url":    instance.Spec.KeystoneAuthURL,
-		"nova_keystone_user":       instance.Spec.ServiceUser,
-		"nova_keystone_password":   string(secret.Data[ServicePasswordSelector]),
-		"cell_db_name":             getCellDatabaseName(instance.Spec.CellName),
-		"cell_db_user":             cellDatabaseAccount.Spec.UserName,
-		"cell_db_password":         string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
-		"cell_db_address":          instance.Spec.CellDatabaseHostname,
-		"cell_db_port":             3306,
-		"openstack_region_name":    "regionOne", // fixme
-		"default_project_domain":   "Default",   // fixme
-		"default_user_domain":      "Default",   // fixme
-		"transport_url":            string(secret.Data[TransportURLSelector]),
-		"MemcachedServers":         memcachedInstance.GetMemcachedServerListString(),
-		"MemcachedServersWithInet": memcachedInstance.GetMemcachedServerListWithInetString(),
-		"MemcachedTLS":             memcachedInstance.GetMemcachedTLSSupport(),
+		"service_name":               "nova-conductor",
+		"keystone_internal_url":      instance.Spec.KeystoneAuthURL,
+		"nova_keystone_user":         instance.Spec.ServiceUser,
+		"nova_keystone_password":     string(secret.Data[ServicePasswordSelector]),
+		"cell_db_name":               getCellDatabaseName(instance.Spec.CellName),
+		"cell_db_user":               cellDatabaseAccount.Spec.UserName,
+		"cell_db_password":           string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
+		"cell_db_address":            instance.Spec.CellDatabaseHostname,
+		"cell_db_port":               3306,
+		"openstack_region_name":      "regionOne", // fixme
+		"default_project_domain":     "Default",   // fixme
+		"default_user_domain":        "Default",   // fixme
+		"transport_url":              string(secret.Data[TransportURLSelector]),
+		"notification_transport_url": string(secret.Data[NotificationTransportURLSelector]),
+		"MemcachedServers":           memcachedInstance.GetMemcachedServerListString(),
+		"MemcachedServersWithInet":   memcachedInstance.GetMemcachedServerListWithInetString(),
+		"MemcachedTLS":               memcachedInstance.GetMemcachedTLSSupport(),
 	}
 	if len(instance.Spec.APIDatabaseHostname) > 0 {
 		apiDatabaseAccount, apiDbSecret, err := mariadbv1.GetAccountAndSecret(ctx, h, instance.Spec.APIDatabaseAccount, instance.Namespace)
--- a/controllers/novascheduler_controller.go
+++ b/controllers/novascheduler_controller.go
@@ -177,16 +177,19 @@ func (r *NovaSchedulerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// detect if something is changed.
 	hashes := make(map[string]env.Setter)
 
-	secretHash, result, secret, err := ensureSecret(
-		ctx,
-		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+	requiredSecretFields := []string{
 		// TODO(gibi): add keystoneAuthURL here is that is also passed via
 		// the Secret. Also add DB and MQ user name here too if those are
 		// passed via the Secret
-		[]string{
-			ServicePasswordSelector,
-			TransportURLSelector,
-		},
+		ServicePasswordSelector,
+		TransportURLSelector,
+		NotificationTransportURLSelector,
+	}
+
+	secretHash, result, secret, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		requiredSecretFields,
 		h.GetClient(),
 		&instance.Status.Conditions,
 		r.RequeueTimeout,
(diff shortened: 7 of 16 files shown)


bogdando:
@auniyal61 Please make sure the tests covering expected behavir should be aligned with some changes done to the design spec https://github.com/openstack-k8s-operators/nova-operator/pull/948

test/functional/base_test.go
```diff
@@ -519,6 +599,7 @@ type NovaNames struct {
 	MemcachedNamespace              types.NamespacedName
 	Cells                           map[string]CellNames
 	NovaTopologies                  []types.NamespacedName
+	notificationsBusName            types.NamespacedName
```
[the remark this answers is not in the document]
gibizer: OK so the cell0 messageBus and the apiMessageBus are the same in a normal deployment so this is why we don't have a test data for the apiMessageBus specifically but we use the cell0 test data for it. 

Here, for the notification bus we know that can be (recommended to be) different bus. So we need test data (a name) for it here. 

I think a single NotificationBusName type.NamespacedName field is enough and we don't need NotificationsBusNames struct at all.

test/functional/base_test.go
```diff
@@ -485,6 +524,47 @@ func GetCellNames(novaName types.NamespacedName, cell string) CellNames {
 	return c
 }
 
+type NotificationsBusNames struct {
+	notificationsBusName   string
+	notificationsBusCRName types.NamespacedName
+	TransportURLName       types.NamespacedName
```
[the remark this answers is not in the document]
gibizer: lets drop the whole struct and only use the single field you added in NovaNames

test/functional/base_test.go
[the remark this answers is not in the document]
SeanMooney: well cell0 is special in that it holds the service records for all top level service and 
it genreall ydoes into make sense to use a different notification message bus for cell0 vs the top level services because you cant have compute services in cell0 as a result you will not have celel level metadata or vnc services at cell 0 either.

so only the real compute cells (cell1+) would ever have a usecase to use a different bus form the top level global services.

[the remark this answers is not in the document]
bogdando: note that implementation shold not focus on different busses for individual cells 
pleaes remove that from test coverage, the focus for today is only using single bus instance for notifications

controllers/nova_controller.go
```diff
@@ -233,6 +233,8 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		mariadbv1.MariaDBAccountReadyMessage,
 	)
 
+	Log.Info("", ":", "t-url-instance", ":", instance.Spec)
```
auniyal61, line 236: this should have, which is not right now
`notificationsBusInstance: rabbitmq-broadcaster `

controllers/nova_controller.go
```diff
@@ -1647,13 +1696,16 @@ func (r *NovaReconciler) ensureMQ(
 	}
 
 	if k8s_errors.IsNotFound(err) || !transportURL.IsReady() || transportURL.Status.SecretName == "" {
+		Log.Info("", ":", "t-url-4", "", transportURL)
+		Log.Info("", ":", "t-url-4", "", transportURL.Status)
```
auniyal61, line 1700: control stucks here.

controllers/nova_controller.go
auniyal61: which is because nova reconciler in oscp is removing it, as its not aware of it.
https://github.com/openstack-k8s-operators/openstack-operator/blob/b0944fc30a9387cf41d019165c5dbc6a8f128597/pkg/openstack/nova.go#L44

so I need to do either
1- build oscp with commit design crd [commit](https://github.com/openstack-k8s-operators/nova-operator/pull/958/commits/c14c644af0ad1368a8e16e672940150bb80d5ed1)

2 - add all those changes manually in local mod ex: nova_types 
~/go/pkg/mod/github.com/openstack-k8s-operators/nova-operator/api\@v0.6.1-0.20250411170009-dcff8bfb312f/v1beta1/nova_types.go

trying 2 since  sometime!

auniyal61: its not working because, `config/crd/bases/nova.openstack.org_nova.yaml` not found in local build ?

auniyal61: oscp change

> 	tempName := "rabbitmq-brdc"
	if instance.Spec.Nova.Template.NotificationsBusInstance == nil {
		instance.Spec.Nova.Template.NotificationsBusInstance = &tempName
	}
this loged in oscp logs, but do not reflect at nova-side.

api/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1624,15 @@ spec:
                   NodeSelector here acts as a default value and can be overridden by service
                   specific NodeSelector Settings.
                 type: object
+              notificationsBusInstance:
+                description: |-
+                  NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+                  the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+                  If undefined, the value will be inherited from OpenStackControlPlane.
+                  An empty value "" leaves the notification drivers unconfigured and emitting no notifications at all.
+                  Avoid colocating it with APIMessageBusInstance or CellMessageBusInstance used for PRC.
```
dprince, line 1633: This should also call out RabbitMqClusterName as it appears every service aside from nova-operator uses this naming convention instead.

controllers/nova_controller.go
bogdando: I drafted the oscp operator side of things https://github.com/openstack-k8s-operators/openstack-operator/pull/1424, if that helps

api/bases/nova.openstack.org_nova.yaml
bogdando: I have fixed this in the base PR https://github.com/openstack-k8s-operators/nova-operator/pull/948

controllers/nova_controller.go
auniyal61: yeah, its reflecting now for me.
I need to `make generate manifests install`

templates/nova.conf
```diff
@@ -127,7 +127,7 @@ api_paste_config = /etc/nova/api-paste.ini
 
 [oslo_messaging_notifications]
 {{ if (index . "nova_enabled_notification") }}
```
SeanMooney, line 129: change this to 
{{ if (index . "notification_transport_url") }}

controllers/novaapi_controller.go
```diff
@@ -451,45 +451,49 @@ func (r *NovaAPIReconciler) generateConfigs(
...
+		"cell_db_name":               NovaCell0DatabaseName,
+		"cell_db_user":               cellDatabaseAccount.Spec.UserName,
+		"cell_db_password":           string(cellDbSecret.Data[mariadbv1.DatabasePasswordSelector]),
+		"cell_db_address":            instance.Spec.Cell0DatabaseHostname,
+		"cell_db_port":               3306,
+		"openstack_region_name":      "regionOne", // fixme
+		"default_project_domain":     "Default",   // fixme
+		"default_user_domain":        "Default",   // fixme
+		"transport_url":              string(secret.Data[TransportURLSelector]),
+		"notification_transport_url": string(secret.Data[NotificationTransportURLSelector]),
```
SeanMooney, line 491: alternitively defien 
```
nova_enabled_notification: string(secret.Data[TransportURLSelector]) != "", 
```

controllers/nova_controller.go
```diff
@@ -384,6 +384,39 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, fmt.Errorf("%w from  for the API MQ: %d", util.ErrInvalidStatus, apiMQStatus)
 	}
 
+	// nova broadcaster rabbit
+	notificationBusName := ""
+	if instance.Spec.NotificationsBusInstance != nil {
+		notificationBusName = *instance.Spec.NotificationsBusInstance
+	}
```
gibizer, line 391: We know that in reality openstack-operator will never live this to nil, but I agree that technically it is a pointer so we should handle nil. Also I agree that setting the bus name to `""` if the field is `nil` make sense.

controllers/nova_controller.go
```diff
@@ -384,6 +384,39 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+
+	notificationTransportURL, notificationMQStatus, notificationMQError := r.ensureMQ(
+		ctx, h, instance, instance.Name+"-notification-transport", notificationBusName)
```
gibizer, line 394: What will happen here when notificationBusName is `""`?

controllers/nova_controller.go
auniyal61: ""          --> notification is disabled
nil         --> its not available in Nova CR, so disable it by setting ""
"foo"    --> send notification to rabbitmq foo

controllers/nova_controller.go
auniyal61: done

controllers/novascheduler_controller.go
```diff
@@ -172,16 +172,18 @@ func (r *NovaSchedulerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	secretHash, result, secret, err := ensureSecret(
 		ctx,
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
 		// TODO(gibi): add keystoneAuthURL here is that is also passed via
 		// the Secret. Also add DB and MQ user name here too if those are
 		// passed via the Secret
-		[]string{
-			ServicePasswordSelector,
-			TransportURLSelector,
-		},
```
auniyal61, line 184: why only these 2,

controllers/novaconductor_controller.go
```diff
@@ -178,6 +178,7 @@ func (r *NovaConductorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	requiredSecretFields := []string{
 		ServicePasswordSelector,
 		TransportURLSelector,
+		// NotificationTransportURLSelector,
```
auniyal61, line 181: why only these 2 ?
in scheduler without asking for new `notification var`, its retrieving, but not here in conductor, 
so I tried passing (the commented one) that but it failed, because it was not present in secret.

auniyal61: can instance.generation be different for both conductor and scheduler reconciler

for scheduler its 7 
for condutor its 1, (scheduler works)  
---

okay, so the reason is, in scheduler 
secret: nova is passed 

while in conductor
secret: nova-cell0
secret: nova-cell1
are passed

auniyal61: works for conductor too now

controllers/novascheduler_controller.go
gibizer: Wee need to list those field selectors the reconciler uses later. So If the reconciler uses the new notififiaction selector then we have to check that here too (and in the other service reconcilers too)

controllers/common.go
```diff
@@ -94,7 +94,8 @@ const (
 	MetadataSecretSelector = "MetadataSecret"
 	// TransportURLSelector is the name of key in the internal cell
 	// Secret for the cell message bus transport URL
-	TransportURLSelector = "transport_url"
+	TransportURLSelector             = "transport_url"
+	NotificationTransportURLSelector = "notificaton_transport_url"
```
mrkisaolamb, line 98: can we add similarr comment eg. NotificationTransportURLSelector is the name of key in the internal cell Secret for the global notification message bus transport URL

controllers/novaapi_controller.go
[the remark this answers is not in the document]
auniyal61: done

api/v1beta1/conditions.go
```diff
@@ -52,6 +52,10 @@ const (
 	NovaComputeServiceConfigReady condition.Type = "NovaComputeServiceConfigReady"
 	// NovaAllControlPlaneComputesReadyCondition indicates that every defined NovaCompute is ready but undiscovered
 	NovaAllControlPlaneComputesReadyCondition condition.Type = "NovaAllControlPlaneComputesReady"
+
+	// notifications
+	// NovaNotificationMQReadyCondition indicated that the top level notification message bus is ready
+	NovaNotificationMQReadyCondition condition.Type = "NovaNotificationMQReady"
```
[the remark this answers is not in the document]
auniyal61: added

controllers/nova_controller.go
```diff
@@ -384,6 +384,45 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+				novav1.NovaNotificationMQReadyCreatingMessage,
+			))
+		case nova.MQCompleted:
+			instance.Status.Conditions.MarkTrue(
+				novav1.NovaNotificationMQReadyCondition, novav1.NovaNotificationMQReadyMessage)
+		default:
+			return ctrl.Result{}, fmt.Errorf("%w from  for the Notification MQ: %d",
+				util.ErrInvalidStatus, notificationMQStatus)
+		}
+	}
```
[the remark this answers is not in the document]
auniyal61: added

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,184 @@ import (
...
+				nova.Spec.NotificationsBusInstance = &notificationsBus.TransportURLName.Name
+				g.Expect(k8sClient.Update(ctx, nova)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			nova := GetNova(novaNames.NovaName)
+			Expect(nova.Spec.NotificationsBusInstance).ToNot(BeNil())
+			Expect(*nova.Spec.NotificationsBusInstance).To(Equal("rabbitmq-broadcaster"))
+
+			// assert that a the top level internal internal secret is created
+			internalTopLevelSecret := th.GetSecret(novaNames.InternalTopLevelSecretName)
```
auniyal61, line 179: this does not retrieve the update secret

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,184 @@ import (
...
+			// logger.Info("XXX", "transport-url", internalTopLevelSecret)
+			// verify if nova secret has notification-transport-url
+			// not sure why its still empty
+			Expect(internalTopLevelSecret.Data).To(
+				HaveKeyWithValue("notificaton_transport_url", []byte("")))
+
+			// verify if nova-api.conf has notification-transport-url
+			configDataMap := th.GetSecret(novaNames.APIConfigDataName)
```
auniyal61, line 187: same, here api coinf is not updated, even nova CR has notification added.

Also nova has status
```
status:
  conditions:
    - type: NovaNotificationMQReady
      status: "True"
      lastTransitionTime: 2025-05-07T06:40:50Z
      reason: Ready
      message: Notification message bus not requested
    - type: NovaSchedulerReady
      status: "True"
      lastTransitionTime: 2025-05-07T06:40:51Z
      reason: Ready
      message: Setup complete
```

controllers/nova_controller.go
```diff
@@ -384,6 +384,49 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	if instance.Spec.NotificationsBusInstance != nil {
+		notificationBusName = *instance.Spec.NotificationsBusInstance
+	}
+
+	var notificationTransportURL string
+	var notificationMQStatus nova.MessageBusStatus
+	var notificationMQError error
+
+	if notificationBusName != "" {
+		notificationTransportURL, notificationMQStatus, notificationMQError = r.ensureMQ(
```
gibizer, line 398: it is a convention to call the error variable as `err`

test/functional/nova_controller_test.go
gibizer: What is the status of the new NovaNotificationMQReadyCondition here?

auniyal61: thanks @gibizer, works now

test/functional/suite_test.go
```diff
@@ -70,21 +70,22 @@ import (
...
+	th               *common_test.TestHelper
+	keystone         *keystone_test.TestHelper
+	mariadb          *mariadb_test.TestHelper
+	infra            *infra_test.TestHelper
+	novaNames        NovaNames
+	cell0            CellNames
+	cell1            CellNames
+	cell2            CellNames
+	cell3            CellNames
+	notificationsBus NotificationsBusNames
```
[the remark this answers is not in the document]
auniyal61: done, removed

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,147 @@ import (
...
+			CreateNotificiationTransportURLSecret(notificationsBus)
+
+			Eventually(func(g Gomega) {
+				nova := GetNova(novaNames.NovaName)
+				nova.Spec.NotificationsBusInstance = &notificationsBus.TransportURLName.Name
+				g.Expect(k8sClient.Update(ctx, nova)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			infra.SimulateTransportURLReady(notificationsBus.TransportURLName)
+
```
gibizer, line 131: Here you can assert that the notificationsBusName is set into TranportULR.Spec.RabbitmqClusterName by getting the TransportURL CR via 		transport := infra.GetTransportURL(notificationsBus.TransportURLName).

gibizer (commented):
I haven't finished reviewing the test case but I have to switch to other things so I publishing my current comments.

auniyal61:
/test functional
unrelated test `updates lastAppliedTopology in NovaNoVNCProxy .Status` failed with string not mached

openstack.org/novanovncproxy-2f4ced32-4332-4679-8953-0-cell1-novncproxy !=
openstack.org/novanovncproxy-2f4ced32-4332-4679-8953-0-cell1-novncproxy

not failing locally.

test/functional/nova_controller_test.go
auniyal61: `SimulateTransportURLReady` only works on `notificationsBus.TransportURLName` and not `notificationsBus.BusName`
```
transportURLName := infra.GetTransportURL(notificationsBus.TransportURLName)
logger.Info("XXX", "t-url", transportURLName.Spec.RabbitmqClusterName)
logger.Info("XXX", "t-url", notificationsBus.BusName)

---
  2025-05-08T06:18:22.339-0400  INFO    ---Test---      XXX     {"t-url": "54ce1a60-edcd-4355-a5f4-6-notification-transport"}
  2025-05-08T06:18:22.339-0400  INFO    ---Test---      XXX     {"t-url": "rabbitmq-broadcaster"}
```

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,147 @@ import (
 	"github.com/openstack-k8s-operators/nova-operator/controllers"
 )
 
+var _ = Describe("Nova controller - notifications broadcaster", func() {
```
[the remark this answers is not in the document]
auniyal61: done

templates/nova.conf
```diff
@@ -126,10 +126,10 @@ api_paste_config = /etc/nova/api-paste.ini
 {{end}}
 
 [oslo_messaging_notifications]
-{{ if (index . "nova_enabled_notification") }}
-transport_url =  {{ .nova_cell_notify_transport_url }}
+{{ if (index . "notification_transport_url") }}
+transport_url = {{.notification_transport_url}}
 driver = messagingv2
-notification_format=versioned
+notification_format=both
```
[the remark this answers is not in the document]
auniyal61: done

controllers/novaconductor_controller.go
```diff
@@ -426,23 +426,27 @@ func (r *NovaConductorReconciler) generateConfigs(
 	cellDatabaseAccount := cellDB.GetAccount()
 	cellDbSecret := cellDB.GetSecret()
 
+	Log := r.GetLogger(ctx)
+	Log.Info("", "t-url - conductor", secret.Data)
```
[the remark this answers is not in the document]
auniyal61: done

controllers/novascheduler_controller.go
auniyal61: done

ratailor (commented):
minor nits, otherwise lgtm.

config/crd/bases/nova.openstack.org_nova.yaml
```diff
@@ -1624,6 +1624,15 @@ spec:
                   NodeSelector here acts as a default value and can be overridden by service
                   specific NodeSelector Settings.
                 type: object
+              notificationsBusInstance:
+                description: |-
+                  NotificationsBusInstance is the name of the RabbitMqCluster CR to select
+                  the Message Bus Service instance used by the Nova top level services and all cells to publish notifications.
+                  If undefined, the value will be inherited from OpenStackControlPlane.
+                  An empty value "" leaves the notification drivers unconfigured and emitting no notifications at all.
+                  Avoid colocating it with APIMessageBusInstance or CellMessageBusInstance used for PRC.
```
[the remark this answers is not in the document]
auniyal61: done

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,150 @@ import (
...
+				g.Expect(k8sClient.Update(ctx, nova)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			infra.SimulateTransportURLReady(notificationsBus.TransportURLName)
+			transportURLName := infra.GetTransportURL(notificationsBus.TransportURLName)
+			logger.Info("XXX", "t-url", transportURLName.Spec.RabbitmqClusterName)
+			logger.Info("XXX", "t-url", notificationsBus.BusName)
+
+			nova := GetNova(novaNames.NovaName)
+			Expect(nova.Spec.NotificationsBusInstance).ToNot(BeNil())
```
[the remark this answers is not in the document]
auniyal61: ack,

test/functional/nova_controller_test.go
auniyal61: done

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,150 @@ import (
...
+			Expect(internalTopLevelSecret.Data).To(
+				HaveKeyWithValue("notificaton_transport_url", []byte("")))
+
+			assertNotHaveNotificationTransportURL := func(configData string) {
+
+				Expect(configData).To(
+					ContainSubstring("[oslo_messaging_notifications]\ndriver = noop"))
+
+				Expect(configData).ToNot(
+					ContainSubstring("[oslo_messaging_notifications]\ntransport_url = "))
```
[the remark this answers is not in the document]
auniyal61: ack.

api/v1beta1/conditions.go
```diff
@@ -183,4 +187,19 @@ const (
...
+	NovaNotificationMQNotRequestedMessage = "Notification message bus not requested"
+
+	// NovaNotificationMQReadyErrorMessage
+	NovaNotificationMQReadyErrorMessage = "Notification message bus creation failed: %s"
+
+	// NovaNotificationMQReadyCreatingMessage
+	NovaNotificationMQReadyCreatingMessage = "Notification message bus creation ongoing"
+
+	// NovaNotificationMQReadyMessage
+	NovaNotificationMQReadyMessage = "Notification message bus creation successfully"
```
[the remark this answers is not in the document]
auniyal61: done

controllers/common.go
auniyal61: done

controllers/novaapi_controller.go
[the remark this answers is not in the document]
auniyal61: added

config/crd/bases/nova.openstack.org_nova.yaml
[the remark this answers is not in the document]
bogdando: I fixed it there

api/v1beta1/conditions.go
```diff
@@ -172,4 +176,14 @@ const (
 
 	//CellHostDiscoverErrorMessage
 	CellHostDiscoverErrorMessage = "CellHostDiscover error occurred %s"
+
+	// notifications
+	// NovaNotificationMQReadyErrorMessage
+	NovaNotificationMQReadyErrorMessage = "Notification message bus creation failed: %s"
```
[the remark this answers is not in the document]
bogdando: @mrkisaolamb do you mean we should refactor API, cells and notifications related status messages into simple unified *messaging* statuses? I think that more granular statuses is exactly what user wants to get as it hints on the cause of possible issues much better than generic messages. @gibizer wdyt?

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,150 @@ import (
...
+				nova := GetNova(novaNames.NovaName)
+				nova.Spec.NotificationsBusInstance = &notificationsBus.BusName
+				g.Expect(k8sClient.Update(ctx, nova)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			infra.SimulateTransportURLReady(notificationsBus.TransportURLName)
+			transportURLName := infra.GetTransportURL(notificationsBus.TransportURLName)
+			// logger.Info("XXX", "t-url", transportURLName.Spec.RabbitmqClusterName)
+			// logger.Info("XXX", "t-url", notificationsBus.BusName)
+			// logger.Info("XXX", "t-url", notificationsBus.TransportURLName.Name)
```
bogdando, line 134: please remove after debugging is done

api/v1beta1/conditions.go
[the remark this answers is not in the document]
gibizer: I have no hard opinion about the unification of the message constant to e.g.: "%s message bus not started" or similar and customizing it at the call sites. But this can be can independently.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,167 @@ import (
...
+				condition.ErrorReason,
+				novav1.NovaNotificationMQReadyCreatingMessage,
+			)
+
+			infra.SimulateTransportURLReady(notificationsBus.TransportURLName)
+			transportURLName := infra.GetTransportURL(notificationsBus.TransportURLName)
+			// logger.Info("XXX", "t-url", transportURLName.Spec.RabbitmqClusterName)
+			// logger.Info("XXX", "t-url", notificationsBus.BusName)
+			// logger.Info("XXX", "t-url", notificationsBus.TransportURLName.Name)
+			Expect(notificationsBus.BusName).To(Equal(transportURLName.Spec.RabbitmqClusterName))
```
gibizer, line 148: nit. I would reorder to:
```golang
Expect(transportURLName.Spec.RabbitmqClusterName).To(Equal(notificationsBus.BusName))
```
as we expect that the transportURL is set accordingly.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,167 @@ import (
...
+
+			// verify if top level internal secret has notification-transport-url
+			th.ExpectConditionWithDetails(
+				novaNames.NovaName,
+				ConditionGetterFunc(NovaConditionGetter),
+				novav1.NovaNotificationMQReadyCondition,
+				corev1.ConditionFalse,
+				condition.ErrorReason,
+				novav1.NovaNotificationMQReadyCreatingMessage,
```
gibizer, line 157: This should be Ready now as SimulateTransportURLReady made the TransportURL ready.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,167 @@ import (
...
+			Expect(internalTopLevelSecret.Data["notification_transport_url"]).ShouldNot(BeEmpty())
+
+			assertHaveNotificationTransportURL := func(configData string) {
+
+				Expect(configData).ToNot(
+					ContainSubstring("[oslo_messaging_notifications]\ndriver = noop"))
+
+				Expect(configData).To(
+					ContainSubstring(
+						fmt.Sprintf("[oslo_messaging_notifications]\ntransport_url = rabbit://%s/fake\ndriver = messagingv2\nnotification_format=both", notificationsBus.TransportURLName.Name)))
```
gibizer, line 174: please wrap the long line

test/functional/nova_controller_test.go
gibizer:
```diff
diff --git a/test/functional/nova_controller_test.go b/test/functional/nova_controller_test.go
index 49d188ce..578a45a6 100644
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -148,13 +148,11 @@ var _ = Describe("Nova controller - notifications", func() {
                        Expect(notificationsBus.BusName).To(Equal(transportURLName.Spec.RabbitmqClusterName))
 
                        // verify if top level internal secret has notification-transport-url
-                       th.ExpectConditionWithDetails(
+                       th.ExpectCondition(
                                novaNames.NovaName,
                                ConditionGetterFunc(NovaConditionGetter),
                                novav1.NovaNotificationMQReadyCondition,
-                               corev1.ConditionFalse,
-                               condition.ErrorReason,
-                               novav1.NovaNotificationMQReadyCreatingMessage,
+                               corev1.ConditionTrue,
                        )
 
                        internalTopLevelSecret := th.GetSecret(novaNames.InternalTopLevelSecretName)
```

With this change the test passes to me.

gibizer (commented):
OK couple of things and it is good to go:
* add coverage for the ensureSecret change (see inline)
* build the change and test it in crc to see if notifications are emitted on the configured bus. You can rsh into the rabbitmq pod created for notifications and use rabbitmqctl to list queues for example to see that new messages are appearing on the notifications queue.

api/v1beta1/conditions.go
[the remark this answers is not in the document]
mrkisaolamb: I had a discussion with Amit a few days ago, and we agreed on a follow-up that will reduce the number of identical messages, using a format similar to what @gibizer suggested.

mrkisaolamb:
Thanks for the test fixes, @auniyal61 ! By the way, are we tracking somewhere the potential update issue related to the newer operator version that introduces notificationsBusInstance?  Or we shoud be good? I'm thinking about situation where cr doesn't have value and secret will have not have new transport url

auniyal61:
> By the way, are we tracking somewhere the potential update issue related to the newer operator version that introduces notificationsBusInstance? Or we shoud be good? I'm thinking about situation where cr doesn't have value and secret will have not have new transport url

we are not yet.
I am not sure from update perspective, when earlier CR was not aware of `notificationsBusInstance`, but aware now.

but I think in this scenario, `notificationsBusInstance` will be set to null, and it should not affect nova-operator in any negative way.

test/functional/nova_controller_test.go
auniyal61: done

test/functional/nova_controller_test.go
auniyal61: done

test/functional/nova_controller_test.go
auniyal61: done

auniyal61:
so using ceilometer, was able to see notification and metrics on a instance (vm1), but need to find out the real events (like rebuild, reboot) can easily be seen.

metrics: https://paste.opendev.org/show/b2qM3lqppXeOlJt475YF/ (using curl)
ceilometer logs: https://paste.opendev.org/show/bfdaJ9LFRYHDYMoMpzOs/

bogdando:
@auniyal61 well done!
the only remaining thing to address then is

> * add coverage for the ensureSecret change

templates/nova.conf
```diff
@@ -126,10 +126,14 @@ api_paste_config = /etc/nova/api-paste.ini
...
```
SeanMooney, line 132: this is still wrong 

so because github it annoying i cant leave this comment on line 141
but 

enable_ceilometer never because a thing so 
```
{{if (index . "enable_ceilometer") }}
[notifications]
notify_on_state_change = vm_and_task_state
{{ end }}
```

should be 

```
{{if (index . "notification_transport_url") }}
[notifications]
notify_on_state_change = vm_and_task_state
notification_format=both
{{ if eq .service_name "nova-compute"}}
versioned_notifications_topics = 'versioned_notifications','versioned_notifications_${HOST}'
{{ end }}
{{ end }}
```

and in the oslo_messaging_notifications section 

you need to add
```
{{ if and (eq .service_name "nova-compute") (index . "notification_transport_url")}}
topics = 'notifications','notifications_${HOST}'
{{ end }}
```

this will have the operato gerneat ethe correct notification endpoint of the computes, it will fall back to only the global topic on the other services
and it will properly enable both notification formats

you technically could generate the topic config in the ansible code but we can avoid that by using oslo interpolation like we do for my_ip in the vnc config.
on line 166

SeanMooney (changes requested):

templates/nova.conf
SeanMooney: just ot be clear this is an existing bug in the config template

notification_format is part of the notification section.

https://docs.openstack.org/nova/latest/configuration/config.html#notifications.notification_format

with the current template only the unversioned notifications are enable not both.

watcher can use either
celiomenter only uses unversioned 
we wanted  to only support version but we cant until ceilometer support that so we should default to both.

test/functional/nova_controller_test.go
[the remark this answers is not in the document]
auniyal61: done

auniyal61:
> * add coverage for the ensureSecret change

@bogdando this is added  in tests-commit [here](https://github.com/openstack-k8s-operators/nova-operator/pull/958/commits/cc0a34bf763f131e2a8bdec35ab2dc1d1289dd93#diff-f97810f6f4d2c225b7a506343c89128ea8fd6366c7b7ae17ad82f066d66f7272R182)

templates/nova.conf
SeanMooney: the datata plaen defienly can have  notification_format=both you are just not setting it in the right place.

auniyal61: so finals confs will  be

control plane
--------------------
```
[notifications]
notify_on_state_change = vm_and_task_state
notification_format=both

transport_url = rabbit://....
driver = messagingv2

```

dataplane
--------------
```
[notifications]
notify_on_state_change = vm_and_task_state
notification_format=both
versioned_notifications_topics = 'versioned_notifications','versioned_notifications_${HOST}'

topics = 'notifications','notifications_${HOST}'

```

auniyal61: @SeanMooney @gibizer

gibizer: @SeanMooney I would not enable the host based topic by default. The NVMe cleaning agent will not be deployed by default so we don't need those topics by default in our config. Those can be added via a custom config map attached to the DataplaneService when the NVMe cleanup feature is enabled.

@auniyal61 You need transport_url on the dataplane too.

My overall suggestion when there is notificationsBusInstance set is the following for both control and data plane:
```
[notifications]
notify_on_state_change = vm_and_task_state
notification_format=both

transport_url = rabbit://....
driver = messagingv2
```

SeanMooney: @gibizer ack i guess that makes sense the host based topics can be enable din the nova-extra-config as needed.

our overall default is to have notifications disbale and i guess the default
when a messaging bus would be to have the global topic (the default topic from upstream) and if the host based topics are required then that can be done via the config override.


gibi's example looks good to me

auniyal61:
```
[notifications]                                                                                                          
notify_on_state_change = vm_and_task_state                                                                                                                                                                                                              
notification_format=both                                                                                                                                                                                                                                
# notification_format=both specific to ceilometer as it depends on unversioned notifications                                                                                                                                                            
# bug https://bugs.launchpad.net/ceilometer/+bug/1665449                                                                                                                                                                                                
# while other services support versioned                                                                                 
# so we emit both verioned and unversioned if notifications are enabled.                          
                                                                                                                       
transport_url = rabbit://b13cd4bc-f75f-49ec-b439-0-notification-transport/fake                                                                                                                                                                          
driver = messagingv2                                                                                                                                                                                                                                    
                             
```

gibizer: this look OK. But the diff shows that you removed the case when driver is set to noop if the notification_transport_url is not provided. We need that.

controllers/novacell_controller.go
```diff
@@ -776,6 +779,7 @@ func (r *NovaCellReconciler) generateComputeConfigs(
 		"default_user_domain":    "Default",   // fixme
 		"compute_driver":         "libvirt.LibvirtDriver",
 		"transport_url":          string(secret.Data[TransportURLSelector]),
+		// "notification_transport_url": string(secret.Data[NotificationTransportURLSelector]),
```
gibizer, line 782: why this is commented out?

test/functional/nova_controller_test.go
```diff
@@ -38,6 +39,208 @@ import (
...
+			// assert in cell0-conductor conf
+			configDataMap = th.GetSecret(cell0.ConductorConfigDataName)
+			Expect(configDataMap.Data).Should(HaveKey("01-nova.conf"))
+			configData = string(configDataMap.Data["01-nova.conf"])
+			assertHaveNotificationTransportURL(configData)
+
+		})
+	})
+
+	When("notification cert is missing", func() {
```
gibizer, line 185: * this is not a cert
* this test is in a wrong file. If you are testing NovaAPI then please add that to novaapi_controller_test.go
* you need similar coverage for all the controllers where you added NotificationTransportURLSelector to the requiredSecretFields.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +39,208 @@ import (
...
+			cell0Account, cell0Secret := mariadb.CreateMariaDBAccountAndSecret(
+				cell0.MariaDBAccountName, mariadbv1.MariaDBAccountSpec{})
+			DeferCleanup(k8sClient.Delete, ctx, cell0Account)
+			DeferCleanup(k8sClient.Delete, ctx, cell0Secret)
+			memcachedSpec := infra.GetDefaultMemcachedSpec()
+			DeferCleanup(infra.DeleteMemcached, infra.CreateMemcached(novaNames.NovaName.Namespace, MemcachedInstance, memcachedSpec))
+			infra.SimulateMemcachedReady(novaNames.MemcachedNamespace)
+		})
+
+		BeforeEach(func() {
```
gibizer, line 205: you don't need two separate BeforEach blocks

gibizer (changes requested):

templates/nova.conf
bogdando: @gibizer @SeanMooney regarding "while other services support versioned". What would consume versioned notifications, not ceilometer, right? If we enable notifications for Nova by default during updates, and nothing consumes it, we will got a problem

controllers/novacell_controller.go
auniyal61: was testing something, before commit, and missed
thanks for noticing.

gibizer (commented):
Besides the necessary test changes noted previously this looks good to me.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +39,208 @@ import (
...
+			Expect(configDataMap.Data).Should(HaveKey("01-nova.conf"))
+			configData = string(configDataMap.Data["01-nova.conf"])
+			assertNotHaveNotificationTransportURL(configData)
+
+			// assert in cell0-conductor conf
+			configDataMap = th.GetSecret(cell0.ConductorConfigDataName)
+			Expect(configDataMap.Data).Should(HaveKey("01-nova.conf"))
+			configData = string(configDataMap.Data["01-nova.conf"])
+			assertNotHaveNotificationTransportURL(configData)
+
```
gibizer, line 115: * We need test coverage that the transport url is in the nova-cellX-compute-config secret. Note that that secret is not generated for cell0.

* We also need coverage for the ironic / fake compute config gets the transport url.

test/functional/novacell_controller_test.go
```diff
@@ -303,10 +303,27 @@ var _ = Describe("NovaCell controller", func() {
...
+
+				Expect(configData).To(
+					ContainSubstring(expectedConf1))
+
+				Expect(configData).To(
+					ContainSubstring(expectedConf2))
+
+			}
+
 			computeConfigData := th.GetSecret(cell1.ComputeConfigSecretName)
```
auniyal61, line 321: cell uses `IronicDriver` from `GetDefaultNovaComputeTemplate`

test/functional/novacell_controller_test.go
```diff
@@ -303,10 +303,27 @@ var _ = Describe("NovaCell controller", func() {
 				corev1.ConditionTrue,
 			)
 
+			assertHaveNotificationTransportURL := func(configData string) {
+
+				expectedConf1 := "[notifications]\nnotify_on_state_change = vm_and_task_state\nnotification_format=both"
+
+				expectedConf2 := fmt.Sprintf(
+					"[oslo_messaging_notifications]\ntransport_url = rabbit://notifications/fake\ndriver = messagingv2")
```
auniyal61, line 311: this is taking `transport_url = rabbit://notifications/fake` from `base_test.CreateCellInternalSecret`

test/functional/novaapi_controller_test.go
```diff
@@ -162,6 +162,42 @@ var _ = Describe("NovaAPI controller", func() {
...
+				th.ExpectCondition(
+					novaNames.APIName,
+					ConditionGetterFunc(NovaAPIConditionGetter),
+					condition.ReadyCondition,
+					corev1.ConditionFalse,
+				)
+			})
+
+			It("reports that the inputs are not ready", func() {
+				th.ExpectCondition(
```
gibizer, line 192: use ExceptionWithDetails to check the actual error message added to the condition at https://github.com/openstack-k8s-operators/nova-operator/blob/main/controllers/common.go#L280C1-L281C1

test/functional/novacell_controller_test.go
gibizer, line 1: you can have a similar test that you added to the novaconductor_controller_test.go about the input ready condition when the notifictation_transport_url key is missing from the cell secret.

test/functional/nova_compute_ironic_controller_test.go
```diff
@@ -139,6 +139,34 @@ var _ = Describe("NovaCompute controller", func() {
...
+
+			It("is not Ready", func() {
+				th.ExpectCondition(
+					cell1.NovaComputeName,
+					ConditionGetterFunc(NovaComputeConditionGetter),
+					condition.ReadyCondition,
+					corev1.ConditionFalse,
+				)
+			})
+
```
gibizer, line 167: you need to assert the input ready condition

test/functional/nova_compute_ironic_controller_test.go
gibizer, line 1: you can have a test case that asserts the content of the config-data secret for the computes here L192.

test/functional/novacell_controller_test.go
gibizer: this secret will have the ironic driver set NovaComputeConfigDataName. But the ComputeConfigSecretName always have the libvirt driver set.

test/functional/nova_controller_test.go
gibizer: * the compute config is covered in the novacell_controller_test.go
* the ironic part is covered in the nova_compute_irnoci_controller_test.go

test/functional/base_test.go
```diff
@@ -914,3 +945,39 @@ func GetSampleTopologySpec(label string) (map[string]interface{}, []corev1.Topol
...
+
+	Expect(configData).To(
+		ContainSubstring(expectedConf1))
+
+	Expect(configData).To(
+		ContainSubstring(expectedConf2))
+
+}
+
+func AssertHaveMockedNotificationTransportURL(configData string) {
```
gibizer, line 975: this can be implemented as :
```golang
AssertHaveNotificationTransportURL("notifications", configData)
```
or even use AssertHaveNotificationTransportURL("notifications", configData) every places where today AssertHaveMockedNotificationTransportURL is used.

test/functional/nova_controller_test.go
```diff
@@ -260,7 +379,7 @@ var _ = Describe("Nova controller", func() {
 			// proper content and the cell subCRs are configured to use the
 			// internal secret
 			internalCellSecret := th.GetSecret(cell0.InternalCellSecretName)
-			Expect(internalCellSecret.Data).To(HaveLen(2))
+			Expect(internalCellSecret.Data).To(HaveLen(3))
```
gibizer, line 382: you can add a check below that the 3rd key there is the notification_transport_url

test/functional/nova_controller_test.go
```diff
@@ -374,7 +493,7 @@ var _ = Describe("Nova controller", func() {
 			// assert that a the top level internal internal secret is created
 			// with the proper data
 			internalTopLevelSecret := th.GetSecret(novaNames.InternalTopLevelSecretName)
-			Expect(internalTopLevelSecret.Data).To(HaveLen(3))
+			Expect(internalTopLevelSecret.Data).To(HaveLen(4))
```
gibizer, line 496: you can add a check below that the 4th key is now the notification_transport_url

gibizer (commented):
OK. This looks good to me. If CI agrees I think we can land this. This means that when the CI is green we should unblock #948 , let it land, and then rebase this on the new main and then land this.

test/kuttl/test-suites/default/cell-tests/01-assert.yaml
```diff
@@ -115,6 +115,10 @@ status:
     reason: Ready
     status: "True"
     type: NovaMetadataReady
+  - message: Notification message bus not requested
+    reason: Ready
+    status: "True"
+    type: NovaNotificationMQReady
```
SeanMooney, line 121: is this a bug?

so by default the notificaotn message bus is not passed in.

and i don't think you are enabling it in this test so this condition should not be reported.

it should be skipped and only reported if the notifications are enabled no?

i could be wrong but i belive that is how we handle the conditions for the nova metadata when it runs at the top level vs the cell level.

as long as we are consistent, i don't particularly mind, but i would like to avoid divergence if we can.

controllers/nova_controller.go
```diff
@@ -391,6 +391,49 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+			instance.Status.Conditions.MarkTrue(
+				novav1.NovaNotificationMQReadyCondition, novav1.NovaNotificationMQReadyMessage)
+		default:
+			return ctrl.Result{}, fmt.Errorf("%w from  for the Notification MQ: %d",
+				util.ErrInvalidStatus, notificationMQStatus)
+		}
+	} else {
+		instance.Status.Conditions.MarkTrue(
+			novav1.NovaNotificationMQReadyCondition, novav1.NovaNotificationMQNotRequestedMessage)
+
```
SeanMooney, line 434: i chatted to gibi on slack baout this about and i think this is misleading

it imples the notification bus is deployed when its not in use.

I'm not sure this is enoch to block the current pr but it is not consistent with the nova metadata API conditions which are only reported when they are enabled at the cell or nova level.

SeanMooney: this should also proably be reporeed on the cell contoelr state too.

not just the top level.

i think initally this is configured cluster wide so i guess that is why this is only reporter dat the top level.

but long term we will need to include this in the cell template.

so perhaps per cell reporting is not required until that is addded.

SeanMooney (commented):
I have not done a super detailed review but over all this looks ok

the main gap i see is we have not enabled testing in either of the tempest jobs.

it would be nice to do that in a followup.


my main concern is with the condition reporting commented inline.

controllers/nova_controller.go
gibizer: > it imples the notification bus is deployed when its not in use.

I'm fine to make it consistent between metadata and notifications and remove the notification condition form the status if notifications are not enabled

> this should also proably be reporeed on the cell contoelr state too.
> not just the top level.

Today the TransportURL is created at the top level and there we know when the Rabbit and therefore the TransportURL CR is ready, this is reflected in the condition on the top level. On the cell level we only get a transport_url value in the internal secret so nothing to wait for, what is there just gets included into the cell specific compute config Secret. So from the cell perspective the important conditions are the 
* InputReady : are the internal secret available and has the expected keys
* NovaComputeServiceConfigReady: the cell specific compute config secret is created successfully

When we have a cellTemplate specific override (if ever) then we can discuss where to create the TransportURL for that. But note that the cell specific RPC TransportURL is still created on the top level not on the cell level for architectural reasons.

test/functional/base_test.go
auniyal61: done

test/functional/nova_controller_test.go
auniyal61: done

gibizer:
Thanks for fixing my test nits. Those changes look good. Let me know when you implemented the removal of the condition if notificationsBusInstance is empty.

test/functional/nova_controller_test.go
```diff
@@ -38,6 +38,125 @@ import (
...
+			DeferCleanup(th.DeleteInstance, CreateNovaWithCell0(novaNames.NovaName))
+
+			keystone.SimulateKeystoneServiceReady(novaNames.KeystoneServiceName)
+			mariadb.SimulateMariaDBDatabaseCompleted(novaNames.APIMariaDBDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(novaNames.APIMariaDBDatabaseAccount)
+			mariadb.SimulateMariaDBDatabaseCompleted(cell0.MariaDBDatabaseName)
+			mariadb.SimulateMariaDBAccountCompleted(cell0.MariaDBAccountName)
+			infra.SimulateTransportURLReady(cell0.TransportURLName)
+			SimulateReadyOfNovaTopServices()
+		})
```
SeanMooney, line 75: we coudl have had an addtional check here to assert that the 
novav1.NovaNotificationMQReadyCondition, is not in the list of status conditions.

but that the over all ready condition is true.

SeanMooney (commented):
/lgtm

there is one minor testing change that we could do but i think I'm fine to merge this

gibizer:
/unhold
Content looks good. The CRD prereq #948 PR merged yesterday. So this is good to go. I will approve it once CI is green...

templates/nova.conf
auniyal61: > What would consume versioned notifications, not ceilometer, right?
cinder I think.

> If we enable notifications for Nova by default during updates, and nothing consumes it, we will get a problem
I think as a broadcaster we should not, its not like we need any locking here.

bogdando:
ok, we need https://github.com/openstack-k8s-operators/nova-operator/pull/985 to land then retry with cherry picking

PR #963, opened 2025-04-29
Updates webhook port number

auniyal61:
This allows to run both openstack and nova operator webhook in same deployment

### DIFF
--- a/config/default/manager_webhook_patch.yaml
+++ b/config/default/manager_webhook_patch.yaml
@@ -9,7 +9,7 @@ spec:
       containers:
       - name: manager
         ports:
-        - containerPort: 9443
+        - containerPort: 9444
           name: webhook-server
           protocol: TCP
         volumeMounts:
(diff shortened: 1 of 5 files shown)


auniyal61:
recheck 
unrelated build failed  with
`Error: initializing source docker://quay.io/openstack-k8s-operators/designate-operator-bundle:9adfe79423df155bbec9a909a54d8ab3d3b4637c: reading manifest 9adfe79423df155bbec9a909a54d8ab3d3b4637c in quay.io/openstack-k8s-operators/designate-operator-bundle: received unexpected HTTP status: 502 Bad Gateway`

SeanMooney (approved):
i guess this looks good to me

we already run the metrics and prob endpoint on non default ports for a similar reason so this is inline with that.

SeanMooney:
ill leave other add approved if they agree

karelyatin:
Instead of hard coding ports across repos, would be good to make it configurable like METRICS_PORT and HEALTH_PORT ?

mrkisaolamb (approved):
lgtm,  I agree with Yating that making it configurable, like METRICS_PORT and HEALTH_PORT, is a good idea. However, it might be better to handle this in a separate issue/PR

PR #983, opened 2025-06-17
Backport Notfications message bus 

auniyal61:


gibizer (changes requested):
FR2 was released almost 3 months ago this backport is not valid.

gibizer:
FR2 was released almost 3 months ago this backport is not valid.

PR #986, opened 2025-06-18
Fix SetDefaultRouteAnnotations reciver type

mrkisaolamb:
This method need to be call from openstack operator lvl which is using NovaSpecCore in template

JIRA: [OSPRH-10955](https://issues.redhat.com//browse/OSPRH-10955)


bogdando (approved):
LGTM

PR #990, opened 2025-07-02
Migrate nova-operator from single to multi-group

ratailor:
This change migrate nova-operator from single group operator to support multi-group operators, which is required to merge placement-operator code base from placement-operator into nova-operator.

Related: [OSPRH-14957](https://issues.redhat.com//browse/OSPRH-14957)


bogdando:
lgtm as this transition looks technically correct, albeit the upgrade path should be covered as well (perpahs in the meta oscp operator)

paramite (commented):
Just out of curiosity, do we plan to have more that one KUTTL test suite for a service? I wonder why the kuttl tests moved to service subfolder while there could be default suite which deploys and verifies both nova and placement and then suite just for nova resources tests and suite just for placement resources tests?

mrkisaolamb:
This initiative was closed by this PR: https://github.com/openstack-k8s-operators/nova-operator/pull/1105

PR #993, opened 2025-07-10
Fix typos in code

ratailor:
This change fixes typos in code in job-name, echo message and function name.


SeanMooney (commented):
by the way i was expecting code spell to catch some of these but 
i guess i never actually got around to turning that on for this repo.
i would suggest adding the following to the pre-commit hooks and fixing any other issues it finds

  - repo: https://github.com/codespell-project/codespell
    rev: v2.4.1
    hooks:
      - id: codespell

.pre-commit-config.yaml
```diff
@@ -76,6 +76,12 @@ repos:
       # E020: Function declaration not in format ^function name {$
       # E012: here doc didn't end before EOF
 
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.4.1
+  hooks:
+    - id: codespell
+      args: ['--ignore-words=.codespell.ignore']
```
stuggi, line 83: maybe we should add this to the pre-commit config of all repos?

SeanMooney: i would encourage that, i try to convince folks to adopt it in all repos i work on but the need for it decreases when im not submitting code there :)

on a more serious note i dislike typo only fixes without a way to prevent there reintroduction, we have been using code spell for this upstream now for a few years.

the reason we selected it is that code spell is not a traditional spell checker.
instead of having a list of known valid word, it has a dictionary of know common misspelling derived form opensource contributions.

that means it has a very low false positive rate.
typically when you have false positives its because of a jargon term which is why we typically add a per repo dictionary file to ignore those terms of art that arise in a given repo.

PR #1001, opened 2025-07-25
[WIP] disable live migration workaround

SeanMooney:
Signed-off-by: Sean Mooney <work@seanmooney.info>

### DIFF
--- a/templates/nova.conf
+++ b/templates/nova.conf
@@ -201,9 +201,6 @@ disable_fallback_pcpu_query=true
 [workarounds]
 enable_qemu_monitor_announce_self=true
 reserve_disk_resource_for_image_cache=true
-# NOTE(gibi): We need this as live migration does not work with
-# cpu_mode=host-model . See https://bugs.launchpad.net/nova/+bug/2039803
-skip_cpu_compare_on_dest = true
 {{end}}
 
 


gibizer:
I think we cannot remove this until we solve https://bugs.launchpad.net/nova/+bug/2039803

gibizer:
I reproduced the original issue today:
* deployed 18 with crc install_yamls and two VMs as EDMP computes
* created a VM
* live migrated the VM back and forth, no issue
* reconfigured nova in the EPDM deployment to have skip_cpu_compare_on_dest = false
* tried to live migrate the VM again and it fails with 
```
Jul 31 15:17:34 edpm-compute-0 nova_compute[354378]: 2025-07-31 15:17:34.056 2 ERROR oslo_messaging.rpc.server nova.exception.MigrationPreCheckError: Migration pre-check error: Unacceptable CPU info: CPU doesn't >
```
The two EDPM compute VM is identically configured via virsh so live migration should be possible between them but nova rejects it.

Nova is configured with cpu_mode=host-model. 
```
[root@edpm-compute-0 ~]# grep cpu_ /var/lib/openstack/config/nova -R
/var/lib/openstack/config/nova/01-nova.conf:initial_cpu_allocation_ratio=4.0
/var/lib/openstack/config/nova/01-nova.conf:# cpu_mode=host-model . See https://bugs.launchpad.net/nova/+bug/2039803
/var/lib/openstack/config/nova/01-nova.conf:skip_cpu_compare_on_dest = true
/var/lib/openstack/config/nova/01-nova.conf:cpu_mode=host-model
/var/lib/openstack/config/nova/01-nova.conf:cpu_power_management=false
/var/lib/openstack/config/nova/25-nova-extra.conf:skip_cpu_compare_on_dest = false
```

So this shows that at least using cpu-mode=host-model prevents an otherwise valid live migration. I think this is a valid bug.

PR #1011, opened 2025-08-26
refactor: use GetKeystoneAPIByName instead of label-based lookup

ratailor:
- Replace GetKeystoneAPI with GetKeystoneAPIByName in nova_controller.go
- use instance.Spec.KeystoneInstance directly instead of empty label map
- Remove TODO comment about changing the API usage.
- Update keystone-operator/api dependency to support new functionality

This provides a more direct and efficient way to get the KeystoneAPI resource by name rather then searching by labels.

Commit message assisted by: claude-4-sonnet


mrkisaolamb:
This PR is outdated and hasn't been updated for a long time, so I'm closing it

PR #1014, opened 2025-09-16
Add missing hook step in vars

danpawlik:
The Ceph playbook is not executed in hook. When the playbook uses hook role, it should contain a post_ceph variable, otherwise the Ceph playbook would not be executed and it would fail with error:

    - 'error: the path "/tmp/k8s_ceph_secret.yml" does not exist'

### DIFF
--- a/ci/nova-operator-tempest-multinode-ceph/ci_fw_vars.yaml
+++ b/ci/nova-operator-tempest-multinode-ceph/ci_fw_vars.yaml
@@ -28,6 +28,13 @@ post_deploy:
     type: playbook
     source: control_plane_kustomize_deploy.yml
 
+post_ceph:
+  - name: 80 Run Ceph hook playbook
+    type: playbook
+    source: ceph.yml
+
+cifmw_cephadm_log_path: /home/zuul/ci-framework-data/logs
+
 # we can use this hook to create flavors, images, etc
 # before we execute tempest simiilar to how we would use
 # local.sh in devstack based jobs


karelyatin:
/lgtm
Needed as an issue triggered by https://github.com/openstack-k8s-operators/ci-framework/pull/3154

mrkisaolamb (approved):
lgtm

karelyatin (approved):
/lgtm
works fine now

danpawlik:
recheck

ci/nova-operator-tempest-multinode-ceph/ci_fw_vars.yaml
```diff
@@ -28,6 +28,13 @@ post_deploy:
     type: playbook
     source: control_plane_kustomize_deploy.yml
 
+post_ceph:
+  - name: 80 Run Ceph hook playbook
+    type: playbook
+    source: ceph.yml
+
+cifmw_cephadm_log_path: /home/zuul/ci-framework-data/logs
```
SeanMooney, line 36: nit we shoudl avodi hardcodign the home dir as /home/zuul in general.

this makes it harder ot reuse the vars files with repoducer script.

on the ohter hand i have never got those to work proably so this is just a nit.

where possible we shoudl use the ansible_user home dir via 
{{ ansible_user_dir }}


https://github.com/openstack-k8s-operators/ci-framework/blob/423b9b6c98abbe0761ae5c70aa9e18467f5850ea/roles/hive/README.md?plain=1#L80

https://github.com/openstack-k8s-operators/ci-framework/blob/423b9b6c98abbe0761ae5c70aa9e18467f5850ea/scenarios/centos-9/edpm_ci.yml#L2

danpawlik: I agree.

There is a dedicated tasks that would remove hardcoded zuul as a user.
In sooo many places zuul is hardcoded, that if it is here, there would be no difference. Trust me.

PR #1015, opened 2025-09-19
test: Fix race condition in NovaNoVNCProxy topology test

ratailor:
The "updates lastAppliedTopology in NovaNoVNCProxy .Status" test was failing intermittently due to a race condition in test setup.

**Problem**: SimulateMemcachedReady() was called after CreateNovaNoVNCProxy(), causing the controller's ensureMemcached() call to fail intermittently when reconciliation triggered before the Memcached instance was simulated as ready.

**Solution**: Move SimulateMemcachedReady() call before CreateNovaNoVNCProxy() to ensure Memcached is ready before controller reconciliation begins.

Commit-message-assisted-by: claude-4-sonnet
Closes: OSPRH-19625


gibizer:
@ratailor Thanks. Indeed I cannot reproduce the problem any more on my machine either.

PR #1017, opened 2025-09-25
Change CI jobs to improve stability

mrkisaolamb:


gibizer (approved):
CI is green. I assume you had some information how this makes the job more stable. So lets go with it.

PR #1018, opened 2025-09-26
Modernize Go code with latest style conventions

fmount:
Modernize using `gopls modernize tool` to update:

- Replace interface{} with any type alias (Go 1.18+)
- Update range loops to use idiomatic range syntax
- Replace fmt.Sprintf with fmt.Appendf where appropriate

These changes improve code readability and align with current Go best practices.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -20,6 +20,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"maps"
 	"net/url"
 	"sort"
 	"strconv"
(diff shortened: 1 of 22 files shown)


gibizer (approved):
If CI is happy then this is fine by me

fmount:
recheck

SeanMooney:
im fin with this over all but one ask
when we do thing like this we shoudl condier cna we incoperate this into pre-commit/make so that its automatically applied if you use the pre-commit tool, runablle by folks if you dont and enforced in ci so that we dont regress.

that can totally be a follow up enhancement if we agree its good to do on an ongoing basies rahter then just when we are bumping the go verions.

fmount:
> im fin with this over all but one ask when we do thing like this we shoudl condier cna we incoperate this into pre-commit/make so that its automatically applied if you use the pre-commit tool, runablle by folks if you dont and enforced in ci so that we dont regress.
> 
> that can totally be a follow up enhancement if we agree its good to do on an ongoing basies rahter then just when we are bumping the go verions.

Right, I agree with you and I have a plan to follow up and introduce it in our Makefile to have the pre-commit task. I was still looking at the module and see how it behaves, but that's the plan I'm going to propose to the broader group.

PR #1082, opened 2026-03-05
Return early when notification TransportURL is not ready

auniyal61:
When a RabbitMQ notification pod is deleted or restarted, the transporturl status changes from ready to others(creating).

This change make reconciler return early when rabbitmq status is creating or failed, this prevents empty transporturl from being propogated in configs.

Adds tests to verify same

Resolves: [OSPRH-26269](https://issues.redhat.com/browse/OSPRH-26269)

### DIFF
--- a/test/functional/nova_controller_test.go
+++ b/test/functional/nova_controller_test.go
@@ -139,6 +139,61 @@ var _ = Describe("Nova controller - notifications", func() {
 
 			infra.AssertTransportURLDoesNotExist(notificationsBus.TransportURLName)
 		})
+
+		It("returns early when notification TransportURL becomes not ready", func() {
+			notificationsBus := GetNotificationsBusNames(novaNames.NovaName)
+			DeferCleanup(k8sClient.Delete, ctx, CreateNotificationTransportURLSecret(notificationsBus))
+
+			// add notification bus and simulate it ready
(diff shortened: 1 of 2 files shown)


gibizer (changes requested):

test/functional/nova_controller_test.go
```diff
@@ -139,6 +139,63 @@ var _ = Describe("Nova controller - notifications", func() {
...
+					Cluster: notificationsBus.BusName,
+				}
+				g.Expect(k8sClient.Update(ctx, nova)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+			infra.SimulateTransportURLReady(notificationsBus.TransportURLName)
+
+			// wait for notification MQ to be ready
+			Eventually(func(_ Gomega) {
+				th.ExpectCondition(
```
gibizer: ExpectCondition already do the check in an Eventually block, so you don't need an explicit Eventually block here.

gibizer (changes requested):
two small changes in the test otherwise this looks good to me

test/functional/nova_controller_test.go
auniyal61: done

gibizer (approved):
Thanks. Looks good

PR #1090, opened 2026-03-27
Restructure templates and tests directory to multi-group layout

auniyal61:
Moves all Nova templates and tests and drops 'nova prefix' from subdirectories
templates under templates/nova/
tests under test/functional/nova/ and updates relative paths.


auniyal61:
recheck

```
2026-03-27 10:25:47.839719 | controller |
2026-03-27 10:25:47.839736 | controller | TASK [edpm_prepare : Wait for OpenStack controlplane to be deployed _raw_params=oc wait OpenStackControlPlane {{ _ctlplane_name }} --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} --for=condition=ready --timeout={{ cifmw_edpm_prepare_timeout }}m] ***
2026-03-27 10:25:47.839749 | controller | Friday 27 March 2026  09:55:47 +0000 (0:00:30.068)       0:09:36.537 **********
2026-03-27 10:25:47.839761 | controller | Friday 27 March 2026  09:55:47 +0000 (0:00:30.068)       0:09:36.535 **********
2026-03-27 10:25:47.839780 | controller | task path: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/edpm_prepare/tasks/kustomize_and_deploy.yml:112
2026-03-27 10:25:47.839790 | controller | fatal: [localhost]: FAILED! =>
2026-03-27 10:25:47.839801 | controller |     changed: true
2026-03-27 10:25:47.839811 | controller |     cmd:
2026-03-27 10:25:47.839821 | controller |     - oc
2026-03-27 10:25:47.839831 | controller |     - wait
2026-03-27 10:25:47.839841 | controller |     - OpenStackControlPlane
2026-03-27 10:25:47.839850 | controller |     - controlplane
2026-03-27 10:25:47.839860 | controller |     - --namespace=openstack
2026-03-27 10:25:47.839870 | controller |     - --for=condition=ready
2026-03-27 10:25:47.839880 | controller |     - --timeout=30m
2026-03-27 10:25:47.839889 | controller |     delta: '0:30:00.107131'
2026-03-27 10:25:47.839899 | controller |     end: '2026-03-27 10:25:47.804036'
2026-03-27 10:25:47.839910 | controller |     msg: non-zero return code
2026-03-27 10:25:47.839919 | controller |     rc: 1
2026-03-27 10:25:47.839929 | controller |     start: '2026-03-27 09:55:47.696905'
2026-03-27 10:25:47.839965 | controller |     stderr: 'error: timed out waiting for the condition on openstackcontrolplanes/controlplane'
2026-03-27 10:25:47.839987 | controller |     stderr_lines:
2026-03-27 10:25:47.839997 | controller |     - 'error: timed out waiting for the condition on openstackcontrolplanes/controlplane'
2026-03-27 10:25:47.840007 | controller |     stdout: ''
2026-03-27 10:25:47.840017 | controller |     stdout_lines: []
2026-03-27 10:25:47.840030 | controller |

```

auniyal61:
recheck

auniyal61:
recheck

gibizer:
```
2026-03-30T13:04:44Z	ERROR	Reconciler error	{"controller": "novaconductor", "controllerGroup": "nova.openstack.org", "controllerKind": "NovaConductor", "NovaConductor": {"name":"nova-cell0-conductor","namespace":"openstack"}, "namespace": "openstack", "name": "nova-cell0-conductor", "reconcileID": "7c27a893-df9a-48c8-a492-843cf0c1bdf4", "error": "Internal error occurred: Job Attempt #1 Failed. Check job logs"}
```

It seems cell0 db sync job fails. I suggest to reproduce the deployment and check the job logs. It seems that the must gather does not contain  them.

test/functional/nova/conductor_controller_test.go
gibizer: this comment refers to an old path. Please grep the repo for these kind of old references

auniyal61: thanks done

internal/controller/nova/common.go
```diff
@@ -520,7 +546,7 @@ func (r *ReconcilerBase) generateConfigsGeneric(
 			Name:               configName,
 			Namespace:          instance.GetNamespace(),
 			Type:               util.TemplateTypeConfig,
-			InstanceType:       instance.GetObjectKind().GroupVersionKind().Kind,
+			InstanceType:       getTemplateInstanceType(instance),
```
gibizer, line 549: Ohh so lib-common "hardcodes" the structure of the templates directory to be only structured by the CRD Kind even if we have a multigroup operator where there are different groups of kinds in the operator. This was probably not a problem on infra-operator as each group there has a single kind in it, but here we have the nova group with multiple kinds and there will be cyborg group with multiple kinds. 

@stuggi how do you feel about this? I personally would like to have the structure in templates following our structure in the code. The code is structured now that impl specific to a given group (e.g. nova) is under a nova subdirectory (e.g api/nova/v1beta1, internal/controller/nova internal/webhook/nova). So I would like to do templates/nova/<~kind> as well. Should we extend lib-common to handle this more gracefully? The current solution basically contradicts with field name InstanceType as we are not passing a type we are passing a directory path instead.

An extra twist that we started removing the extra "nova" prefix from the pathnames as now that is signified in a specific "nova" directory in the path. I.e. we don't have templates/novaapi neither we have templates/nova/novaapi but we have templates/nova/api .

internal/controller/nova/nova_controller.go
```diff
@@ -1146,14 +1146,14 @@ func (r *NovaReconciler) ensureNovaManageJobSecret(
 			Name:         scriptName,
 			Namespace:    instance.Namespace,
 			Type:         util.TemplateTypeScripts,
-			InstanceType: "nova-manage",
```
gibizer, line 1149: @stuggi as you are already pulled in, this is another example where the "CRD kind" mapping is not applicable for templating. Here we are creating config for a ~"Job kind" as that job calls a nova CLI it expects the same configuration as a nova (~nova-conductor) service, but we actually have 3 different instances of this nova-manage job kind, with 3 different script to execute.

internal/controller/nova/common.go
stuggi: Using the kind was the easiest to map a kind to a template dir. When we created that, multi-group was not a thing. We could make the func smarter for the use case you describe.
What you currently can already to is to use the `AdditonalTemplate` to add templates which are not part of the kind dir. This is currently used where you have a common template, like:
```
                        AdditionalTemplate: map[string]string{
                                "ironic.conf": "/common/config/ironic.conf",
                        },
```

Downside is that you'd have to list all the templates.

internal/controller/nova/nova_controller.go
stuggi: right, for this we initially added the `TemplateTypeNone` , do not lookup for a kind and then use  `AdditionalTemplate` to reference the needed templates. Would that work?

internal/controller/nova/common.go
auniyal61: the concern about improving template structure would require lib-common changes first. Can we merge this PR as-is and fix lib-common and other affected multi-group operators (if any) in follow-ups?

SeanMooney: so i originally deviating form the way this works in lib common because i didnt want to have kind based discovery of template in general.

i explicitly didn't want to have inb/config dirs per kind, 
when we were first doing the bring up.
But over time we kind of ended up havign them for shell scipts 

originally i want to have a single flat template directory with all template in that single dir and rely mainly on extraTemplatee so pass the file explcity.

so listing all the templst is what i wanted to do form day 1 so if there is no technical object to thtat i woudl peropsoe

templates/nova
templates/placement
templsates/cyborg

and have that directory be flat. no bin/config seperation within them

SeanMooney: to be clear im suggesting we do not have bin an config or other sub dirs for each service

so /template/nova/metadata-httpd.conf
not /template/nova/metadata/config/httpd.conf

SeanMooney: by the way these are template so i think we can have a singel httpd.conf and just share it between the metadata api and the main api im just suggeting not changign the content in this pr and following up with that later

SeanMooney: can we just pass nil for InstanceType by the way to disable the kind based lookup? is that the chagne you are saying we would have to make to lib-common

im ok with deferring that for now and looking at restructuring this more later

gibizer: @stuggi thanks for you reply. The additionalTemplate way is bit better than passing a path to a field that expects a kind (even if it works today) from code readability perspective. 

@SeanMooney My concern about the flatter structure, and the use of additionalTemplate to support it, is that it moves the decision of file incluson into the codebase. So we will only know what is passed to the pod by either examining a pod or reading the code. Today I can infer most of it from the structure of the git repo as well. Other the then httpd.conf and the related sshl.conf I don't see any reusable piece of files in the per service template dirs. Or do you suggest to template the kolla configs as well and fill the template form code?

---

I have no hard opinion. (I feel less and less ownership over this repository due to not spending time working in it). For the sake of progress I suggest to land this patch. It works. Then have one or more follow up to fine tune the structure. 
@SeanMooney does cyborg-operator work started somewhere? What pattern they are using there? (Placement-operator uses the "kind"/config directory pattern today).

internal/controller/nova/nova_controller.go
gibizer: It might. I haven't tried. I'm not sure how I feel about going from the autodiscovery to the manually listing files solutions. I liked having an explicit directory structure to know where to look for the files of a given "type" of pod (being a service pod or a job pod). But I also see that the current automation is hackish for us.

internal/controller/nova/common.go
SeanMooney: so my orgianlly reason for wanted it to be flato was partly becasue if we add a file we shoudl also add a test for that file so there should never be a casue wehre we just drop in a file to the git repo and it just end up in the pod.

but the main reason was ot align to how kolla ansibel works.
if the intreface in lib-common took a a directory path i would be more ok with it but its always felt a bit too magic the way it works today and less discoverabel to me as a result but im ok with the status quo if we want to maintain it

SeanMooney: for the kolla config no i was gogn to suggest keping those seperate
we could template them i guess but that was not my intent.

cybrog is started but it following the watcher convetion which is what nova does today since that was the refence for watcher.

in the very first patches for the config gen i wrote i started with a flat stucre because that predated the existign fo the templateing code in lib-common
i didnt like the interface expsoe by lib-common when i adapted my code to it but it was also not a high enough priroty to fix which is why i raised it now since we are revisiting it so it more a nice to have form my perspect then a we shoudl ro must

what ever pattern we choose now is what we would use for everythign else.

gibizer: >  but its always felt a bit too magic the way it works today and less discoverabel

I agree with that. The kind-path mapping is a magic in the codebase indeed. This is why I did not like abusing that logic here in this patch further by passing a complex path just because we can. 

At the same time additionalTemplate also feel like a lucky workaround available and not like a well made solution. I.e. it is "additional" so not the main way of passing files.

I guess what I would do (given time and resources) is to  transform the lib-common interface to be explicit about taking a dir path and pulling files from that dir. 

> so my orgianlly reason for wanted it to be flato was partly becasue if we add a file we shoudl also add a test for that file so there should never be a casue wehre we just drop in a file to the git repo and it just end up in the pod.

I guess this would be some kind of protection against unintentional exposure of additional files to the pods. And I'm not sure this is the good tool for such a protection. If we did not notice a new file being added during code review then in that same code review we might ignore the envtest change as well. But sure I have no better ideas for such protection. 

BTW, I think the current test can easily be updated to assert the number of keys in the config CM / Secret. So if unintentional file is added the test fails.

> but the main reason was ot align to how kolla ansibel works.

Could you expand on this?

> what ever pattern we choose now is what we would use for everythign else.

+100. 

> for the kolla config no i was gogn to suggest keping those seperate
> we could template them i guess but that was not my intent.

I agree to keep them separate, not templated.

---
I don't see pushing against landing this. So I will do one more read through of the patch itself and if I find nothing I will land it. We can iterate on this dir structure pretty freely as it is totally operator internal as far as I understand,  it does not cause update/upgrade issues.

As the next big change for nova-operator is to add cyborg into it. So maybe we can get some time out from that work to make this structuring right not only on nova-operator level but also in lib-common and in other operators.

gibizer (approved):

internal/controller/nova/common.go
SeanMooney: sure so the orgianl proape i brought forward a few year ago about adopting kolla ansibel config gen with "config override" was base in my expirince usign and contibuting to it

if we look a kolla ansible each role (each service/operator repo in our case )
has a single template folder

https://github.com/openstack/kolla-ansible/tree/master/ansible/roles/neutron/templates

back in 2023 i porpsoed direfctly adopting kollas approach to customization 
https://github.com/openstack/kolla-ansible/blob/master/doc/source/admin/deployment-philosophy.rst#kollas-solution-to-customization
including originally using jinja2 format which later evolved to go template format.

both for the files we render and for the crd files for the overrides.

we ened up with a hybrid that was mostly based on the core idea of kollas config system but with some modification like the addtion of the bin and config subdirs

with in the single template folder at the root of the repo.

we never got around to allowing template to be passed in the the input of the crds although we could if the need ever came up.

orginally since i also envisioned the ansibel runing fomr the service operator i wanted to effectivly allow access within the template to the content of the inventory ectra just as you do for ansible roels when you have ansible/jinja templates.

anyway that getting a bit of topic but that was my orginal hope for the config system.

flat directory stucture with all files encoded as templates, sharing templstaes between pod when easy or separate when not
you notice that kolla has sepeate kolla json files per service named after the service
https://github.com/openstack/kolla-ansible/blob/master/ansible/roles/nova/templates/nova-metadata.json.j2
as i was sugesting above

anyway we can revisit this later.

PR #1093, opened 2026-04-07
Remove unnecessary allow-merge comment from go.mod

gibizer:
The local api directory override is allow listed in the lint github action, so no need for the explicit allow-merge comment

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -117,7 +117,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-replace github.com/openstack-k8s-operators/nova-operator/api => ./api //allow-merging
+replace github.com/openstack-k8s-operators/nova-operator/api => ./api
 
 // mschuppert: map to latest commit from release-4.18 tag
 // must consistent within modules and service operators


go.mod
```diff
@@ -117,7 +117,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-replace github.com/openstack-k8s-operators/nova-operator/api => ./api //allow-merging
+replace github.com/openstack-k8s-operators/nova-operator/api => ./api
```
auniyal61, line 120: I think I added this when we changed to apis, and that time CI build was failing without this comment w.r.t openstack-operator build.
but then again as we changed back to api, this might not be needed now if CI passed

will wait for CI to pass.

gibizer: Ohh OK. apis / api could have caused that. Still if we had gone with the "apis" dir then I would have changed the regex in the lint github action to match with the new apis dir naming, as the intention in the lint is clear, the local api/apis override is always allowed.

auniyal61 (approved):
/lgtm

PR #1096, opened 2026-04-09
Use MultiTemplateDir for the support of multi-segment dir in template

amartyasinha:
With the restructure of the templates dir to have multi-segment dir, this commit ensures the proper templating from these dirs.

lib-common change introducing MultiTemplateDir: https://github.com/openstack-k8s-operators/lib-common/pull/682


bogdando:
So for Kinds that already had explicit cases in #1090, `GetTemplateSubpath` is intended to reproduce the same `nova/...` paths. For Kinds that previously fell through the old default ``strings.ToLower(kind)``, this PR implicitly switches them to the `GetTemplateSubpath` rule (NovaCell -> `nova/cell` instead of `novacell`). The
  PR, as shown, does not:
  * document that audit,
  * add per-Kind overrides, or
  * move template trees for those Kinds (create templates/nova/cell/... and retire templates/novacell/... etc)

amartyasinha:
This PR intends to remove the InstanceType overloading (done in #1090), and leave `InstanceType` to only have kind as value. `MultiTemplateDir ` is an optional parameter, when used will take over to set the subdir during templating. Only when this parameter is set, `GetTemplateSubpath` will be executed.

Regarding 
> * document that audit,

I did not get where to audit this change. Should I amend the commit message to include all details, or add them somewhere else?
> * add per-Kind overrides, or

Sorry, did not get it. Could you please explain it a bit (maybe an example would be helpful)
> * move template trees for those Kinds (create templates/nova/cell/... and retire templates/novacell/... etc)

IIUC, #1090 has already moved template trees for all kinds as `templates/nova/<kind>/..`, and we do not have `templates/nova<kind>/..`. Please correct me if I've missed something.

internal/controller/nova/common.go
```diff
@@ -541,12 +515,14 @@ func (r *ReconcilerBase) generateConfigsGeneric(
...
 	maps.Copy(extraTemplates, additionalTemplates)
+	gvk := instance.GetObjectKind().GroupVersionKind()
 	cms := []util.Template{
 		{
 			Name:               configName,
 			Namespace:          instance.GetNamespace(),
 			Type:               util.TemplateTypeConfig,
-			InstanceType:       getTemplateInstanceType(instance),
+			InstanceType:       gvk.Kind,
+			MultiTemplateDir:   util.GetTemplateSubpath(gvk.Group, gvk.Kind),
```
gibizer: I think this is where the change go too far by trying to *generate* the path instead of just defining the path per service. Trying to support the generation is just unnecessary complexity. Simply define the path per service in the nova-operator then we can avoid this complexity.

internal/controller/nova/nova_controller.go
```diff
@@ -1143,17 +1143,19 @@ func (r *NovaReconciler) ensureNovaManageJobSecret(
...
-			Name:         scriptName,
-			Namespace:    instance.Namespace,
-			Type:         util.TemplateTypeScripts,
-			InstanceType: "nova/nova-manage",
-			Labels:       cmLabels,
+			Name:             scriptName,
+			Namespace:        instance.Namespace,
+			Type:             util.TemplateTypeScripts,
+			InstanceType:     instance.GetObjectKind().GroupVersionKind().Kind,
+			MultiTemplateDir: "nova/nova-manage",
```
gibizer, line 1150: Yepp this is the correct use of the new interface. 

> Explicit is better than implicit.
> Simple is better than complex.

internal/controller/nova/common.go
gibizer: * You probably need a new param here to pass the template dir.
* Or you can try to change the `instance client.Object` to be a specific interface type where the interface defines a getter for the template dir and then implement that 
interface function for our Objects
* Or you can add the getter for the ReconcileBase and then implement it in each Reconciler to return the service specific path.

gibizer (changes requested):
You can try to run a build and test in CI by temporarily adding the new lib-common hash to the go.mod file so the build can see the new field.

internal/controller/nova/common.go
amartyasinha: ack! I thought we need to generate path from kind, i.e. `NovaAPI -> nova + api`. Since we can define the path directly instead of generating, I'll remove the generation functions from `lib-common`, and close the `CRD` spec modification PR.

internal/controller/nova/common.go
amartyasinha: added a new param `templateDir`.

internal/controller/nova/common.go
amartyasinha: defined the path instead of generating them

amartyasinha:
recheck nova-operator-kuttl

internal/controller/nova/nova_controller.go
```diff
@@ -2145,7 +2147,7 @@ func (r *NovaReconciler) ensureCellSecret(
 		Name:         secretName,
 		Namespace:    instance.Namespace,
 		Type:         util.TemplateTypeNone,
-		InstanceType: getTemplateInstanceType(instance),
+		InstanceType: instance.GetObjectKind().GroupVersionKind().Kind,
 		Labels:       labels,
 		CustomData:   data,
```
gibizer, line 2153: I guess we are not loading any template from file here hence we don't need to pass a templateDir. If I'm correct then a comment would be nice here about it.

go.mod
```diff
@@ -144,3 +144,5 @@ replace k8s.io/component-base => k8s.io/component-base v0.31.14 //allow-merging
 replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20250929174222-a0d328fa4dec //allow-merging
 
 replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20250627150254-e9823e99808e //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/amartyasinha/lib-common/modules/common v0.0.0-20260416075014-135bf0acdeaf
```
gibizer: yepp this way we can get CI results. After the lib-common change lands we need to remove this line but that is a small price to pay to have early CI results

internal/controller/nova/nova_controller.go
```diff
@@ -56,7 +56,7 @@ import (
 
 	novav1 "github.com/openstack-k8s-operators/nova-operator/api/nova/v1beta1"
 	"github.com/openstack-k8s-operators/nova-operator/internal/nova"
-	"github.com/openstack-k8s-operators/nova-operator/internal/nova/api"
+	novaapi "github.com/openstack-k8s-operators/nova-operator/internal/nova/api"
```
gibizer, line 59: was this an automatic change from go fmt? I don't see the reason of it. But if this is automatic then we need to accept it.

gibizer (commented):
Overall looks good to me.

internal/controller/nova/nova_controller.go
amartyasinha: seems so! I tried removing that, but autosaving file again added that.

internal/controller/nova/nova_controller.go
amartyasinha: In lib-common, we do not check `InstanceType` (old logic) or `MultiTemplateDir` (new change) if template type is none.
Ref: https://github.com/amartyasinha/lib-common/blob/135bf0acdeafc623f9f9f855ca6850748664ff69/modules/common/util/template_util.go#L297

Will add a comment in there in code to clarify.

internal/controller/nova/nova_controller.go
gibizer: OK. Thanks for confirming my suspicion.

amartyasinha:
recheck nova-operator-kuttl

internal/controller/nova/novametadata_controller.go
```diff
@@ -898,7 +898,8 @@ func (r *NovaMetadataReconciler) generateNeutronConfigs(
 			Name:               configName,
 			Namespace:          instance.GetNamespace(),
 			Type:               util.TemplateTypeNone,
-			InstanceType:       getTemplateInstanceType(instance),
+			InstanceType:       instance.GetObjectKind().GroupVersionKind().Kind,
+			MultiTemplateDir:   "nova/metadata",
```
bogdando: it seems that other callers with TemplateTypeNone do NOT provide MultiTemplateDir.
Why this one is special?

Please evaluate it for consistency with lib-common behavior and other call sites

internal/controller/nova/common.go
```diff
@@ -533,20 +507,25 @@ func (r *ReconcilerBase) generateConfigsGeneric(
...
+	}
 
 	maps.Copy(extraTemplates, additionalTemplates)
 	cms := []util.Template{
 		{
 			Name:               configName,
 			Namespace:          instance.GetNamespace(),
 			Type:               util.TemplateTypeConfig,
-			InstanceType:       getTemplateInstanceType(instance),
+			InstanceType:       instance.GetObjectKind().GroupVersionKind().Kind,
```
bogdando, line 529: For a `NovaCell` instance, lib-common will look under a subdir derived from `"novacell"`, not `"nova/compute"`. Is that expected?

internal/controller/nova/novacell_controller.go
```diff
@@ -815,7 +815,7 @@ func (r *NovaCellReconciler) generateComputeConfigs(
 
 	configName := instance.GetName() + "-compute-config"
 	err := r.GenerateConfigs(
-		ctx, h, instance, configName, &hashes, templateParameters, extraData, cmLabels, map[string]string{},
+		ctx, h, instance, configName, &hashes, templateParameters, extraData, cmLabels, map[string]string{}, "nova/compute",
```
bogdando: ok, I see that this resolved the 'novacell' mismatching behavior I thought still takes place

internal/controller/nova/common.go
bogdando: looks about right, sorry

internal/controller/nova/novametadata_controller.go
amartyasinha: Thanks for the catch! removed `MultiTemplateDir` here as this specific case is of `TemplateTypeNone`.

internal/controller/nova/common.go
amartyasinha: np!

amartyasinha:
lib-common bump PR: https://github.com/openstack-k8s-operators/nova-operator/pull/1088

stuggi:
> lib-common bump PR: #1088

you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.

amartyasinha:
> > lib-common bump PR: #1088
> 
> you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.

The bump has happened in that PR. What's the procedure to get that merged?

stuggi:
> > > lib-common bump PR: #1088
> > 
> > 
> > you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.
> 
> The bump has happened in that PR. What's the procedure to get that merged?

the lib-common PR got merged, but the reference in the go.mod of the nova-operator was not bumped yet. you'd have to do that. https://github.com/openstack-k8s-operators/nova-operator/blob/main/go.mod#L15-L17

amartyasinha:
> > > > lib-common bump PR: #1088
> > > 
> > > 
> > > you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.
> > 
> > 
> > The bump has happened in that PR. What's the procedure to get that merged?
> 
> the lib-common PR got merged, but the reference in the go.mod of the nova-operator was not bumped yet. you'd have to do that. https://github.com/openstack-k8s-operators/nova-operator/blob/main/go.mod#L15-L17

Yes, get that. This automation PR [1] is bumping the lib-common changes in nova-operator. It now points to the latest commit id `81c71b3` of lib-common. So, I was wondering will you merge that PR or @gibizer ?

[1] https://github.com/openstack-k8s-operators/nova-operator/pull/1088/changes

stuggi:
> > > > > lib-common bump PR: #1088
> > > > 
> > > > 
> > > > you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.
> > > 
> > > 
> > > The bump has happened in that PR. What's the procedure to get that merged?
> > 
> > 
> > the lib-common PR got merged, but the reference in the go.mod of the nova-operator was not bumped yet. you'd have to do that. https://github.com/openstack-k8s-operators/nova-operator/blob/main/go.mod#L15-L17
> 
> Yes, get that. This automation PR [1] is bumping the lib-common changes in nova-operator. It now points to the latest commit id `81c71b3` of lib-common. So, I was wondering will you merge that PR or @gibizer ?
> 
> [1] https://github.com/openstack-k8s-operators/nova-operator/pull/1088/changes

yes, when tests passed, we can approve it

amartyasinha:
> > > > > > lib-common bump PR: #1088
> > > > > 
> > > > > 
> > > > > you'd have to bump lib-common in go.mod and api/go.mod in this pr, or wait for a next bump to happen.
> > > > 
> > > > 
> > > > The bump has happened in that PR. What's the procedure to get that merged?
> > > 
> > > 
> > > the lib-common PR got merged, but the reference in the go.mod of the nova-operator was not bumped yet. you'd have to do that. https://github.com/openstack-k8s-operators/nova-operator/blob/main/go.mod#L15-L17
> > 
> > 
> > Yes, get that. This automation PR [1] is bumping the lib-common changes in nova-operator. It now points to the latest commit id `81c71b3` of lib-common. So, I was wondering will you merge that PR or @gibizer ?
> > [1] https://github.com/openstack-k8s-operators/nova-operator/pull/1088/changes
> 
> yes, when tests passed, we can approve it

ack! thanks

amartyasinha:
Now placement controller files also use `MultiTemplateDir`.

gibizer (approved):
This is a refactoring to the new interface so I'm OK having no new test cases. The existing test coverage ensures we have the CMs and Secrets populated properly.

PR #1099, opened 2026-04-13
api: add openStackServiceIdentity to Nova service and template specs

amartyasinha:
Add optional OpenStackServiceIdentity (openstackService, serviceComponent) on NovaServiceBase and Nova/NovaCell *Template types for logical service/component identity.


amartyasinha:
recheck nova-operator-tempest-multinode

gibizer (changes requested):
I don't see the reason to make this API change.

amartyasinha:
> I don't see the reason to make this API change.

@gibizer In the initial draft, I did not make this API change. To get `nova/api` from `NovaAPI` value, I introduced `groupSegmentAndKindSuffix` function (check https://github.com/openstack-k8s-operators/lib-common/pull/682), which would work as `nova.openstack.org + NovaAPI → nova/api`. @bogdando discouraged to split like this, and suggested to introduced new data type. If you think we should go with other way, kindly suggest.

Once the lib-common gets the ability to handle multi-segment path, then we can go with https://github.com/openstack-k8s-operators/nova-operator/pull/1096/changes.

If we decide to proceed with adding these new fields in the specs, then the order would be:
1. Add new specs to nova-operator (this PR)
2. Modify lib-common ongoing PR to take care of multi-segment path based on new spec fields https://github.com/openstack-k8s-operators/lib-common/pull/682
3. Finally introduce MultiTemplateDir changes in nova-operator #1096

If we decide not to add new fields in the specs, then the order would be:
1. Find a way in lib-common to get the core service name and component's name (`nova` and `api` from `NovaAPI`). groupSegmentAndKindSuffix function in the PR https://github.com/openstack-k8s-operators/lib-common/pull/682 is currently doing that.
2. Finally introduce MultiTemplateDir changes in nova-operator https://github.com/openstack-k8s-operators/nova-operator/pull/1096

amartyasinha:
Closing as we do not want to modify the specs.

PR #1113, opened 2026-05-07
Add AGENTS.md

fmount:
This patch adds `AGENTS.md` to the repo to ensure any `AI agent` can follow the conventions defined for this operator. At the same time, it allows to not waste time navigating the repo to find initial information and save tokens.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,64 @@
+# AGENTS.md - nova-operator
+
+## Project overview
+
+nova-operator is a Kubernetes operator that manages
+[OpenStack Nova](https://docs.openstack.org/nova/latest/) and
+[OpenStack Placement](https://docs.openstack.org/placement/latest/)
+on OpenShift/Kubernetes. It is part of the
+[openstack-k8s-operators](https://github.com/openstack-k8s-operators) project.
+
+## Tech stack
+
+| Layer | Technology |
+|-------|------------|
+| Language | Go (modules, multi-module workspace via `go.work`) |
+| Scaffolding | [Kubebuilder v4](https://book.kubebuilder.io/) + [Operator SDK](https://sdk.operatorframework.io/) |
+| CRD generation | controller-gen (DeepCopy, CRDs, RBAC, webhooks) |
+| Config management | Kustomize |
+| Packaging | OLM bundle |
+| Testing | Ginkgo/Gomega + envtest (functional), KUTTL (integration) |
+| Linting | golangci-lint (`.golangci.yaml`) |
+| CI | Zuul (`zuul.d/`), Prow (`.ci-operator.yaml`), GitHub Actions |
+
+## Directory structure and Custom Resources
+
+The project directory structure and design choices are documented in
+[doc/design.md](doc/design.md).
+
+## Build commands
+
+After modifying Go code, always run: `make generate manifests fmt vet`.
+
+## Code style guidelines
+
+- Follow standard openstack-k8s-operators conventions and lib-common patterns.
+- Use `lib-common` modules for conditions, endpoints, TLS, storage, and other
+  cross-cutting concerns rather than re-implementing them.
+- CRD types are organized per service under `api/<service>/v1beta1/`
+  (e.g. `api/nova/v1beta1/`, `api/placement/v1beta1/`).
+- Controller logic goes in `internal/controller/<service>/`.
+  Resource-building helpers go in `internal/<service>/` packages.
+- Config templates are split per service under `templates/` (e.g.
+  `templates/nova/`, `templates/placement/`) and mounted at runtime via
+  the `OPERATOR_TEMPLATES` environment variable.
+- Webhook logic is split between the kubebuilder markers in `api/` and
+  the implementation in `internal/webhook/`.
+
+## Testing
+
+- Functional tests use the envtest framework with Ginkgo/Gomega and live in
+  `test/functional/`.
+- KUTTL integration tests live in `test/kuttl/`.
+- Run all functional tests: `make test`.
+- When adding a new field or feature, add corresponding test cases in
+  `test/functional/` and update fixture data accordingly.
+
+## Key dependencies
+
+- [lib-common](https://github.com/openstack-k8s-operators/lib-common): shared modules for conditions, endpoints, database, TLS, secrets, etc.
+- [infra-operator](https://github.com/openstack-k8s-operators/infra-operator): RabbitMQ and topology APIs.
+- [mariadb-operator](https://github.com/openstack-k8s-operators/mariadb-operator): database provisioning.
+- [keystone-operator](https://github.com/openstack-k8s-operators/keystone-operator): identity service registration.
+- [gophercloud](https://github.com/gophercloud/gophercloud): Go OpenStack SDK.
+- [Developer guide](https://github.com/openstack-k8s-operators/nova-operator/blob/main/doc/developer.md): local development setup, building, running, and debugging the operator.


AGENTS.md
```diff
@@ -0,0 +1,106 @@
...
+| `internal/webhook/` | Webhook implementation |
+| `templates/` | Config files and scripts mounted into pods via `OPERATOR_TEMPLATES` env var |
+| `config/crd,rbac,manager,webhook/` | Generated Kubernetes manifests (CRDs, RBAC, deployment, webhooks) |
+| `config/samples/` | Example CRs (Kustomize overlays). Includes `collapsed_cell`, `compute-fake`, `compute-ironic`, `upcall`, and TLS variants. |
+| `test/functional/` | envtest-based Ginkgo/Gomega tests |
+| `test/kuttl/` | KUTTL integration tests |
+| `hack/` | Helper scripts (CRD schema checker, local webhook runner) |
+| `doc/` | Design docs, developer guide, testing guide |
+| `ci/` | CI job definitions |
+
```
SeanMooney, line 28: Agents.md shoudl not contian info that can be easilly discoverd form introspecting the repo.

the directory structre woudl be ok in the readme but not here.

the CRs and tech stack shoudl also not be only defiend for the llm so i woudl add both of those to the readme as well.


i woudl perer to replace all of this with a pointer to

https://github.com/openstack-k8s-operators/nova-operator/blob/main/doc/design.md

you shoudl add the crd over view to that and you can add another doc
for the techstack beside it adn add a pointer to that as well

similar to what you dided with the key dependecies.

AGENTS.md
```diff
@@ -0,0 +1,106 @@
...
+  the implementation in `internal/webhook/`.
+
+## Testing
+
+- Functional tests use the envtest framework with Ginkgo/Gomega and live in
+  `test/functional/`.
+- KUTTL integration tests live in `test/kuttl/`.
+- Run all functional tests: `make test`.
+- When adding a new field or feature, add corresponding test cases in
+  `test/functional/` and update fixture data accordingly.
```
SeanMooney, line 55: this belongs in https://github.com/openstack-k8s-operators/nova-operator/blob/main/doc/developer.md with just a refence to that here.

AGENTS.md
```diff
@@ -0,0 +1,106 @@
...
+nova-operator is a Kubernetes operator that manages
+[OpenStack Nova](https://docs.openstack.org/nova/latest/) (the compute
+service: VM lifecycle, scheduling, cells, live migration, VNC console, and
+metadata) on OpenShift/Kubernetes. It is part of the
+[openstack-k8s-operators](https://github.com/openstack-k8s-operators) project.
+
+Key Nova domain concepts: **cells** (cell0, cell1+), **conductors**,
+**scheduler** (placement-based), **compute agents**, **VNC proxy**
+(noVNCProxy), **metadata service**, **live migration**, **evacuation**,
+**DB purge**, **upcall**.
```
SeanMooney: while this is perhaps relevant for nova its not really relevant for an operator that deploys nova

SeanMooney (dismissed):
my meta comment is much of this info shoudl live in the project docs and only high signal to noise ratio instruction shoudl be in the agents.md

the majority of agents.md shoudl effectivly be a routing layer like
https://llmstxt.org/

instead of telling it what CRDs exist
tell it where to find there defintions

instead of embdeing detailed developer practices
tell it where to discover them in a doc in the repo that is writeen for humans when it needs them

AGENTS.md
fmount: I will remove 11-14

AGENTS.md
fmount: true, but that would be referenced for implementation guidelines, while this should give just the idea of the repo to reduce the initial amount of grep/find.

fmount: I will add to the references.

AGENTS.md
fmount: the idea of this file is to have some info to avoid initial grep/find to discover the basic structure that hopefully is not changing much (or changing at all at this point of the lifecycle of the operator). +1 to link `design.md` since this is internal knowledge for the nova-operator.
If you like this part to go away in favor of design.md this is fine by me.

fmount: The other things I wanted to avoid here are:
1. don't start with curl remote content immediately
2. don't start grepping or trying to find files to build the initial understanding of the project.

For this reason reference to other files within the project are probably good, and stay <150 lines helps you to give an overview of the basic things.

Note that this is an initial template. I expect each team to customize/optimize depending on their experience and experiments with AI agents.

SeanMooney: i would prefer to have it in design.md as much as possibel

and yes i actually would prefer to have no links to external soruce in the agents.md and have those behind 1 layer of indirection to avoid the curlign issue too.

in general the local discover is relatively cheap and in general it not required unless you are impresices with your prompts.

having a basic overview is ok but ideally not in the AGENTS.md as its very easy for ti to get out of sync

we are in the process or movign the palcement operator into this repo and then adding cyborg too sot eh directory stucrue is going to change alotner in the short term

that why im hesitant to encode it here because its not currently stable.

fmount: > i would prefer to have it in design.md as much as possibel

+1 and it is a good idea to defer to that layer more knowledge. I think I will apply the same approach across the board in a follow up.
> 
> and yes i actually would prefer to have no links to external soruce in the agents.md and have those behind 1 layer of indirection to avoid the curlign issue too.

+1 agreed
> 
> in general the local discover is relatively cheap and in general it not required unless you are impresices with your prompts.
> 
> having a basic overview is ok but ideally not in the AGENTS.md as its very easy for ti to get out of sync
> 
> we are in the process or movign the palcement operator into this repo and then adding cyborg too sot eh directory stucrue is going to change alotner in the short term
> 
> that why im hesitant to encode it here because its not currently stable.

ack and this is why I removed the entire directory tree section that can be discovered. I also pointed to docs/design.md so we defer to that layer most of the guidelines.

SeanMooney (commented):
/lgtm 

i think we can use this as a base and refine it but i want gibi or kamil ot approve

AGENTS.md
```diff
@@ -0,0 +1,62 @@
+# AGENTS.md - nova-operator
+
+## Project overview
+
+nova-operator is a Kubernetes operator that manages
+[OpenStack Nova](https://docs.openstack.org/nova/latest/) (the compute
+service: VM lifecycle, scheduling, cells, live migration, VNC console, and
```
mrkisaolamb: Unfortunately, this is outdated because placement has now been merged into nova-operator. As a result, nova-operator is now more like an infrastructure operator containing multiple services and APIs, so this part need to be updated

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+| Config management | Kustomize |
+| Packaging | OLM bundle |
+| Testing | Ginkgo/Gomega + envtest (functional), KUTTL (integration) |
+| Linting | golangci-lint (`.golangci.yaml`) |
+| CI | Zuul (`zuul.d/`), Prow (`.ci-operator.yaml`), GitHub Actions |
+
+## Directory structure and Custom Resources
+
+The project directory structure and design choices are documented in
+[doc/design.md](doc/design.md).
```
mrkisaolamb, line 27: his isn't part of this PR, but a note for myself: our design.md is outdated

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Code style guidelines
+
+- Follow standard openstack-k8s-operators conventions and lib-common patterns.
+- Use `lib-common` modules for conditions, endpoints, TLS, storage, and other
+  cross-cutting concerns rather than re-implementing them.
+- CRD types go in `api/v1beta1/`. Controller logic goes in
+  `internal/controller/`. Resource-building helpers go in `internal/nova*`
+  packages matching the CR they support.
+- Config templates are plain files in `templates/` -- they are mounted at
```
mrkisaolamb: I'm not sure, but it might be worth mentioning that the templates are split between nova and placement.

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+## Build commands
+
+After modifying Go code, always run: `make generate manifests fmt vet`.
+
+## Code style guidelines
+
+- Follow standard openstack-k8s-operators conventions and lib-common patterns.
+- Use `lib-common` modules for conditions, endpoints, TLS, storage, and other
+  cross-cutting concerns rather than re-implementing them.
+- CRD types go in `api/v1beta1/`. Controller logic goes in
```
mrkisaolamb: because of the Placement merge, this struct needs to be updated to something like: api/<nova/placement>/v1beta1/

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+  `internal/controller/`. Resource-building helpers go in `internal/nova*`
+  packages matching the CR they support.
+- Config templates are plain files in `templates/` -- they are mounted at
+  runtime via the `OPERATOR_TEMPLATES` environment variable.
+- Webhook logic is split between the kubebuilder markers in `api/v1beta1/` and
```
mrkisaolamb: path are now different

mrkisaolamb (changes requested):
Thanks @fmount! Unfortunately, because of the placement merge, we need to update this agent.md

AGENTS.md
fmount: Makes sense, let me update it in the next PS.

AGENTS.md
fmount: sure, updating this part as well.

AGENTS.md
fmount: it's a good idea.

AGENTS.md
fmount: Done

AGENTS.md
fmount: Done

AGENTS.md
fmount: Done

mrkisaolamb (approved):
lgtm

PR #1118, opened 2026-05-18
Add missing configs related to placementAPI

amartyasinha:
default image for placement and rbac for placementapi are added through this commit.


stuggi (approved):
/lgtm

PR #1119, opened 2026-05-18
Consume common templates via CommonTemplates field

fmount:
The `ssl.conf` `httpd` configuration is duplicated across service operators.
This patch removes the local copy and consumes it from [lib-common](https://github.com/openstack-k8s-operators/lib-common/pull/694).

Jira: [OSPRH-30239](https://redhat.atlassian.net/browse/OSPRH-30239)


stuggi (approved):
/lgtm

SeanMooney:
sharing the ssl.conf may make sense but the httpd config tends to be at least partly service specific

nova has 2 rest apis for example and neither can run with multiple threads they have to be scaled by process or replicas. the most obvious delta is the path to the wsgi module is service dependent 

This commit seam ok to me but also i don't really see the maintenance of these files being worth the added complexity of now needing the go to lib common to understand how that is deployed.

fmount:
> sharing the ssl.conf may make sense but the httpd config tends to be at least partly service specific
> 
> nova has 2 rest apis for example and neither can run with multiple threads they have to be scaled by process or replicas. the most obvious delta is the path to the wsgi module is service dependent
> 
> This commit seam ok to me but also i don't really see the maintenance of these files being worth the added complexity of now needing the go to lib common to understand how that is deployed.

Thanks for the comment, @SeanMooney.

Deduplicating common configurations (e.g., `ssl.conf` in this case) is part of a broader effort to maintain consistency across the board. This approach significantly streamlines maintenance by allowing us to update shared components with a single PR, rather than creating ~20 individual patches and the associated bureaucracy.
Additionally, using `CommonTemplates []string` allows you to selectively pull the templates you need from a shared location. Ultimately, it gives the operator the flexibility to either use the common template or define a custom one directly within the service operator.
I agree that `httpd.conf` does not fit how nova generates the config, but in the storage operators area I plan to deduplicate this file as well and convert it into a template. This makes sense because in that case it includes the specific `VHost` (which is really the custom part) which resides in the service operator's `templates/` directory.

PR #1121, opened 2026-05-21
Make rabbitmq user and vhost name shorter in kuttl tests

amoralej:
Some recent change (maybe [1a)], the string used in infra-operator for rabbitmquser creation has changed from
`nova-kuttl-notification-transport-nova-notifications-user` to `rabbitmq-notifications-nova-notifications-user-nova-notifications` which is longer that 63 char openshift limit, leading to error:

RabbitMQ user error occurred Secret "rabbitmq-user-rabbitmq-notifications-nova-notifications-user-nova-notifications" is invalid: metadata.labels: Invalid value: "rabbitmq-notifications-nova-notifications-user-nova-notifications": must be no more than 63 characters

While this may be understood as a bug, this patch is making vhost and user names shorter to unblock the kuttl test.

[1] https://github.com/openstack-k8s-operators/infra-operator/commit/fcc8556324f5a6b5e24031a2be97c17cbc1c4baa#diff-627c3e6f14aec9dd16405a764cf2e60199490e15a399bd3df182df26f00e999d

### DIFF
--- a/test/kuttl/test-suites/default/rmquser-vhost/01-assert.yaml
+++ b/test/kuttl/test-suites/default/rmquser-vhost/01-assert.yaml
@@ -14,8 +14,8 @@ metadata:
   name: nova-kuttl-notification-transport
 spec:
   rabbitmqClusterName: rabbitmq-notifications
-  username: nova-notifications
-  vhost: nova-notifications
+  username: nova-notif
+  vhost: nova-notif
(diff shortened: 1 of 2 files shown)


amoralej:
fixed in infra-operator https://github.com/openstack-k8s-operators/infra-operator/pull/583

PR #1125, opened 2026-06-05
[DNM] Move placement kuttl tests under default kuttl namespace

amartyasinha:
Currently, zuul only runs default kuttl tests, thus placement kuttl tests were not getting validated. This commit ensure they are also tested.


test/kuttl/test-suites/default/deps/namespace.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: nova-kuttl-multi-cell
+  name: nova-kuttl-default
```
SeanMooney, line 4: so each test suite is ment ot have its onw namespace
and nova and palcmenet shoudl not share oen.

test/kuttl/test-suites/default/placement-tests/common/assert_sample_deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   secret: osp-secret
   serviceUser: placement
 status:
-  databaseHostname: openstack.placement-kuttl-tests.svc
+  databaseHostname: openstack.nova-kuttl-default.svc
```
SeanMooney, line 20: so this is incorrect

the placement test are a diffent test quite and shoudl eb using ther own nameapces.

you shoudl be able to run the palcemetn test and the nova tests at the same time without them interfering with each other

amartyasinha: Thanks @SeanMooney! I can try looking how to enable both tests run in zuul instead of consolidating them in single namespace. The reason I was trying to move placement tests under default namespace was because the cyborg tests are also defined under `kuttl/test-suites/default`. But yeah, I understand the placement tests are different than nova/cyborg.

amartyasinha:
Closing for now. Will check soon how to run both kuttl tests in zuul.

PR #1129, opened 2026-06-15
Add max_retries and retry_interval to [(api_)database] conf

melwitt:
Nova's default database connection retry settings (max_retries=10, retry_interval=10) mean nova-api, for example, can spend up to 100 seconds retrying a connection to an unavailable database. This exceeds the Kubernetes startup probe window (60 seconds), so the pod gets killed before Nova finishes retrying, leading to a CrashLoopBackOff when a cell database is down.

In RHOSO, Kubernetes provides its own higher-level retry mechanism by killing and recreating pods that fail to start. This is preferable to Nova retrying internally because it reports the situation clearly via CR status fields and events, and allows Kubernetes to reschedule the pod to another worker if needed.

Set max_retries to 3 seconds and retry_interval to 1 second for both [database] and [api_database] so that Nova gives up on an unreachable database quickly and lets Kubernetes handle the recovery.

For nova-conductor services, set max_retries to 30 seconds and retry_interval to 1 second to enable them to be more tolerant of temporary database connection disruptions.

Resolves: [OSPRH-30130](https://redhat.atlassian.net/browse/OSPRH-30130)

### DIFF
--- a/templates/nova/nova.conf
+++ b/templates/nova/nova.conf
@@ -241,12 +241,24 @@ cpu_power_management=false
 
 {{if (index . "cell_db_address")}}
 [database]
+{{ if eq .service_name "nova-conductor"}}
+max_retries = 30
+{{ else }}
+max_retries = 3
+{{ end }}
+retry_interval = 1
 connection = mysql+pymysql://{{ .cell_db_user }}:{{ .cell_db_password}}@{{ .cell_db_address }}/{{ .cell_db_name }}?read_default_file=/etc/my.cnf
 {{end}}
 
 
 {{if (index . "api_db_address")}}
 [api_database]
+{{ if eq .service_name "nova-conductor"}}
+max_retries = 30
+{{ else }}
+max_retries = 3
+{{ end }}
+retry_interval = 1
 connection = mysql+pymysql://{{ .api_db_user }}:{{ .api_db_password }}@{{ .api_db_address }}/{{ .api_db_name }}?read_default_file=/etc/my.cnf
 {{end}}
 
--- a/test/functional/nova/multicell_test.go
+++ b/test/functional/nova/multicell_test.go
@@ -156,7 +156,7 @@ var _ = Describe("Nova multi cell", func() {
 			Expect(configData).To(
 				ContainSubstring(
 					fmt.Sprintf(
-						"[database]\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_cell0?read_default_file=/etc/my.cnf",
+						"[database]\nmax_retries = 30\nretry_interval = 1\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_cell0?read_default_file=/etc/my.cnf",
 						cell0Account.Spec.UserName, cell0Secret.Data[mariadbv1.DatabasePasswordSelector],
 						cell0.MariaDBDatabaseName.Name, novaNames.Namespace)),
 			)
@@ -167,7 +167,7 @@ var _ = Describe("Nova multi cell", func() {
 			Expect(configData).To(
 				ContainSubstring(
 					fmt.Sprintf(
-						"[api_database]\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_api?read_default_file=/etc/my.cnf",
+						"[api_database]\nmax_retries = 30\nretry_interval = 1\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_api?read_default_file=/etc/my.cnf",
 						apiAccount.Spec.UserName, apiSecret.Data[mariadbv1.DatabasePasswordSelector],
 						novaNames.APIMariaDBDatabaseName.Name, novaNames.Namespace)),
 			)
@@ -229,7 +229,7 @@ var _ = Describe("Nova multi cell", func() {
 			Expect(configData).To(
 				ContainSubstring(
 					fmt.Sprintf(
-						"[database]\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_cell0?read_default_file=/etc/my.cnf",
+						"[database]\nmax_retries = 3\nretry_interval = 1\nconnection = mysql+pymysql://%s:%s@hostname-for-%s.%s.svc/nova_cell0?read_default_file=/etc/my.cnf",
 						cell0Account.Spec.UserName, cell0Secret.Data[mariadbv1.DatabasePasswordSelector],
 						cell0.MariaDBDatabaseName.Name, novaNames.Namespace)),
 			)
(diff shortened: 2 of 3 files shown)


gibizer:
> Nova's default database connection retry settings (max_retries=-1, retry_interval=10) cause nova-api pods to hang indefinitely when a cell database is unavailable. This prevents the pods from becoming ready within the Kubernetes startup probe window (60 seconds), leading to a CrashLoopBackOff.

I tis not -1 it is 10 by default:
https://docs.openstack.org/nova/latest/configuration/config.html#api_database.max_retries
https://docs.openstack.org/nova/latest/configuration/config.html#database.max_retries

Still enough to slow down startup and trigger the failure of the startup k8s probe that will trigger the kill of the pod.

gibizer (changes requested):
Please fix the commit message, otherwise looks good to me.

gibizer (approved):
looks good. Thank for fixing

gibizer:
/retest
```
    - '2026-06-16 12:03:36,198 - repo-setup-get-hash - ERROR - Invalid response received
      from the delorean server. Queried URL: https://trunk.rdoproject.org/centos9-antelope/current-podified/delorean.repo.md5.
      Response code: -1. Response text: The read operation timed out. Failed to create
      HashInfo object.'
```

gibizer:
recheck

gibizer (changes requested):
The functional test failure is relevant. You can reproduce it via `make test` locally. We have some tests that are too specific about the content of the generated content. e.g. https://github.com/openstack-k8s-operators/nova-operator/blob/7ba93ca4794341b745f11e6adfcc867f5159a569/test/functional/nova/multicell_test.go#L232

templates/nova/nova.conf
```diff
@@ -241,12 +241,16 @@ cpu_power_management=false
 
 {{if (index . "cell_db_address")}}
 [database]
+max_retries = 3
```
SeanMooney: https://docs.openstack.org/nova/latest/configuration/config.html#api_database.max_retries
https://docs.openstack.org/nova/latest/configuration/config.html#api_database.retry_interval

so im not actully sure that relying on k8s retry is actully preferable

the defualt is 10
im not sure turing this to 3 relly helps with stablity in gerneal in ci.

i woudl suggest we increase max_reties to say 30
but lower retry_interval to 1 as your proposing

that keeps the time less the the k8s interval but should be more stable over all.

with that said gibi is right that we will need ot update

https://github.com/openstack-k8s-operators/nova-operator/blob/7ba93ca4794341b745f11e6adfcc867f5159a569/test/functional/nova/multicell_test.go#L230-L246

the simple solution woul db to put this after the existing assert but we shoudl jsut add the new content to them as well so that we dont reguress this overried

melwitt: Fair point ... otherwise it would give up after only 3 seconds. I have updated to `max_retries = 30`.

SeanMooney (approved):
thanks this looks good form my perspective.
as an asside we shoudl get you added to the opnestack-k8s-operators org so that your contibutos are allowed to be tested directly

templates/nova/nova.conf
gibizer: > so im not actully sure that relying on k8s retry is actully preferable

Why?

> im not sure turing this to 3 relly helps with stablity in gerneal in ci.

We have to balance CI stability (I haven't seen this being unstable so far) and the actual performance degradation when a cell DB is down and therefore each scatter-gather taking excessive time due to nova retrying the DB connection to that cell DB for 30 seconds.

SeanMooney: my concern i smore with the conductor then with the api.

these setting are common to both and we dont really want the conducer restarting because of a tempery db disconnect wheile a pod is been updated

so i dont think tis a greate idea to only allow a 3 second interval for the db to come back in production ignoreing the ci.

SeanMooney: just to be clear my concer about stablity is more that
if a conenction drops due to a reconfiation of the contole plane that bounces the db pod i dont wnat the conductor or novnc proxy to bail after 3 seconds if it take longer for the db to start back up.

for the api that more reasonable
perhaps we want to make this service specific?

use 3 seconds for the api adn 30 for the rest?
woudl that be a better balance?

that pretty easy to do in the template via an if condtion on teh service type

gibizer: Hm. I understood your concern but I'm still wondering a bit here.

If the system is up and running and then the cell DB is going down due to container image change (update) while the cell conductor is handling traffic probably causes those requests to fail, but does not cause the conductor pod to crash. So the blast radius is smaller. Sure if we loose a DB update from the compute due to the conductor is giving up too early on retrying on the DB can cause DB inconsistencies as I don't think the compute will retry the DB update ever.

In the other hand production deployments should have DBs with multiple replicas and k8s should roll those replicas during update so the cell conductor only need to reconnect during DB server update but never loose connection to the DB. Sure this can be more complicated with Galera in the picture. So meh. 

Both API and Scheduler performance is affected by the slowness of scatter-gather if one of the cell DBs are down. But you are correct cell conductors are not affected by that. So for cell conductors I'm OK to bump the retries higher as you suggested to be avoid failing DB updates from the computes during DB server updates. 
But lets keep the scheduler and the API retries low if possible. 

I'm not sure about the super conductor (cell0 conductor in the installer reality), I guess that conductor can do scatter-gather technically but I'm not sure when it would do it. ... OK for simplicity lets keep the default conductor config the same across the board.

tl;dr: 
* keep API and Scheduler default retries low due to performance
* have conductor default retries higher for DB consistency reasons.

How does it sound?

melwitt: I think this sounds a reasonable compromise -- I will make the change to make it conditional depending on these services.

SeanMooney: yep that works for me too.

SeanMooney (approved):
if gibi is happy with this version im fine with proceedign with this
thanks melaine for the quick turnaround

melwitt:
Looks like `nova-operator-tempest-multinode` failed with:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: Value of unknown type: <class 'urllib3.exceptions.NewConnectionError'>, <urllib3.connection.HTTPSConnection object at 0x7f250bd2a6d0>: Failed to establish a new connection: [Errno 111] Connection refused
task path: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/ci_local_storage/tasks/main.yml:23
fatal: [localhost]: FAILED! => 
    changed: false
    module_stderr: |
      Traceback (most recent call last):
        File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 169, in _new_conn
          conn = connection.create_connection(
        File "/usr/lib/python3.9/site-packages/urllib3/util/connection.py", line 96, in create_connection
          raise err
        File "/usr/lib/python3.9/site-packages/urllib3/util/connection.py", line 86, in create_connection
          sock.connect(sa)
      ConnectionRefusedError: [Errno 111] Connection refused
```

gibizer (approved):
Looks good to me. Thanks folks!

gibizer:
recheck infra issue

gibizer:
\o/ Thanks @melwitt to push this through

PR #1130, opened 2026-06-18
Unify common code

amartyasinha:
Introduce internal/common for shared nova/placement
  controller code: reconciler base, ensure helpers,
  config generation, watch mappers, Kolla constants,
  and static errors. Nova and placement controllers
  call these helpers; nova-only logic stays in
  internal/controller/nova/common.go.

  This is a move-first refactor: logic, comments, and
  naming are preserved where possible. Only
  unavoidable mechanical changes are included (e.g.
  exported names for cross-package use, import
  updates, r.Client.Update because Client is a named
  field on ReconcilerBase).

  Co-authored-by: Cursor <cursoragent@cursor.com>

To make CI pass
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/4039

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,10 @@ test: manifests generate fmt vet envtest ginkgo ## Run tests.
 	OPERATOR_TEMPLATES="$(PWD)/templates" \
 	$(GINKGO) --trace --cover --coverpkg=../../internal/...,../../api/nova/v1beta1,../../api/placement/v1beta1 --coverprofile cover.out --covermode=atomic --randomize-all ${PROC_CMD} $(GINKGO_ARGS) ./test/...
 
+.PHONY: gotest-unit
+gotest-unit: ## Run unit tests under test/unit/ (also run via ginkgo in the test target).
+	go test -v ./test/unit/...
+
 ##@ Build
 
 .PHONY: build
--- a/internal/common/errors.go
+++ b/internal/common/errors.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import "errors"
+
+// Static errors shared across operator services.
+var (
+	// ErrACSecretMissingKeys indicates that the ApplicationCredential secret is missing required keys.
+	ErrACSecretMissingKeys = errors.New("ApplicationCredential secret missing required keys")
+	// ErrTemplateDirUnset indicates that no template directory was provided.
+	ErrTemplateDirUnset = errors.New("templateDir must be set")
+
+	// ErrUnexpectedObjectType is returned when a webhook receives an unexpected object type.
+	ErrUnexpectedObjectType = errors.New("unexpected object type")
+)
--- a/internal/common/network.go
+++ b/internal/common/network.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	nad "github.com/openstack-k8s-operators/lib-common/modules/common/networkattachment"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// EnsureNetworkAttachments - checks the requested network attachments exists and
+// returns the annotation to be set on the deployment objects.
+func EnsureNetworkAttachments(
+	ctx context.Context,
+	h *helper.Helper,
+	networkAttachments []string,
+	conditionUpdater ConditionUpdater,
+	requeueTimeout time.Duration,
+) (map[string]string, ctrl.Result, error) {
+	var nadAnnotations map[string]string
+
+	// networks to attach to
+	nadList := []networkv1.NetworkAttachmentDefinition{}
+	for _, netAtt := range networkAttachments {
+		nadObject, err := nad.GetNADWithName(ctx, h, netAtt, h.GetBeforeObject().GetNamespace())
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				// Since the net-attach-def CR should have been manually created by the user and referenced in the spec,
+				// we treat this as a warning because it means that the service will not be able to start.
+				log.FromContext(ctx).Info(fmt.Sprintf("network-attachment-definition %s not found", netAtt))
+				conditionUpdater.Set(condition.FalseCondition(
+					condition.NetworkAttachmentsReadyCondition,
+					condition.ErrorReason,
+					condition.SeverityWarning,
+					condition.NetworkAttachmentsReadyWaitingMessage,
+					netAtt))
+				return nadAnnotations, ctrl.Result{RequeueAfter: requeueTimeout}, nil
+			}
+			conditionUpdater.Set(condition.FalseCondition(
+				condition.NetworkAttachmentsReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.NetworkAttachmentsErrorMessage,
+				err.Error()))
+			return nadAnnotations, ctrl.Result{}, err
+		}
+
+		if nadObject != nil {
+			nadList = append(nadList, *nadObject)
+		}
+	}
+
+	nadAnnotations, err := nad.EnsureNetworksAnnotation(nadList)
+	if err != nil {
+		return nadAnnotations, ctrl.Result{}, fmt.Errorf("failed create network annotation from %s: %w",
+			networkAttachments, err)
+	}
+
+	return nadAnnotations, ctrl.Result{}, nil
+}
--- a/internal/common/topology.go
+++ b/internal/common/topology.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import (
+	"context"
+	"fmt"
+
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// topologyHandler provides topology reference state for a reconciled object.
+type topologyHandler interface {
+	GetSpecTopologyRef() *topologyv1.TopoRef
+	GetLastAppliedTopology() *topologyv1.TopoRef
+	SetLastAppliedTopology(t *topologyv1.TopoRef)
+}
+
+// EnsureTopology - when a Topology CR is referenced, remove the
+// finalizer from a previous referenced Topology (if any), and retrieve the
+// newly referenced topology object
+func EnsureTopology(
+	ctx context.Context,
+	helper *helper.Helper,
+	instance topologyHandler,
+	finalizer string,
+	conditionUpdater ConditionUpdater,
+	defaultLabelSelector metav1.LabelSelector,
+) (*topologyv1.Topology, error) {
+	topology, err := topologyv1.EnsureServiceTopology(
+		ctx,
+		helper,
+		instance.GetSpecTopologyRef(),
+		instance.GetLastAppliedTopology(),
+		finalizer,
+		defaultLabelSelector,
+	)
+	if err != nil {
+		conditionUpdater.Set(condition.FalseCondition(
+			condition.TopologyReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.TopologyReadyErrorMessage,
+			err.Error()))
+		return nil, fmt.Errorf("waiting for Topology requirements: %w", err)
+	}
+	// update the Status with the last retrieved Topology (or set it to nil)
+	tr := instance.GetSpecTopologyRef()
+	instance.SetLastAppliedTopology(tr)
+	// update the Topology condition only when a Topology is referenced and has
+	// been retrieved (err == nil)
+	if tr != nil {
+		// update the TopologyRef associated condition
+		conditionUpdater.MarkTrue(
+			condition.TopologyReadyCondition,
+			condition.TopologyReadyMessage,
+		)
+	}
+	return topology, nil
+}
--- a/internal/common/watch.go
+++ b/internal/common/watch.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/openstack-k8s-operators/lib-common/modules/common"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// FindObjectsForSrcByField returns reconcile requests for CRs in src's namespace
+// that reference src via one of the indexed watchFields.
+func FindObjectsForSrcByField(
+	ctx context.Context,
+	log logr.Logger,
+	reader client.Reader,
+	src client.Object,
+	watchFields []string,
+	newList func() client.ObjectList,
+) []reconcile.Request {
+	var requests []reconcile.Request
+
+	for _, field := range watchFields {
+		crList := newList()
+		listOps := &client.ListOptions{
+			FieldSelector: fields.OneTermEqualSelector(field, src.GetName()),
+			Namespace:     src.GetNamespace(),
+		}
+		err := reader.List(ctx, crList, listOps)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("listing %s for field: %s - %s", crList.GetObjectKind().GroupVersionKind().Kind, field, src.GetNamespace()))
+			return requests
+		}
+
+		items, err := meta.ExtractList(crList)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("extracting items from %s", crList.GetObjectKind().GroupVersionKind().Kind))
+			return requests
+		}
+		requests = appendRequestsForObjects(log, src, requests, items)
+	}
+
+	return requests
+}
+
+// FindObjectsForSrcInNamespace returns reconcile requests for all CRs in src's namespace.
+func FindObjectsForSrcInNamespace(
+	ctx context.Context,
+	log logr.Logger,
+	reader client.Reader,
+	src client.Object,
+	newList func() client.ObjectList,
+) []reconcile.Request {
+	crList := newList()
+	listOps := &client.ListOptions{
+		Namespace: src.GetNamespace(),
+	}
+	err := reader.List(ctx, crList, listOps)
+	if err != nil {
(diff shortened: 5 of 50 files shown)


internal/controller/nova/common.go
```diff
@@ -282,132 +222,6 @@ func allSubConditionIsTrue(conditionsGetter conditionsGetter) bool {
...
-	if err != nil {
-		return nadAnnotations, ctrl.Result{}, fmt.Errorf("failed create network annotation from %s: %w",
-			networkAttachments, err)
-	}
-
-	return nadAnnotations, ctrl.Result{}, nil
-}
-
 // ReconcilerBase provides a common set of clients scheme and loggers for all reconcilers.
 type ReconcilerBase struct {
```
SeanMooney, line 205: so ReconcilerBase is intened to be the thing that hold the common code

so you shoudl be moving this as well if your going to do this.

the orginal inte was for this to become the common bas for the nova placment and evnetully cyborg contolers.

amartyasinha: Thanks @SeanMooney! Sure, I'll be moving `ReconcilerBase` soon to common code. I had it in the initial draft, but removed it because it was getting overwhelming for me to self-review. Once I believe the current code I'm moving are well-placed, I'll move `ReconcilerBase` as well.

amartyasinha: I've tried moving `ReconcilerBase` and other related func to internal/common. PTAL.

internal/controller/placement/api_controller.go
```diff
@@ -307,34 +217,26 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	hash, result, secret, err := internalcommon.EnsureSecret(
 		ctx,
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
 		[]string{
 			instance.Spec.PasswordSelectors.Service,
 		},
 		h.GetClient(),
-		&instance.Status.Conditions)
+		&instance.Status.Conditions,
+		time.Second*10,
```
mrkisaolamb: I think thanks to unifying ReconcilerBase, we can use r.RequeueTimeout

internal/controller/placement/api_controller.go
```diff
@@ -307,34 +217,26 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	hash, result, secret, err := ensureSecret(
+	hash, result, secret, err := internalcommon.EnsureSecret(
```
mrkisaolamb, line 223: Because we unified this method, it's a little bit different. It would be nice to leave a comment about that, and maybe add some tests for the placement case (for condition msg, returned error, requeue etc)

internal/common/network.go
```diff
@@ -0,0 +1,82 @@
...
+	networkAttachments []string,
+	conditionUpdater ConditionUpdater,
+	requeueTimeout time.Duration,
+) (map[string]string, ctrl.Result, error) {
+	var nadAnnotations map[string]string
+
+	// networks to attach to
+	nadList := []networkv1.NetworkAttachmentDefinition{}
+	for _, netAtt := range networkAttachments {
+		nad, err := nad.GetNADWithName(ctx, h, netAtt, h.GetBeforeObject().GetNamespace())
```
mrkisaolamb: maybe nadObject (nad is a name of package already)

internal/common/errors.go
```diff
@@ -0,0 +1,29 @@
...
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import "errors"
+
+// Static errors shared across operator services.
+var (
+	// ErrACSecretNotFound indicates that the ApplicationCredential secret was not found.
+	ErrACSecretNotFound = errors.New("ApplicationCredential secret not found")
+	// ErrACSecretMissingKeys indicates that the ApplicationCredential secret is missing required keys.
```
mrkisaolamb, line 23: I think we missed a few errors, like errUnexpectedObjectType or ErrInvalidObjectType (of course, assuming we want to deduplicate all similar errors in the future, e.g., all errors like errExpectedNova<name>Object)

mrkisaolamb (changes requested):

internal/common/errors.go
amartyasinha: moved all remaining errors here.

internal/common/network.go
amartyasinha: Done. Thanks!

internal/controller/placement/api_controller.go
amartyasinha: done

internal/controller/placement/api_controller.go
amartyasinha: added comment + tests

internal/controller/placement/api_controller.go
mrkisaolamb: I think this can be fix Log := r.GetLogger(ctx)

mrkisaolamb: can we fix all this kind of timeouts  using r.RequeueTimeout?

internal/common/errors.go
mrkisaolamb: Cool thanks, now when we have all errors in one place they can be unified like ErrInvalidObjectType and ErrUnexpectedObjectType probably can be one error in code. Maybe similar thing can be with more specific one eg ErrExpectedNovaAPIObject

internal/controller/placement/api_controller.go
mrkisaolamb, line 280: Do you think we can generalize the existing nova generateConfigsGeneric and use it here?

mrkisaolamb (changes requested):
Something else we can try: unifying FindObjectsForSrcByField and findObjectsWithAppSelectorLabelInNamespace

internal/common/errors.go
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+package common //nolint:revive // common is the established package name for multi-group shared code
+
+import "errors"
+
+// Static errors shared across operator services.
+var (
+	// ErrACSecretNotFound indicates that the ApplicationCredential secret was not found.
+	ErrACSecretNotFound = errors.New("ApplicationCredential secret not found")
```
mrkisaolamb: question: wdyt about removing this error and use in placement similar condition+log+nil error like in nova controllers?

internal/common/errors.go
amartyasinha: done. PTAL

internal/controller/placement/api_controller.go
amartyasinha: done

amartyasinha: updated

amartyasinha: Yes, we can do that. Though it won't be a clean move of nova's `generateConfigsGeneric` to `internal/common`, and there will be a few changes, thus will required some tests. I'll try to do it tomorrow, or if you want, we can have it in a follow up PR too, whatever you prefer.

amartyasinha: added a new commit for this. PTAL

internal/controller/nova/common.go
```diff
@@ -269,320 +201,54 @@ func cleanNovaServiceFromNovaDb(
...
-	commonTemplates []string,
-	templateDir string,
-) error {
-	return r.generateConfigsGeneric(
-		ctx, h, instance, nova.GetServiceConfigSecretName(instance.GetName()),
-		envVars, templateParameters, extraData,
-		cmLabels, additionalTemplates, commonTemplates, templateDir, true,
-	)
+// NovaAdditionalTemplates are the default extra config templates for nova services.
+var NovaAdditionalTemplates = map[string]string{
```
mrkisaolamb: I don't think this should be global, as it can be mutated by any caller during execution. It's better to scope this locally inside the functions

internal/common/network.go
```diff
@@ -0,0 +1,82 @@
...
+	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	nad "github.com/openstack-k8s-operators/lib-common/modules/common/networkattachment"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// EnsureNetworkAttachments - checks the requested network attachments exists and
+// returns the annotation to be set on the deployment objects.
+func EnsureNetworkAttachments(
```
mrkisaolamb, line 35: can we add some tests for this?

internal/common/topology.go
```diff
@@ -0,0 +1,77 @@
...
+type topologyHandler interface {
+	GetSpecTopologyRef() *topologyv1.TopoRef
+	GetLastAppliedTopology() *topologyv1.TopoRef
+	SetLastAppliedTopology(t *topologyv1.TopoRef)
+}
+
+// EnsureTopology - when a Topology CR is referenced, remove the
+// finalizer from a previous referenced Topology (if any), and retrieve the
+// newly referenced topology object
+func EnsureTopology(
```
mrkisaolamb, line 39: pls add some unit tests

mrkisaolamb (changes requested):

internal/controller/nova/common.go
amartyasinha: Wrapped inside a local function `func novaAdditionalTemplates() map[string]string {}` which returns the map now. Can't be mutated by caller during execution.

internal/common/network.go
amartyasinha: added

internal/common/topology.go
amartyasinha: added

internal/common/errors.go
amartyasinha: i've tried! PTAL if I understood it correctly.

internal/placement/volumes.go
```diff
@@ -39,7 +41,7 @@ func getVolumes(name string) []corev1.Volume {
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
 					DefaultMode: &configMode,
-					SecretName:  name + "-config-data",
+					SecretName:  internalcommon.GetServiceConfigSecretName(name),
```
amoralej, line 44: This patch left hardcoded secret name in https://github.com/openstack-k8s-operators/nova-operator/blob/5acd3e57e015ff27f78ba6f87f30504a6b431d9b/internal/controller/placement/api_controller.go#L1525 but it's cleaned later, when commonizing GenerateConfigs so no problem.

mrkisaolamb (approved):
looks good

internal/controller/nova/common.go
amartyasinha: Hey @SeanMooney! Sorry for the ping, but I'll appreciate if you could review the PR.

internal/webhook/nova/v1beta1/novaapi_webhook.go
```diff
@@ -64,7 +62,7 @@ func (d *NovaAPICustomDefaulter) Default(_ context.Context, obj runtime.Object)
 	novaapi, ok := obj.(*novav1beta1.NovaAPI)
 
 	if !ok {
-		return fmt.Errorf("%w but got %T", errExpectedNovaAPIObject, obj)
+		return fmt.Errorf("expected a NovaAPI object but got %T: %w", obj, internalcommon.ErrUnexpectedObjectType)
```
amoralej, line 65: This change is changing Error types from specific to generic ErrUnexpectedObjectType and slightly changing error text by adding `unnexpected object type` at the end. I haven't found any code checking error types and error message change seems fine..

internal/controller/placement/api_controller.go
amoralej: this and another couple of errors were left in commit e655c73cd3a0f2cd2be602e307c48a32e149c5fa . This commit fails to build:

```
go vet ./...
# github.com/openstack-k8s-operators/nova-operator/internal/controller/placement
internal/controller/placement/api_controller.go:1019:12: r.List undefined (type *PlacementAPIReconciler has no field or method List)
internal/controller/placement/api_controller.go:1051:11: r.List undefined (type *PlacementAPIReconciler has no field or method List)
# github.com/openstack-k8s-operators/nova-operator/internal/controller/nova
internal/controller/nova/common.go:34:2: "k8s.io/apimachinery/pkg/runtime" imported and not used
make: *** [Makefile:131: vet] Error 1
```
While this has been fixed on the following commit in the PR, it's not a good practice, all commits should be buildable,

internal/controller/nova/common.go
```diff
@@ -29,10 +29,7 @@ import (
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/runtime"
```
amoralej, line 35: Should be part of the previous commit.

internal/controller/placement/api_controller.go
```diff
@@ -1017,7 +884,7 @@ func (r *PlacementAPIReconciler) findObjectsForSrc(ctx context.Context, src clie
 			FieldSelector: fields.OneTermEqualSelector(field, src.GetName()),
 			Namespace:     src.GetNamespace(),
 		}
-		err := r.List(ctx, crList, listOps)
+		err := r.Client.List(ctx, crList, listOps)
```
amoralej: Should be part of the previous patch.

internal/controller/placement/api_controller.go
```diff
@@ -1049,7 +916,7 @@ func (r *PlacementAPIReconciler) findObjectForSrc(ctx context.Context, src clien
 	listOps := &client.ListOptions{
 		Namespace: src.GetNamespace(),
 	}
-	err := r.List(ctx, crList, listOps)
+	err := r.Client.List(ctx, crList, listOps)
```
amoralej: Should be part of the previous patch.

Makefile
```diff
@@ -150,11 +150,15 @@ PROC_CMD = --procs ${PROCS}
...
 # for the operator code that changed (e.g., skip Placement tests if only Nova code changed).
-test: manifests generate fmt vet envtest ginkgo ## Run tests.
+test: manifests generate fmt vet envtest gotest-unit ginkgo ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) -v debug --bin-dir $(LOCALBIN) use $(ENVTEST_K8S_VERSION) -p path)" \
 	OPERATOR_TEMPLATES="$(PWD)/templates" \
 	$(GINKGO) --trace --cover --coverpkg=../../internal/...,../../api/nova/v1beta1,../../api/placement/v1beta1 --coverprofile cover.out --covermode=atomic --randomize-all ${PROC_CMD} $(GINKGO_ARGS) ./test/...
 
+.PHONY: gotest-unit
+gotest-unit: ## Run unit tests under test/unit/.
+	go test ./test/unit/...
```
amoralej: adding -v would be nice to make sure all tests are being executed.

```
go test -v ./test/unit/...
```

Makefile
```diff
@@ -150,11 +150,15 @@ PROC_CMD = --procs ${PROCS}
 .PHONY: test
 # TODO: Currently runs all tests (Nova + Placement). In future, optimize CI to run only tests
 # for the operator code that changed (e.g., skip Placement tests if only Nova code changed).
-test: manifests generate fmt vet envtest ginkgo ## Run tests.
+test: manifests generate fmt vet envtest gotest-unit ginkgo ## Run tests.
```
amoralej: We rely mostly in envtest functional test in service operators, but I think using unit tests for this common libraries is a good addition, thanks!

I think this is running unit tests twice as the ginkgo command below on ./test/ already runs these tests.

test/unit/common/secret_test.go
```diff
@@ -0,0 +1,173 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// placement uses EnsureSecret for the OpenStack password secret referenced in
+// PlacementAPI.Spec.Secret with PasswordSelectors.Service defaulting to
+// PlacementPassword.
+func TestEnsureSecret_placementMissingSecret(t *testing.T) {
```
amoralej, line 38: Nit: It only validates the NotFound case which has specific logic. For other errors in `eader.Get(ctx, secretName, secret)` it's does not have coverage.

test/unit/common/network_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	internalcommon "github.com/openstack-k8s-operators/nova-operator/internal/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func TestEnsureNetworkAttachments_noAttachments(t *testing.T) {
```
amoralej, line 35: Nit: similar to ensureSecret, missing coverage for other errors than NotFound.

internal/common/watch.go
```diff
@@ -0,0 +1,134 @@
...
+// FindObjectsForSrcByField returns reconcile requests for CRs in src's namespace
+// that reference src via one of the indexed watchFields.
+func FindObjectsForSrcByField[L client.ObjectList, E any, S ~[]E](
+	ctx context.Context,
+	log logr.Logger,
+	reader client.Reader,
+	src client.Object,
+	watchFields []string,
+	newList func() L,
+	getItems func(L) S,
```
amoralej: There may be a simpler way to simplify this by passing a single parameter NewList. AI proposes to use `meta.ExtractList(crList) ` so that only the first parameter is needed. i.e.

```
 // FindObjectsForSrcByField returns reconcile requests for CRs in src's namespace
 // that reference src via one of the indexed watchFields.
-func FindObjectsForSrcByField[L client.ObjectList, E any, S ~[]E](
+func FindObjectsForSrcByField(
        ctx context.Context,
        log logr.Logger,
        reader client.Reader,
        src client.Object,
        watchFields []string,
-       newList func() L,
-       getItems func(L) S,
+       newList func() client.ObjectList,
 ) []reconcile.Request {
-       requests := []reconcile.Request{}
+       var requests []reconcile.Request
 
        for _, field := range watchFields {
                crList := newList()
@@ -54,23 +55,25 @@ func FindObjectsForSrcByField[L client.ObjectList, E any, S ~[]E](
                        return requests
                }
 
-               requests = appendRequestsForObjects(log, src, requests, itemsAsObjects(getItems(crList)))
+               items, err := meta.ExtractList(crList)
+               if err != nil {
+                       log.Error(err, fmt.Sprintf("extracting items from %s", crList.GetObjectKind().GroupVersionKind().Kind))
+                       return requests
+               }
+               requests = appendRequestsForObjects(log, src, requests, items)
        }
 
        return requests
 }
 
```

Additional changes are needed in appendRequestsForObjects.

```
 func appendRequestsForObjects(
        log logr.Logger,
        src client.Object,
        requests []reconcile.Request,
-       items []client.Object,
+       items []runtime.Object,
 ) []reconcile.Request {
        for _, item := range items {
-               log.Info(fmt.Sprintf("input source %s changed, reconcile: %s - %s", src.GetName(), item.GetName(), item.GetNamespace()))
+               accessor, err := meta.Accessor(item)
+               if err != nil {
+                       log.Error(err, "extracting object metadata")
+                       continue
+               }
+               log.Info(fmt.Sprintf("input source %s changed, reconcile: %s - %s", src.GetName(), accessor.GetName(), accessor.GetNamespace()))
 
                requests = append(requests,
                        reconcile.Request{
                                NamespacedName: types.NamespacedName{
-                                       Name:      item.GetName(),
-                                       Namespace: item.GetNamespace(),
+                                       Name:      accessor.GetName(),
+                                       Namespace: accessor.GetNamespace(),
                                },
                        },
                )
```

Also, the rest of functions can follow the same approach. so that from services call is only:

```
+++ b/internal/controller/placement/api_controller.go
@@ -879,8 +879,7 @@ func (r *PlacementAPIReconciler) findObjectsForSrc(ctx context.Context, src clie
                r.Client,
                src,
                allWatchFields,
-               func() *placementv1.PlacementAPIList { return &placementv1.PlacementAPIList{} },
-               func(l *placementv1.PlacementAPIList) []placementv1.PlacementAPI { return l.Items },
+               func() client.ObjectList { return &placementv1.PlacementAPIList{} },
        )
 }
```
I'm attaching the entire change that AI proposed. At least this is passing the tests.

[watchers.patch](https://github.com/user-attachments/files/29597169/watchers.patch)

amoralej (commented):
PR looks mostly good to me. Some coments in line.

- One of the intermediate commits does not build. Althoug the PR is fine, having unbuildable commits is undesired IMO.
- The impelementation of the new common watcher may be simplified. I'm proposing an approach. Current implementation works.
- The rest of comments are mostly nits.

Thanks for splitting the PR in commits, it'd be much harder to review otherwise!

internal/controller/placement/api_controller.go
amartyasinha: Thanks @amoralej ++ for catching this. Yes, I agree that all inidividual commits should be buildable. I was taking care of it, but seems like during the last regrouping of commits, it got messed up.

Anyways, it's been fixed now.

```
[AI-Generated]

Verified end to end. Results:


  Final tree unchanged

  ┌────────────────────┬──────────────────────────────────────────┐
  │ Reference          │ Tree hash                                │
  ├────────────────────┼──────────────────────────────────────────┤
  │ Old tip (096833f2) │ a780cab3467aefac63c0dbfd81d93dbeedb7c96d │
  ├────────────────────┼──────────────────────────────────────────┤
  │ New tip (2aaf127d) │ a780cab3467aefac63c0dbfd81d93dbeedb7c96d │
  └────────────────────┴──────────────────────────────────────────┘

  git diff 096833f2^{tree} HEAD^{tree} is empty — file content at the tip is identical to
  before the regroup. Only commit hashes changed.

  Every commit builds on its own

  Checked each of the 8 commits from 2085b0cf..HEAD individually:

  • go vet ./...
  • go vet ./api/...
  • go build -o /dev/null ./...
  • make vet (same as CI)

  All passed on every commit:
```

internal/webhook/nova/v1beta1/novaapi_webhook.go
amartyasinha: Yes, the reason was to keep things simple. So, we changed Error types from specific to generic, since we were not checking those specific Error type later in our code.

Makefile
amartyasinha: Thanks for the catch! I've removed dedicated `gotest-unit` from `test` target, as `ginkgo` will already run them.

Makefile
amartyasinha: Thanks! done.

test/unit/common/secret_test.go
amartyasinha: Added more coverage. PTAL

test/unit/common/network_test.go
amartyasinha: Added more coverage. PTAL

internal/controller/nova/common.go
amartyasinha: Done

internal/controller/placement/api_controller.go
amartyasinha: Done. Thanks

internal/controller/placement/api_controller.go
amartyasinha: Done. Thanks

internal/placement/volumes.go
amartyasinha: Ack! Will leave as it is since it got cleaned up later.

internal/common/watch.go
amartyasinha: Updated. Thanks

amartyasinha:
recheck

internal/common/watch.go
```diff
@@ -0,0 +1,137 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// FindObjectsForSrcByField returns reconcile requests for CRs in src's namespace
+// that reference src via one of the indexed watchFields.
+func FindObjectsForSrcByField(
```
mrkisaolamb, line 36: Can we add some unit tests for these functions: FindObjectsForSrcByField, FindObjectsForSrcInNamespace, and FindObjectsWithAppSelectorLabelInNamespace

amartyasinha: added. PTAL

amartyasinha:
recheck

amartyasinha:
recheck

mrkisaolamb (approved):
Hey @amartyasinha  and @amoralej  o/ Thanks for the review, @amoralej ! It looks like all comments from me and @amoralej  have been addressed, so I'll go ahead and merge this PR. If @amoralej has any additional feedback later, @amartyasinha  can handle it in a follow-up PR

PR #1136, opened 2026-06-25
Set service_type in [keystone_authtoken] for access rule validation

Deydra71:
Without service_type configured, keystonemiddleware cannot validate application credentials with custom access rules, causing HTTP 401 for end users.

Closes: [OSPRH-22365](https://redhat.atlassian.net/browse/OSPRH-22365)

### DIFF
--- a/templates/nova/nova.conf
+++ b/templates/nova/nova.conf
@@ -283,6 +283,7 @@ username = {{ .nova_keystone_user }}
 password = {{ .nova_keystone_password }}
 {{ end }}
 region_name = {{ .openstack_region_name }}
+service_type = compute
 # This is part of hardening related to CVE-2023-2088
 # https://docs.openstack.org/nova/latest/configuration/config.html#keystone_authtoken.service_token_roles_required
 # when enabled the service token user must have the service role to be considered valid.
--- a/templates/placement/api/config/placement.conf
+++ b/templates/placement/api/config/placement.conf
@@ -33,6 +33,7 @@ project_name = service
 region_name = {{ .Region }}
 {{ end -}}
 interface = internal
+service_type = placement
 
 [oslo_policy]
 policy_file=/etc/placement/policy.yaml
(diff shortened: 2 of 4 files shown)


gibizer (approved):
Make sense. Thanks

mrkisaolamb (approved):
lgtm

SeanMooney (approved):
looking at https://redhat.atlassian.net/browse/OSPRH-22365 this does seam valid

im not sure that this is docuemented partically well upstream and that reminds me we aslo need to enable apllciation creditial in devstakc in the upstream ci.

but this is good to knwo that we shoudl be populating this to have the correct behaivor

Deydra71:
Hi! @SeanMooney @gibizer There's been a question raised in [glance-operator](https://github.com/openstack-k8s-operators/glance-operator/pull/933) to propagate service_type using templateParameters eg - [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/glance/const.go#L36

In nova we would need to put it in every nove controller, correct? WDYT about hardcoding vs using the template parameter.

SeanMooney:
> Hi! @SeanMooney @gibizer There's been a question raised in [glance-operator](https://github.com/openstack-k8s-operators/glance-operator/pull/933) to propagate service_type using templateParameters eg - [1]
> 
> [1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/internal/glance/const.go#L36
> 
> In nova we would need to put it in every nove controller, correct? WDYT about hardcoding vs using the template parameter.

hardcoding it is the correct approch 
service typs are not configurable they are dirfed by 
https://opendev.org/openstack/service-types-authority/src/branch/master/service-types.yaml

and are quried to have specific value for cloud interoperatblity

so for nova nad placmeent this patch is good as is

we dont really need to pass it as a template parmater since its not something we will be changing so there is no need to add a constant in go and pass that in to the template rendering

Deydra71:
recheck

PR #1144, opened 2026-06-29
[vncproxy]Disable traffic logging

gibizer:
By default we set [DEFAULT]debug=true in all service config to increase our troubleshooting chances. But vncproxy with that setting emits a lot of traffic logging. By setting debug=false (and with the nova fix in depends-on) we disable traffic logging by default.

Depends-On: https://gitlab.cee.redhat.com/eng/openstack/nova/-/merge_requests/95
Related: [OSPRH-29916](https://redhat.atlassian.net/browse/OSPRH-29916)

### DIFF
--- a/templates/nova/nova.conf
+++ b/templates/nova/nova.conf
@@ -14,7 +14,14 @@ log_rotation_type=size
 {{if (index . "log_file") }}
 log_file = {{ .log_file }}
 {{end}}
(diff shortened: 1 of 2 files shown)


gibizer:
/unhold 
nova backport landed

mrkisaolamb (approved):

PR #1146, opened 2026-07-07
Add configurable probes and APITimeout-based probe defaults

amartyasinha:
Expose spec.override.probes on Nova and PlacementAPI CRDs and build pod
probes via lib-common CreateProbeSet helpers with shared defaults in
internal/common/probes.go. Tune placement-api liveness/readiness from
APITimeout to avoid restarting busy pods.


internal/placement/deployment.go
```diff
@@ -40,34 +41,22 @@ func Deployment(
...
-		// TODO might need tuning
-		TimeoutSeconds:      30,
-		PeriodSeconds:       30,
-		InitialDelaySeconds: 5,
-	}
-	readinessProbe := &corev1.Probe{
-		// TODO might need tuning
-		TimeoutSeconds:      30,
-		PeriodSeconds:       30,
-		InitialDelaySeconds: 5,
```
mrkisaolamb, line 53: I think we lost initialDelaySeconds because it's missing from the readinessProbe in our common probes

mrkisaolamb (changes requested):
Could you add tests to verify we aren't changing the probes values and behavior of compute, conductor, novncproxy, scheduler etc. (we currently only have checks for api and metadata)? I think this should be the first step to ensure we don't break the current behavior that we know is working (at least for now). As a next step, we can try finding more accurate values and leaving comments explaining why we changed them

internal/placement/deployment.go
amartyasinha: I've patched to include `InitialDelaySeconds`. If we later decide to have startup probe for placement api, we can again drop `InitialDelaySeconds` at that time.

amartyasinha: I've also modified the tests to increase coverage and asserting that probe values are not getting changed

amartyasinha:
recheck

mrkisaolamb (approved):
lgtm

PR #1155, opened 2026-07-24
Run Cyborg containers as cyborg user in the controlplane

amoralej:
Currently, it's running as root (uid = 0) needlessly instead of as cyborg user.

This patch sets both the dbsync, conductor and api to run as cyborg user.

### DIFF
--- a/internal/cyborg/constants.go
+++ b/internal/cyborg/constants.go
@@ -45,12 +45,13 @@ const (
 	// DefaultsConfigFileName is the file name with default configuration
 	DefaultsConfigFileName = "00-default.conf"
 
-	// DBSyncCommand is the kolla command to run the dbsync job
-	DBSyncCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
-
 	// CyborgLogPath is the default path for the cyborg service logs
 	CyborgLogPath = "/var/log/cyborg/"
 
 	// LogVolume is the name of the EmptyDir volume used for log streaming
 	LogVolume = "logs"
+
+	// CyborgUserID is the linux user ID used by Kolla for the cyborg user
+	// in the service containers
+	CyborgUserID int64 = 42483
 )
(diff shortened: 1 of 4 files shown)


amoralej:
check-rdo

amoralej:
check-rdo

internal/cyborg/constants.go
```diff
@@ -45,12 +45,13 @@ const (
 	// DefaultsConfigFileName is the file name with default configuration
 	DefaultsConfigFileName = "00-default.conf"
 
-	// DBSyncCommand is the kolla command to run the dbsync job
-	DBSyncCommand = "/usr/local/bin/kolla_set_configs && /usr/local/bin/kolla_start"
```
amartyasinha, line 49: Overall LGTM, but I was wondering since we removed `kolla_set_configs` execution, don't we need some workaround for this command? (asking since I don't know what this command was doing) Can we just simply remove it? If yes, then it's good.

SeanMooney: kolla_start calles kolla_set_configs internally

and we are now using internalcommon.ServiceComman

https://github.com/openstack-k8s-operators/nova-operator/blob/main/internal/common/kolla.go#L24

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/start.sh#L8

so calling /usr/local/bin/kolla_set_config was never requried when usign kolla_start

its only requried expclity if your going to run a raw command instead fo going though kolla_start

SeanMooney (approved):

internal/cyborg/constants.go
amoralej: thanks @SeanMooney, that's exactly the reason.

amartyasinha: Thanks @SeanMooney for the explanation and thanks @amoralej for the changes.

amartyasinha (approved):
/lgtm

PR #1166, opened 2026-07-27
switch anyuid SCC to nonroot-v2 for Nova, Placement, and Cyborg

amartyasinha:
the Nova, PlacementAPI, and Cyborg controllers unconditionally granted use on the anyuid SCC to their workload service accounts, allowing a namespace-admin to launch arbitrary pods as UID 0 via the operator-created RoleBinding.

- Replace anyuid with nonroot-v2 in RBAC kubebuilder markers and runtime rbacRules for all three controllers (nova_controller.go, placement/api_controller.go, cyborg_controller.go)
- Add SeccompProfile RuntimeDefault to all Nova, Placement, and Cyborg container SecurityContexts
- Set AutomountServiceAccountToken: false on all Nova, Placement, and Cyborg pod specs
- Update kuttl assertions from openshift.io/scc: anyuid to nonroot-v2

Assisted-by: Claude Sonnet 4.5

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -268,7 +268,7 @@ rules:
 - apiGroups:
   - security.openshift.io
   resourceNames:
-  - anyuid
+  - nonroot-v2
   resources:
   - securitycontextconstraints
   verbs:
--- a/internal/controller/cyborg/cyborg_controller.go
+++ b/internal/controller/cyborg/cyborg_controller.go
@@ -89,7 +89,7 @@ func (r *CyborgReconciler) GetLogger(ctx context.Context) logr.Logger {
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=roles,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=get;list;watch;create;update;patch
-// +kubebuilder:rbac:groups="security.openshift.io",resourceNames=anyuid,resources=securitycontextconstraints,verbs=use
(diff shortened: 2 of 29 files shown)


SeanMooney (changes requested):
nonroot-v2 is incompatble with our use of sudo and wourl break the kolla image abi which is part of our public continer image contract with our custoemrs

we intentially did not do this in 18 becasue of that.

so i am aware that this is being dicsused as a potical encucnemet for 19 but we shoudl nto rush to make this chagne

once this land main fundementally will not be compatbale with 18 content as we cant proceed with thsi while using kolla copy config and the ca cert or kolla-start to manage teh permsiosn and ownershipc of the configfiles proejcted form teh cofnig maps

amartyasinha:
Thanks @SeanMooney for the comment. We definitely don't want to break anything. These were the recommended steps as part of security hardening. But we will not blindly apply them and break our workflow.

> nonroot-v2 is incompatble with our use of sudo and wourl break the kolla image abi which is part of our public continer image contract with our custoemrs

sure, should we plan to remove kolla? it's also getting discussed (will ping you personally). if yes, what are the initial challenges do you think we will face?

> we intentially did not do this in 18 becasue of that.

ack! but due to security hardening, we cannot postpone it for longer period.

> so i am aware that this is being dicsused as a potical encucnemet for 19 but we shoudl nto rush to make this chagne

agree! maybe we can prioritize to see how we can do it.

> once this land main fundementally will not be compatbale with 18 content as we cant proceed with thsi while using kolla copy config and the ca cert or kolla-start to manage teh permsiosn and ownershipc of the configfiles proejcted form teh cofnig maps

yes, we won't rush to get it merged without looking into the compatibility issues.

amartyasinha:
Closing this PR in favor of https://github.com/openstack-k8s-operators/nova-operator/pull/1168. Might reopen this one if we plan on dropping kolla.

PR #1168, opened 2026-07-28
set AutomountServiceAccountToken=false on Nova, Placement and Cyborg pods

amartyasinha:
Disable automatic service-account token projection for all Nova, Placement and Cyborg workload pods. None of these pods need to call the Kubernetes API, so mounting the token is unnecessary and increases the blast radius of a container-escape compromise.

### DIFF
--- a/internal/cyborg/api/statefulset.go
+++ b/internal/cyborg/api/statefulset.go
@@ -158,7 +158,8 @@ func StatefulSet(
 					Labels: labels,
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: instance.Spec.ServiceAccount,
+					ServiceAccountName:           instance.Spec.ServiceAccount,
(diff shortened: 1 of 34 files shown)


SeanMooney (dismissed):
the approach looks reasonable to me but this is missing tests

can you add some envtest coverage or some kuttle assert to test the pod spec changes propagate properly

amartyasinha:
> the approach looks reasonable to me but this is missing tests
> 
> can you add some envtest coverage or some kuttle assert to test the pod spec changes propagate properly

Added! thanks

SeanMooney (approved):

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

PR #1170, opened 2026-07-29
enable periodic host discovery by default

amartyasinha:
Set [scheduler]discover_hosts_in_cells_interval=900 so nova-scheduler
automatically discovers new compute nodes every 15 minutes without any
manual intervention.

### DIFF
--- a/templates/nova/nova.conf
+++ b/templates/nova/nova.conf
@@ -113,6 +113,7 @@ query_placement_for_availability_zone=true
 query_placement_for_image_type_support=true
 enable_isolated_aggregate_filtering=true
 image_metadata_prefilter=true
+discover_hosts_in_cells_interval=900
 
 {{end}}
 


api/nova/v1beta1/nova_types.go
```diff
@@ -179,10 +179,6 @@ type NovaStatus struct {
 	// configuration.
 	RegisteredCells map[string]string `json:"registeredCells,omitempty"`
 
-	// DiscoveredCells is a map keyed by cell names that have discovered all kubernetes managed
-	// computes in cell value is a hash of config from all kubernetes managed computes in cell
-	DiscoveredCells map[string]string `json:"discoveredCells,omitempty"`
-
```
SeanMooney, line 185: i think this is technically a backward incompatable crd change

its in the status field so technially we might not break existing CRs but if we were beign strcit about version that would resutl in a version bump.

internal/nova/host_discover.go
```diff
@@ -1,104 +0,0 @@
-/*
```
SeanMooney, line 1: im not sure that adding the ablity to enable the expensive auto discoverty (this is not something we woudl want to turn on for large deployments)
should also remove the manual discovery

api/nova/v1beta1/nova_types.go
amartyasinha: well! since I was removing the manual discovery job, I thought of cleaning up the other resources related to it. but if we won't remove the manual discovery job, we will keep DiscoveredCells field in CRD.

gibizer: yeah removing fields from status is interesting, somebody might started depending for that field from outside. So technically we can keep the field but if the job does not run then the value of the field is still not useful for whoever started depending on it. So it is almost like the same effect if we just drop the field we are stopped populating.

internal/nova/host_discover.go
gibizer: This is more complicated than that unfortunately.
1. What this patch removes is an auto discovery for nova-computes running in k8s implemented in the operator. This was added as 
  * in Antelope enabling auto discovery in the scheduler is racy and unnecessarily expensive if we have many scheduler replicas.
  * for nova-computes running in k8s the operator knows when a new compute is deployed and therefore knows exactly when to run the discovery.
2. This operator implemented efficient auto discovery was never applied to compute nodes run on the EDPM node as the operator had no trigger when an EDPM compute is successfully deployed.
3. As the operator master branch is now tracking work for upstream version Hibiscus+ we have the leader election patch https://github.com/openstack/nova/commit/e98393c5c26743ec4c862af3e1a4beaa7f2d174b in the scheduler making the auto discovery in the scheduler viable without the race and the extra cost. Hence this patch is being born to start using such improved scheduler autodiscovery as that can be applied to EDPM computes as well.

My recommendations:

1. If we feel that the current operator based auto discovery of k8s based nova-computes is still more efficient than the scheduler periodic task based one. Then sure we can keep the operator logic in place. It might race with the auto discovery in the scheduler but I'm OK to take that risk and just document it how to turn it off.
2. I still recommend to enable the scheduler based periodic job now that we made it a lot better upstream as it will help removing a manual step for most of the deployments out there when they are deploying or scaling out the deployment. It is especially relevant for any cases where the deployment / scale-out is automated and therefore the manual discovery created a challenge knowing when to run it in that deployment automation.
3. We can document that for very large very static deployments the auto discovery might be a waste and it can be turned off via customServiceConfig passed to the scheduler as usual.

SeanMooney: so i woudl agree that we should enable the perodic as the primary way to do the discovery.

my concern is that that is still not that cheap for large clouds so we may hit issue in larger multi cell deployment if it is the only way we support in the intealler.

if we move to a model whre we just document the way to trigger this manulally as the alternitive and docuement how to disable it via config overried i guess that is ok


upstream we default to -1 for the https://docs.openstack.org/nova/latest/configuration/config.html#scheduler.discover_hosts_in_cells_interval so we would have to choose a sane default. 

probaly something in the 15-60 min range so maybe 900s as a default could be ok.

in ci we can tune it to be shorter but people do not add new nodes often so most of the tiem this will just increase the load on the schduler for no reason so 

discover_hosts_in_cells_interval=300 seams agressive

out side of the initall deployment i woudl be much more inclide to say this shoudl run hourly or even daily espically for larger deployment so every 5 mins seams too frequequent.

its unfortunet we never got aroudn to makeing the dataplane opertor call us back or notify us some how that we need to trigger this.


this is an upstrema converstaion but given the compute need to call the conductor on startup anyway to create the compute service, perhaps we could have the conductor call the schduler to register the compute so that the conductor does not need direct api db access. in general it would be nicer if we coudl get ride of the perosidc or the manual step upstream and make the cell registation automatic in the long term.

but over all i agree it woudl be ncie to remove the manual step on scaling so im not againt enabelting the perodic as long as we dont create a lot of load for no reaons.
there is a tradeoff between idel efficency when you are not actively expanding the cloud and latency when you do.

api/nova/v1beta1/nova_types.go
SeanMooney: if this was a spec field then we could not remove it, we have been told that there is no plan ot make a new major crd version change for the next major release at this time.

i think we can drop it because we said that the only status condtion the other operatrs should depend on is the top level ready condtion and the rest are just diganostic for a human to review.
with that said we do have to be a litele careful and not do status change blindly.

i just wanted to rasie this to dicusce it but i was not nessisarly suggetion we dont drop it if i we remvoe the k8s job.
i was half expecting this to fail the crd version cehck lint rule but im not sure how strict that is confired to be.

internal/nova/host_discover.go
amartyasinha: I agree 300s seems to be aggressive and will put more load on the scheduler, and 900s or some other reasonable value like that will be better.

Regarding dropping or keeping manual discovery steps, I'll leave it to you folks since I don't have much expertise in large scale deployment.

gibizer: > my concern is that that is still not that cheap for large clouds 

I thought we removed most of the inefficiencies by not running the same periodic in every scheduler in parallel (i.e. with the leader election)

> so we may hit issue in larger multi cell deployment if it is the only way we support in the intealler.

Today and in the future running the nova-manage CLI command externally is and will be supported. So it is definitely not the only way. (And today for EDPM nodes the CLI is the only way)

> probaly something in the 15-60 min range so maybe 900s as a default could be ok.

Making the default period long means we are optimizing for those rare an big deployments and therefore each small deployment either needs to reconfigure the period or do what they did before, calling the CLI command manually. I would rather optimize for the more frequent small deployments where having a modest 60 seconds or even less does not cause any DB spike and also creates the UX out of the box that you add computes and they are automatically discovered in a reasonable time. Sure we can add a note to our scaling guide that if you are deploying a huge cluster and you see DB spikes periodically then you can tune our default period higher accepting that you might need to run manual discovery if you need a compute add quickly.

SeanMooney: the nova changes i made with the leader election where to prevent the race conditon fo two concurrent mappings, that didnt really improve efficncy

on one hadn its only one scatter gather across all cells but adding hosts is a rare event in most clouds. so i would say optimising for addign and removing node on less then a 15 min interval si not really optimising for a normal usecase.

60 second seam for to short to me if we want to go back to 300 that is fine but i think if you need to rapidly discover hosts usign the peroid ic not really the right approch.

if that is the case we are optimsiting for we shoudl modify nova ot have the conductor call the scheduler to trigger the maping when it creates a new compute service mapping or simialr so that we move this to be part fo the normal compute node enrolment process automaticly rather then async

gibizer: > the nova changes i made with the leader election where to prevent the race conditon fo two concurrent mappings, that didnt really improve efficncy

but before that change we run N parallel scatter gathers in N scheduler replicas. After the leader election change we run 1 scatter-gather even if there are N = 3 or more scheduler pods. So I think we improved the used resources a lot by reducing the number of DB queries from N to 1 per period.

gibizer: Anyhow I rest my case. I'm not blocking this to land.

amartyasinha: @SeanMooney Hey, Can you please review the PR? It's no longer running scheduler based discovery aggressively, as well as k8s based discovery job is there too. Thanks!

SeanMooney (approved):

SeanMooney:
if we want to tweak this down to 300s which was the orginal value we can, below that i think would be too frequent for a default but as long as we document some guidence on this im not against choosing a persona and trunign for that.  300-900 range i think would work for most.

if we want to follow up and remove the job based reconcile thats valid but we need the doc change ready to go with that and there is curretly a docs freeze so i woudl avoid making that change until that is lifted because it will be needed before the next release happens.

amartyasinha:
> if we want to tweak this down to 300s which was the orginal value we can, below that i think would be too frequent for a default but as long as we document some guidence on this im not against choosing a persona and trunign for that. 300-900 range i think would work for most.
> 
> if we want to follow up and remove the job based reconcile thats valid but we need the doc change ready to go with that and there is curretly a docs freeze so i woudl avoid making that change until that is lifted because it will be needed before the next release happens.

noted @SeanMooney! I'll track these down in Jira, and once doc freeze is lifted, we can follow up.

PR #1175, opened 2026-08-06
Update UID for cyborg user

amoralej:
Current one is used by frrvty user [1] so I am
switching to the next unused one.

[1] https://github.com/openstack-k8s-operators/tcib/blob/0f06bc7234b76aee419b1a6055a8324c06477b83/container-images/kolla/base/uid_gid_manage.sh#L43

### DIFF
--- a/internal/cyborg/constants.go
+++ b/internal/cyborg/constants.go
@@ -53,5 +53,5 @@ const (
 
 	// CyborgUserID is the linux user ID used by Kolla for the cyborg user
 	// in the service containers
-	CyborgUserID int64 = 42483
+	CyborgUserID int64 = 42485
 )


amoralej:
check-rdo

internal/cyborg/constants.go
```diff
@@ -53,5 +53,5 @@ const (
 
 	// CyborgUserID is the linux user ID used by Kolla for the cyborg user
 	// in the service containers
-	CyborgUserID int64 = 42483
```
SeanMooney, line 56: ack this was the kolla version

we can make this change now only because we have not release with it so i agree this is important to change now. we will also need to align edpm-ansible to this.

once we release with a uid it become part of the upgrade contract

SeanMooney (approved):

amoralej:
check-rdo

amoralej:
check-rdo

PR #1180, opened 2026-08-11
Add transport URL secret rotation with consumer finalizer

lmiccini:
## What this does

When infra-operator rotates a RabbitMQ transport URL (creating a new secret and user), consumer operators must hold a **consumer finalizer** on the old secret until all their pods have rolled out with the new credentials. Without this, infra-operator cleans up the old RabbitMQ user while pods are still connected with the old credentials, causing message-bus outages.

## Approach (updated)

> **Note:** this PR was reworked from the original grace-period design to an event-driven "applied input hash" design. The old approach released the finalizer after a fixed grace period based on the parent `Ready` condition, which was racy against the informer cache and could revoke credentials prematurely. The new approach releases the finalizer only once every child has provably rolled out.

1. Add a consumer finalizer to the current transport URL secret early in reconcile. `Status.TransportURLSecret` is set for first-time setup only (empty or unchanged); during rotation the status is updated solely by `FinalizeSecretRotation` at the end of reconcile.

2. Pass `transportURL.Status.SecretName` directly to sub-CR creation and config generation as a parameter — never read `Status.TransportURLSecret` for sub-CR specs.

3. Each child controller records an `AppliedInputSecretHash` in its status, set only after `statefulset.IsReadyForInput` / `deployment.IsReadyForInput` confirms — via an uncached API read — that the workload is fully rolled out with the expected `CONFIG_HASH`.

4. The parent mirrors a child's Ready condition only when its `Generation == ObservedGeneration` and `AppliedInputSecretHash` matches the current input hash; otherwise it sets the condition to Unknown.

5. Guard: `FinalizeSecretRotation` removes the consumer finalizer from the old secret only when every child reports the expected hash and is ready. The guard is computed from `Conditions.AllSubConditionIsTrue()` (not `IsReady()`), because `Conditions.Init()` resets the `Ready` condition to Unknown on every reconcile.

The same pattern applies to notification transport URL secrets and application-credential secrets where applicable.

## Dependency

Depends on the lib-common `IsReadyForInput` / `FinalizeSecretRotation` helpers (currently pinned via a `replace` to the fork commit while the lib-common PR is in review).


lmiccini:
recheck

lmiccini:
recheck

lmiccini:
recheck

SeanMooney:
i have a pure port of the watcher chagne here #1197  so perhaps it woudl make sense to leqave the config hash work to that sicne it also cover cybrog and placement then  your pr coudl focus on the secret roation feature

lmiccini:
recheck

lmiccini:
recheck

internal/controller/nova/nova_controller.go
```diff
@@ -821,6 +877,99 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+	}
+	for _, cellName := range orderedCellNames {
+		if s := instance.Annotations[cellTransportAnnotationKey(cellName)]; s != "" {
+			transportKeep = append(transportKeep, s)
+		}
+		if s, ok := cellSecretNames[cellName]; ok && s != "" {
+			transportKeep = append(transportKeep, s)
+		}
+	}
+	if err := object.PruneSecretConsumerFinalizers(
```
mrkisaolamb: Shouldn't we check allServicesReady to ensure we don't remove the finalizer from intermediate secret B while pods may still be using its credentials?

SeanMooney:
if you rebase i think the ci might now be passing again

lmiccini:
recheck

lmiccini:
recheck

lmiccini:
recheck

internal/controller/nova/nova_controller.go
```diff
@@ -367,12 +369,22 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
 	orderedCellNames := []string{novav1.Cell0Name}
+	otherCellNames := []string{}
 	for cellName := range instance.Spec.CellTemplates {
 		if cellName != novav1.Cell0Name {
-			orderedCellNames = append(orderedCellNames, cellName)
+			otherCellNames = append(otherCellNames, cellName)
 		}
 	}
+	sort.Strings(otherCellNames)
+	orderedCellNames = append(orderedCellNames, otherCellNames...)
```
SeanMooney: i see before we were taking all teh cells other then cell0 in what ever orter they were in the the cell template which may not have been stable

now all cells are strictly order with cell0 alwasy first the lexagraphic sort applied ot the rest

lmiccini:
recheck

mrkisaolamb (approved):
lgtm

PR #1198, opened 2026-08-25
Cleanup leftover kolla references

amartyasinha:
Now that we no longer use kolla, this commit cleanup leftover references of kolla


mrkisaolamb (approved):
lgtm

PR #1199, opened 2026-09-03
Adjust default config for threading mode in Nova services

mrkisaolamb:
Performance tests show we never reach the default number of concurrent tasks executed in the thread pool, so we can lower it to 32. Since threading mode will be enabled across all services, we should begin collecting logs for monitoring. Additionally, heartbeat_in_pthread is deprecated and no longer needs to be set.

### DIFF
--- a/templates/nova/nova.conf
+++ b/templates/nova/nova.conf
@@ -4,6 +4,14 @@
 # overall then serializing live migrations so set this to 1 explicitly
 max_concurrent_live_migrations=1
 state_path = /var/lib/nova
+# Note(ksambor) We want to monitor threads closely, at least in 19
+# so for debugging purposes, we want to collect logs at least once every
+# 5 minutes
+thread_pool_statistic_period=300
+concurrency_backend=threading
+{{if eq .service_name "nova-conductor" "nova-scheduler"}}
+executor_thread_pool_size=32
+{{end}}
 {{if eq .service_name "nova-api"}}
 allow_resize_to_same_host = true
 {{end}}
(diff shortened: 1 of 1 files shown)


gibizer (commented):
looks good to me

gibizer (approved):

gibizer:
@ashu-011 when this land could you rebase you s2i testing PR to get a tempest result after this config change?

ashu-011:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

templates/nova/nova.conf
```diff
@@ -4,6 +4,14 @@
 # overall then serializing live migrations so set this to 1 explicitly
 max_concurrent_live_migrations=1
 state_path = /var/lib/nova
+# Note(ksambor) We want to monitor threads closely, at least in 19
+# so for debugging purposes, we want to collect logs at least once every
+# 5 minutes
+thread_pool_statistic_period=300
+concurrency_backend=threading
+{{if eq .service_name "nova-conductor" "nova-scheduler"}}
+executor_thread_pool_size=32
```
SeanMooney, line 13: this feels larger then we proably shoud ldefault too i would suggest 4-8

the priamy scaling mechanium is still via replcias

SeanMooney: i dont know we will likely want to tune this over time based on data i just think 32 treads is rather high starting point

mrkisaolamb: Agreed, that’s a lot, but it’s still half of our current default, which is 64 ( https://docs.openstack.org/ocata/config-reference/compute/config-options.html ). We can reduce it to 24 (and I feel comfortable with that value), but going down to 4–8 without additional performance testing sounds riskier, IMO.

mrkisaolamb:
recheck

templates/nova/nova.conf
SeanMooney: i would ignore upstream and instead look at our requirements for the openshift cluster

https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/plan-ref_rhocp_cluster_requirements#ref_RHOCP-hardware-requirements

```
64 GB RAM
16 CPU cores
120GB NVMe or SSD for the root disk plus 250 GB storage 
```

we shoudl be tuning our pod to be able to run correctly on a worker with the min number of cores we require so wee shoudl not set it over 16

that why i was suggestin 4-8 because there are other things runing on those cores as well so i would not start with 32 unless we have tested 4-8 and see that it does nto work.

PR #1207, opened 2026-09-10
[s2i] Add nova content-provider and tempest job

ashu-011:


.zuul.yaml
```diff
@@ -212,6 +212,51 @@
...
+      - name: openstack-k8s-operators/s2i-openstack-containers
+        override-checkout: main
+    vars:
+      s2i_ci_images:
+        - nova/nova-api
+        - nova/nova-compute
+        - nova/nova-conductor
+        - nova/nova-novncproxy
+        - nova/nova-scheduler
+
```
SeanMooney, line 230: i woudl add placment and cyborg images ehre as well

.zuul.yaml
SeanMooney, line 126: instead of the new job you shoudl be updateing this to have
```
parent:  s2i-test-base
dependencies:
  - worklaods-s2i-content-provider
  - openstack-meta-content-provider
```

SeanMooney, line 146: and this should have 
```
parent: s2i-test-base-ceph
dependencies:
  - worklaods-s2i-content-provider
  - openstack-meta-content-provider
```

.zuul.yaml
```diff
@@ -212,6 +212,51 @@
         cpu_mode = custom
         cpu_models = Nehalem
 
+- job:
+    name: nova-s2i-content-provider
```
SeanMooney, line 216: i woudl rename this to worklaods-s2i-content-provider

that is assumign we do not want to have 3 content providers for nova cybrog and placment and the openstack-content-provider for the operators


openstack-meta-content-provider should be replaceed by  openstack-content-provider going forward. the differnece is that the non meta version does not supprot building rpms from upstream souces. since rpms are going away for the services we should be swapping back to the standard openstack-content-provider which only builds the operators.

.zuul.yaml
```diff
@@ -212,6 +212,51 @@
...
+        override-checkout: main
+    vars:
+      s2i_ci_images:
+        - nova/nova-api
+        - nova/nova-compute
+        - nova/nova-conductor
+        - nova/nova-novncproxy
+        - nova/nova-scheduler
+
+- job:
```
SeanMooney, line 231: and delete this entirly.

.zuul.yaml
```diff
@@ -246,6 +291,13 @@
               - ^go\.work.*
               - ^api/go\.mod$
               - ^api/go\.sum$
+        - nova-s2i-content-provider:
+            voting: false
+        - nova-s2i-test:
+            voting: false
```
SeanMooney, line 297: while we could start with thise as non voting these shoudl be voting in the next week or two and must be voting before beta so im inlcined to jsut start with them as voting.

SeanMooney (changes requested):

.zuul.yaml
ashu-011: Heya @SeanMooney, I applied this (existing tempest jobs parent s2i-test-base, depend on the s2i content-provider + meta CP). Then I found https://github.com/openstack-k8s-operators/s2i-openstack-containers/blob/main/docs/operator-onboarding.md, which says to keep the old kuttl/tempest jobs unchanged and add s2i as new non-voting siblings until they’re stable:

`Keep the existing kuttl and test-operator jobs unchanged, which can be used to test the newer operators with the old content. The s2i jobs are new siblings, not replacements. Once they are stable they should be turned to voting.`

Should we stick with updating the existing jobs, or go back to siblings (nova-s2i-content-provider / nova-s2i-test / nova-s2i-test-ceph) and leave these jobs as they are?

ashu-011: cc @amartyasinha FYI as you are going to take over

SeanMooney: So, we are not going to merge this patch until the job is stable and passing, and the deadline for that is next week as these new jobs need to then be refence in a new nova-operator-s2i-compoent-line template like this
https://github.com/openstack-k8s-operators/watcher-operator/pull/457/changes
and added to the s2i repo like this https://github.com/openstack-k8s-operators/s2i-openstack-containers/pull/203
to gate the promtoion of nova contaienr images/shas.

The current Tempest jobs are deploying the wrong version of OpenStack, which is why I'm asking you to remove them in this PR since they cannot correctly deploy hibiscus/2026.2

The Kuttl jobs should be updated to use the new images by updating the Golang constants and defaulting logic in this repo in the next two weeks as well; again, that is needed prior to Beta.

You can use the prebuilt quay.io images for that as a starting point, but they should be swapped over to the new S2I content provider when the Tempest jobs are.

Nova partly had support for using master but that was based rdo rpm builing and is not functional anymore,the nova operator repo is clsoer to  Watcher-operator than the other projects:

https://github.com/openstack-k8s-operators/s2i-openstack-containers/blob/main/docs/operator-onboarding.md#watcher-only-extras

So we should be moving to the standard content provider for the operator build only, and using the S2I content provider to build the service images for nova,placement and cybrog.

Kuttl and Tempest should both be using the operator and service images from those two content providers so that they are testing the correct content.

ashu-011:
recheck

.zuul.yaml
amartyasinha: thanks @SeanMooney for the comment, I'll go through your suggestions in a couple of hours.

